PR #221, opened 2015-01-05
This fixes a crash bug when switching between RTLSDR and USRP devices

benburkhart1:
When the RTLSDR is successfully used, and then the user switches over to their USRP device, gqrx starts a crash loop that cannot be escaped without deleting a line in your default.conf.

This change modifies the receiver.cpp to catch the exception when an invalid gain type is set via osmosdr, and fails with an error message.

I believe this to be the proper behavior.

This PR should fix https://github.com/csete/gqrx/issues/125

### DIFF
--- a/applications/gqrx/receiver.cpp
+++ b/applications/gqrx/receiver.cpp
@@ -436,7 +436,13 @@ receiver::status receiver::get_gain_range(std::string &name, double *start, doub
 
 receiver::status receiver::set_gain(std::string name, double value)
 {
-    src->set_gain(value, name);
+    try {
+	    src->set_gain(value, name);
+    }
+    catch (std::runtime_error &e) {
+      std::cout << "Error changing gain for " << name << " to " << value << ": " << e.what() << std::endl;
+      return STATUS_ERROR;
+    }
 
     return STATUS_OK;
 }


benburkhart1:
Error before:

terminate called after throwing an instance of 'std::runtime_error'
  what():  set_rx_gain: gain "LNA" not found for channel 0.
Available gains: PGA

[1]    15431 abort (core dumped)  gqrx
ben-biglaptop :: ~/gqrx/build ‹master*› »


new non-fatal error after:
Error changing gain for LNA to 0: set_rx_gain: gain "LNA" not found for channel 0.

csete:
Thanks, but the patch doesn't fix the actual bug, which is gqrx trying to set a non existent gain. I think we should fix the real bug instead of hiding it by catching the exception.

benburkhart1:
I disagree a bit. Setting a gain for an invalid gain type should not crash the application, in fact, I feel as though the way gqrx crashes all the time is a big usability problem. That being said, you feel as though the real bug is the fact that gqrx does not clear out the gains when switching devices? If so I can submit a patch fixing just that issue, but I feel as though there's some fundamentally wrong patterns with relying on crashes to signify a problem in operation.

csete:
> you feel as though the real bug is the fact that gqrx does not clear out the gains when switching devices? 

Yes, wouldn't it be much better to prevent the exception from happening? Surely there is nothing wrong pattern about that...

benburkhart1:
Yes, I fully agree with that, will close this and submit a different PR for that.

csete:
Sounds great, thanks.

PR #226, opened 2015-03-07
RDS

stanojr:
Hi,
this is RDS support for gqrx in WFM mode, please check and decide if you want to merge.
It depends on gr-rds from https://github.com/bastibl/gr-rds but gqrx can be compiled without it.
Maybe gr-rds decoder/parser can be bundled to gqrx? license is GPL.

To start decoder you need to click on it in menu like AFSK1200.

screenshot from 2015-03-07 22 52 27


csete:
Thanks for the patch. Overall it looks good but I am worried about the external dependency, even if only optional. 

I think that relying on external gr-rds package will make this function of limited use since people have know that they need gr-rds installed prior to building gqrx. Also note that gr-rds can be installed in a random directory, not only /usr/ or /usr/local.

If we can get away by including the CPP sources and headers under e.g. dsp/rds/ and only rely on gr-digital as optional dependency this could be very interesting. Are you interested in making these changes?

I also noted some wrong comments that look like copy & paste leftovers ;-)

stanojr:
Hi, thanks for review, i will make that changes a fix leftovers

stanojr:
So it's done

csete:
I have now merged it manually.
Perhaps we can reconsider the logic about how it is activated. Since RDS data goes into a dock window, it would probably not rely on activation from a separate menu item but perhaps have an enable button in the dock window.

mstyne:
@stanojr Thanks for your work on this feature!

stanojr:
@csete I get your point. I was first thinking that all special decoders should be in one place. But it doesn't make sense to have RDS in View menu and able to enable it via Tools menu. I make a patch and new pull request. I keep RDS dock by default hidden because IMHO that decoder will people not use very often, so it dont need to be always visible in docks. And some enable button/checkbox in that RDS dock. Is that OK ?

Also i found out there is some problem with decoder itself in rx_rds.cpp, for example it dont work with E4000, and it is not as reliable as gr-rds with weaker radio signals with R828T2 and FC0013 tuners. I try to debug/fix it.

@mstyne You are welcome :) I just want to learn something about gnuradio and maybe c++

tomswartz07:
Perhaps it could actively only be visible when GQRX is tuned to the FM-radio band (87.5-108MHz)?
As far as I know, RDS isn't used elsewhere...

Issue #71, opened 2013-06-05
Snap to Frequency Grid


m0mik:
When viewing 25MHz of bandwidth at once with my USRP2/WBX, clicking on a signal that pops up always fails to hit the exact frequency first time round.  I have to zoom in first and by the time I've tuned to the correct frequency the signal is gone!

A 'snap to frequency grid' feature which allows the user to click on a signal and auto tune to the nearest channel spacing would be great.  This would also enable the user to scroll the mouse up and down in the chosen channel spacing.

Here are some typical channel spacings which would be ideal:

1e3
5e3
6.25e3
8.33e3
12.5e3
25e3
50e3
100e3

Mike
## 

Mike Jameson M0MIK BSc MIET
Email: mike@scanoo.com
Web: http://scanoo.com

pinkavaj:
Is this still relevant? Btw would actually make sense to use the current peak detection and enable snapping to detected peask?

argilo:
"Snap to grid" would be useful for bands where channels have a fixed spacing, and other programs (like SDR#) have this feature. I think it's still relevant.

PR #231, opened 2015-03-23
Tuning steps

ssoqboss:
This request makes possible tuning by steps when you click on a spectrum - it takes freq and tune demod to closest to specified step, say 12.5kHz or 25kHz etc. Step is selected in rx options panel. Any other tuning works as usual.

### DIFF
--- a/applications/gqrx/mainwindow.cpp
+++ b/applications/gqrx/mainwindow.cpp
@@ -205,6 +205,7 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
     connect(uiDockRxOpt, SIGNAL(noiseBlankerChanged(int,bool,float)), this, SLOT(setNoiseBlanker(int,bool,float)));
     connect(uiDockRxOpt, SIGNAL(sqlLevelChanged(double)), this, SLOT(setSqlLevel(double)));
     connect(uiDockRxOpt, SIGNAL(sqlAutoClicked()), this, SLOT(setSqlLevelAuto()));
+    connect(uiDockRxOpt, SIGNAL(stepChanged(qint64)), ui->plotter, SLOT(setStep(qint64)));
     connect(uiDockAudio, SIGNAL(audioGainChanged(float)), this, SLOT(setAudioGain(float)));
     connect(uiDockAudio, SIGNAL(audioStreamingStarted(QString,int)), this, SLOT(startAudioStream(QString,int)));
     connect(uiDockAudio, SIGNAL(audioStreamingStopped()), this, SLOT(stopAudioStreaming()));
--- a/qtgui/dockrxopt.cpp
+++ b/qtgui/dockrxopt.cpp
@@ -33,7 +33,8 @@ DockRxOpt::DockRxOpt(qint64 filterOffsetRange, QWidget *parent) :
     QDockWidget(parent),
     ui(new Ui::DockRxOpt),
     agc_is_on(true),
-    hw_freq_hz(144500000)
+    hw_freq_hz(144500000),
+    step_hz(stepIdxToHz(STEP_1))
 {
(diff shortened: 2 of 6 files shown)


csete:
Thanks for the patch.

I don't think it is necessary to introduce a tuning step in the plotter widget. There is already a tuning step called "click resolution" which is currently set depending on the mode. We should be able to use that.

Furthermore, I think a standard int should be sufficient to represent a tuning  step.

ssoqboss:
> There is already a tuning step called "click resolution" which is currently set depending on the mod

It's not the same. Main reason for tuning by steps is that there are freq grids with fixed steps and when some ordinary signal appears on the waterfall it's the very convenient to tune exactly on the signal (which is respects freqs grid), even if you click somewhere around. This is especially helps on wide range spectrum, say 3 MHz, when you can't point signal very precisely with mouse. It's not connected with freq resolition, it's just a tuning helper.

csete:
> It's not the same. Main reason for tuning by steps is that there are freq
> grids with fixed steps and when some ordinary signal appears on the
> waterfall it's the very convenient to tune exactly on the signal (which is
> respects freqs grid), even if you click somewhere around.

Assuming that with "freq grids" you mean some fictive grids that are not actually drawn, what you describe is exactly what the "click resolution" (m_ClickResolution) does today. It snaps the mouse click or wheel events to 10 Hz (SSB), 100 Hz (AM, NFM) and 1 kHz (WFM) respectively. There is also m_FilterClickResolution which does the same for the filter widths.

ssoqboss:
> what you describe is exactly what the "click resolution" (m_ClickResolution) does today. It snaps the mouse click or wheel events to 10 Hz (SSB), 100 Hz (AM, NFM) and 1 kHz (WFM) respectively

Well, no, you misunderstood me. I mean that there are some common channels like on LPD, PMR and many other VHF/UHF services which are equally spaced by some fixed step, usually 25kHz, 12.5kHz, 6.25kHz and so on. HF broadcasters spaced by 5kHz, HF utility stations often work on freqs rounded by 1kHz. My patch makes it easy to quick tune rx demodulator with respect to chosen step by clicking on waterfall, but it not change real freq resolution - you can still tune to any freq by direct input or by mouse wheel. You'd better try it out to catch an idea, this approach realy helps tune exactly on signal on wide spectrum.

I can't say my code is well written and clean, just a first approach. But I believe it will help a lot to many as it helps me to monitor VHF and quickly jump between various channels without need to fine tune freq after mouse pointing.

nm_ClickResolution can't serve to it because even when you listen on some predefined channel it can be needed to little tune off from exact freq to prevent QRM/QRN or when buggy transmitter works with some shift from channel freq. In such situations it'd better to tune exactly on channel freq by click (my patch step) and then to fine tune with mouse wheel (nm_ClickResolution step).

csete:
I believe I have a good grasp of the concept of channel spacings and tuning steps on radios - I have  been in telecom for 25 years. What I don't fully understand is what you are trying to accomplish.

Do you mean that if you click on the spectrum it will round the frequency according to your tuning step but if you use the mouse wheel on the spectrum it will still round according to m_ClickResolution? If so aren't things upside down??? What you added is actually click resolution while what currently is called m_ClickResolution is in reality a tuning step.

Is it necessary to have both? I would expect that is there is a widget that allows me to select, say 12.5 kHz tuning step, then using the mouse wheel on the spectrum would take 12.5 kHz steps.

csete:
Let me try to rephrase my question.

With your patch applied, assume mode is AM (meaning m_ClickResolution = 100 Hz) and tuning step = 5 kHz.

If I click on the spectrum it will be rounded to nearest 5 kHz but then I can use the mouse wheel to make 100 Hz steps?

csete:
Related issue: #71

PR #238, opened 2015-04-03
remove filterButton code

michaelld:
The filterButton was removed from the UI in df629bd5 (2015-Apr-02).


csete:
Oh, I missed that, thanks.

PR #251, opened 2015-07-02
Volk dependency

beitler:
Without this modification the linker complains:

```
/usr/bin/ld: warning: libvolk.so.1.0, needed by [...]/lib/libgnuradio-analog.so, not found (try using -rpath or -rpath-link)
```


csete:
Thanks, but shouldn't this go into the pkgconfig file in the libgnuradio-analog package? Gqrx does not directly use volk and it seems strange to me to include dependencies not directly used.

beitler:
True. Thanks for pointing this out.

PR #260, opened 2015-08-24
first implementation of OIRT stereo decoder

dcherednik:
It works quite similar with pilot tone stereo decoder but using 31250 subcarrier for pll.

### DIFF
--- a/applications/gqrx/mainwindow.cpp
+++ b/applications/gqrx/mainwindow.cpp
@@ -893,6 +893,7 @@ void MainWindow::selectDemod(int index)
         /* Broadcast FM */
     case DockRxOpt::MODE_WFM_MONO:
     case DockRxOpt::MODE_WFM_STEREO:
+    case DockRxOpt::MODE_WFM_STEREO_OIRT:
         quad_rate = rx->get_input_rate();
         if (quad_rate < 200.0e3)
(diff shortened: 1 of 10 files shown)


csete:
Looks interesting, but can you provide some background info about this? Is it a specific stereo used some places or is it a better stereo decoder that could replace the existing one? I tried searching the web but couldn't find any explanation.

dcherednik:
OIRT was main stereo standard in the USSR, but still used in Russia and some former USSR countries for FM broadcast band 65.8 - 74 MHz .
OIRT also separates stereo for L+R and L-R and transmits L-R signal by AM modulation of subcarrier. The frequency of subcarrier is 31250Hz and gain is suppressed for 14db. Accordingly there is no pilot tone in OIRT. 

There are two main methods how to decode it:
- restore subcarrier for 14db and use AM detector to get L-R (this method was widely used in old receivers, they do it by high-Q LC)
- use synchronous demodulation

In the latter case (implemented by this patch) decoder works quite similarly to CCIR 
stereo decoder but uses subcarrier for PLL instead of doubling pilot-tone frequency. Also frequency range of demodulated signal is a bit narrower so i changed cutoff frequency to reduce noise in case of OIRT stereo.

csete:
Ok, got it, thanks for the info!

PR #261, opened 2015-08-27
FFT averaging

tejeez:
It's often possible to calculate several FFTs for each waterfall line and average them to make weak signals easier to see and to make the waterfall look better.


csete:
If we are to implement something like this we need to differentiate between FFT rate and refresh rate so that the user can have a clear understanding of what is happening.

If FFT rate > refresh rate we can average otherwise not. This way the FFT rate = refresh rate case would just be a special case and not require any dedicate mode, which IMHO is just confusing.

csete:
No update for 2 years.

Issue #253, opened 2015-07-04
Make RDS info panel scrollable


troffasky:
Using 2.3.2.110.222bf623-gqrx~utopic1 the length of the RDS string can cause the whole right-hand panel to jump around in size. If it had a scroll bar, this wouldn't happen.

PR #269, opened 2015-09-06
Small changes in RDS panel

rwnobrega:
Although no scroll bar is added, I believe that 40bca80 addresses #253.


rwnobrega:
I see... Since I don't have access to OS X, I better close this request.

csete:
I have cherry picked e606228788882e006712515d66b34fa3939201fa manually so at least thanks for that.

PR #289, opened 2015-11-08
Tweaking the UI to layout better.

treaves:
These changes better render the UI on Mac.  The original is pretty bad; icons overlaying fields, etc.  

Tested on Linux as well.  Should be fine on all platforms.

### DIFF
--- a/qtgui/dockfft.ui
+++ b/qtgui/dockfft.ui
@@ -36,27 +36,9 @@
    <string>FFT Settings</string>
   </property>
   <widget class="QWidget" name="dockWidgetContents">
-   <layout class="QVBoxLayout" name="verticalLayout">
-    <property name="spacing">
-     <number>5</number>
-    </property>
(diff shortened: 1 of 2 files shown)


csete:
Thanks for the patch. I was aware of the bad layout but didn't really know how to fix it. Do you know what caused the misalignments so I can avoid it in the future?

treaves:
I don’t.  I do know that layout with Qt is very particular.  There is still one icon thais off, for no apparent reason.  Another thing I’ve found is that once it is looking OK. on Mac, it’ll still be good on Linux for Windows.  

Basically, with the apps I’ve done this for, it’s a matter of tearing apart the layouts, and recreating them.  Grid in particular can be tricky; with this, there were several unused grid squares.  So it’s a matter of moving things around.

Lastly, things will look different in Designer versus the actual running app.  The one mis-aligned icon shows fine in Designer, yet is still slightly off in the app.

PR #290, opened 2015-11-11
Transition to cmake

treaves:
This is the minimum needed for a transition to cmake.  I've left the .pro file in place temporarily, for reference (it won't work as-is, as I did not update the paths).  I've updated the README to have the build instructions.

There may be some fine tuning required around Windows; the way the .pro file was written took for grated that there were things on default paths that just 'happened' to be found.  For those on Windows, please do not be discouraged, and just work through the issues & update the CMakeLists.txt files as needed.

Tested on Mac & Ubuntu 10.04.


csete:
Thanks for the patch. It works fine.

Windows is not an issue at the moment, since I don't think gnuradio & co are available on windows.

I would like to change the cmake required version to 2.8 because that is the version shipped with Ubuntu 14.04 (the oldest ubuntu officially supported by us). Was there any particular reason for requiring 3.2?

treaves:
I thought that was the version for Ubuntu 10.04.  Perhaps I upgraded mine.

I just tested, and 2.8.0 should be fine as a min version.

csete:
I'm  a little confused by this section in the readme file to me:

> To build in various mode, pass -DBUILDTYPE={typeOfBuild} to cmake. Supported build types are:
> Debug GProf Valgrind Release (This is currently disabled due to a conflict with log4cpp DEBUG #define)

Does this refer to cmake in general or only when building through Qt Creator?

AFAIK the cmake switch is -DCMAKE_BUILD_TYPE=... and possible options are Debug, Release, RelWithDebInfo and MinSizeRel. Also, what exactly is disabled due to a conflict with log4cpp?

-DBUILTYPE has no effect on my computer. On the other hand, -DCMAKE_BUILD_TYPE=Valgrind seems to work even though Valgrind is not listed as an option in the cmake docs.

Any idea what's going on?

treaves:
Sorry it’s taken so long to respond.

Those options would be used to set compiler flags for the various types of builds.  To build a debug version, you’d pass -DCMAKE_BUILD_TYPE=DEBUG to cmake.  Unfortunately, that conflicts with something in log4cpp.  So I removed all of the various compiler options, as I wasn’t sure the best way to handle it.  But, I left in that snippet about the build types to remember to go back in and ‘make it work’.

As Mac no longer uses GCC, and neither do several distro’s of Linux, setting the command line for the various build types should be fairly straight forward.  Once someone looks into the log4cpp thing.

PR #291, opened 2015-11-12
Clean up compiler warnings.

treaves:
- Updated use of abs() to std::abs() to match parameter types.
- in docbookmarks.cpp:156, change !col==1 to !(col==1) which I believe is the intended behavior.

### DIFF
--- a/applications/gqrx/receiver.cpp
+++ b/applications/gqrx/receiver.cpp
@@ -23,6 +23,7 @@
 #include <cmath>
 #include <iostream>
 #include <unistd.h>
+#include <cstdlib>
 
 #include <gnuradio/prefs.h>
 #include <gnuradio/top_block.h>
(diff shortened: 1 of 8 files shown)


applications/gqrx/receiver.cpp
```diff
@@ -23,6 +23,7 @@
 #include <cmath>
 #include <iostream>
 #include <unistd.h>
+#include <cstdlib>
```
csete: I think <cmath> is sufficient to have std::abs().

qtgui/dockbookmarks.cpp
```diff
@@ -151,7 +153,7 @@ void DockBookmarks::on_tableWidgetTagList_itemChanged(QTableWidgetItem *item)
     if(ui->tableWidgetTagList->m_bUpdating) return;
 
     int col = item->column();
-    if(!col==1) return;
+    if(!(col==1)) return;
```
csete: I have to check because I am not familiar with this part of the code. I would have written col != 1 so it could be intentional.

applications/gqrx/receiver.cpp
treaves: Corrected.

qtgui/dockbookmarks.cpp
csete: You are right. He meant col != 1, it just happened to also work the other way.

PR #292, opened 2015-11-12
Allow setting minimum FFT dB to plot

treaves:
Adds a slider & text field to the DockFft panel.  This allows for the modification of the minimum FFT dB to plot in the plotter class, which makes it easier to see a signal as the waterfall will be cooler in color and not display so much noise.


csete:
We can set the reference level by dragging the mouse on the Y-axis. Zoom happens using the mouse wheel / scroll on the Y-axis. The same applies on the X-axis, but slider was added here for convenience (some people did not have scroll capability).

treaves:
Ah.  I see.  That was not clear.  

One thing I thought about doing was writing some help files.  

I wonder if having this as an explicit control would still be desirable? This code would need to be modified to integrate with the dragging behavior.  Thoughts?

csete:
I'm fine with adding the slider if it is useful but we should perhaps also include a slider for setting the range and use one signal that sends both the min and max dB values.

I don't think any change is necessary in plotter.cpp since we already have the API:

    void setMaxDB(float max);
    void setMinDB(float min);
    void setMinMaxDB(float min, float max);

We can simply move these declarations to public slots.

treaves:
O.K., merged in upstream/master and added the minimum FFT dB slider as well.

PR #294, opened 2015-11-13
Support direct & band-based tuning

treaves:
Added widgets to the DockRxOpt panel that allows a user to click on a frequency band to tune directly to that band.  Also a numeric keypad to allow for directly entering a frequency.

There is better validation that could happen around this; as i'm new to Ham radio, I'm just grossly checking.  For example, the maximum allowed input frequency is 4.5 gHz.  I really don't know if that's the best value.


csete:
I'm sorry, but I don't think these widgets belong in the main user interface. They take up too much space and have in my opinion very limited use.

In general, I dislike the idea of taking elements from physical control panels and copying them into a computer user interface. Usually there are better ways to implement the same or similar functionality while taking advantage of the fact that it is a computer desktop.

The band selector is too specific for "ham radio on shortwaves" and so is irrelevant for most gqrx users. Keep in mind that gqrx supports many types of hardware and only a fraction of it covers shortwaves. Even on shortwaves there is much more than the ham radio bands.

The direct frequency entry could be useful, but clicking the frequency using the mouse feels wrong. A pop-up activated from menu/toolbar/shortcut that lets the user enter frequency using the keyboard would probably be more efficient. There has been a discussion about this not so long ago here but I don't remember the exact bug number.

PR #296, opened 2015-11-15
Draw lines to better show where plot area can be dragged on X & Y axis.

treaves:
This draws a line to the tight of the numbers on the x-axis, and just above the numbers on the y-axis.  The color is the same as the numbers, but with a lower alpha so they look good.

Also slightly adjusted were the x-axis scrolling bounds are.


csete:
The plan is to make the axis areas more distinguished.

PR #300, opened 2015-11-20
Fix use of std::complex variables

michaelld:
Fix use of std::complex variables, for both cleaner code as well as backwards compatibility to c++98.

I came upon this issue because the MacPorts buildbot for 10.8, which uses g++ 4.2, failed to build Gqrx because of the usage of std::complex variables. I hope I did the math correctly!

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1305,7 +1305,7 @@ void MainWindow::iqFftTimeout()
         }
 
         /* calculate power in dBFS */
-        pwr = pwr_scale * (pt.imag() * pt.imag() + pt.real() * pt.real());
+        pwr = pwr_scale * std::norm(pt);
         d_realFftData[i] = 10.0 * log10f(pwr + 1.0e-20);
 
         /* FFT averaging */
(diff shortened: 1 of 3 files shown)


csete:
Hello Michael,

I'm fine with the changes in decimator.cpp, although I am really puzzled that one can not assign real and complex values separately. However, I am going to revert the changes to using std::norm() because it uses std::abs() and is much slower than the direct multiplication in these cases.

You can try it for yourself with this simple code snippet:

```C++
#include <iostream>
#include <complex>

int main()
{
    std::complex<float>     num1;
    float                   num2 = 0.0;
    int i, j;

    for (i = 0; i < 10000; i++)
        for (j = 0; j < 10000; j++)
        {
            num1.real(2.3f * j);
            num1.imag(3.7f * i);
            //num2 += std::norm(num1);
            num2 += num1.real() * num1.real() + num1.imag() * num1.imag();
        }

    std::cout << "NUM: " << num2 << std::endl;
}
```

Using -O2 I measure a 6x speed penalty on x64 with gcc 4.8 and up to 40x penalty on ARM using gcc-4.9.

csete:
The changes have bee applied but didn't show up here, probably because I cherry-picked instead of merging. So, if there is nothing more, I think we can close it.

michaelld:
The assignment of real and complex separately is a GCC extension to std::complex; clang does not support it.

I had wondered if std::norm might do (std::abs()^2); pity that it does because it's so nice and clean in the code! No problems on not using it, since your code is then more efficient.

Thanks! I'll close the ticket now.

csete:
> The assignment of real and complex separately is a GCC extension to std::complex; clang does not support it.

Ok, thanks for the info. I will try to keep this in mind.

PR #305, opened 2015-12-07
Added description of the remote protocol.

kukabu:
Hello

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -1,0 +1,19 @@
+Remote control protocol.
+
+
+
+Supported command:
+ f - Get frequency [Hz]
+ F - Set frequency [Hz]
+ m - Get demodulator mode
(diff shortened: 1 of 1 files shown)


csete:
Thanks, but have you seen the existing documentation on the website?
http://gqrx.dk/doc/remote-control

If something is missing there I will be happy to update, but I would really prefer to have the documentation at only one place to make maintenance as easy as possible. Otherwise we risk having out of date documents one place.

kukabu:
Oh, I don't seen web documentation, but I think short description is need.
1. it's offline documentation.
2. it's can be included in binary packages for different distributions.

csete:
If you change the format to plain text and move it to resources/remote-control.txt, we can include it in the program under the Help menu, just like the news.txt file.

kukabu:
is it ok?

csete:
Yes, it is fine, thanks you. I have now included it under the Help menu.

kukabu:
Thanks!

PR #308, opened 2015-12-07
Added libgnuradio-pmt library dependency to build files

jynik:
Without libgnuradio-pmt, linking fails in dsp/rds/decoder_impl.cc due
to an undefined reference to pmt_make_blob().

Without libgnuradio-pmt, linking fails in dsp/rds/decoder_impl.cc due
to an undefined reference to pmt_make_blob().

I experienced this linking failure with:
- gqrx @ bbed2c95dcb7a222eebe495611a753a9ed5ff5de
- gnuradio @ d55fde354f257a1dd615a66c4ca98e3b6c6a5854

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ endfunction(add_source_files)
 # 3rd Party Dependency Stuff
 find_package(Qt5 COMPONENTS Core Network Widgets REQUIRED)
 find_package(Boost COMPONENTS system program_options REQUIRED)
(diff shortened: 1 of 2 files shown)


csete:
Thanks, I missed this one.

jynik:
@csete My apologies, it looks like I broke the .pro file by specifying gnuradio-pmt instead of gnuradio-runtime. 

I'll submit a PR in a few moments.

PR #312, opened 2015-12-14
cmake: Make pulseaudio (really) optional

andreamerello:
Introduce a cmake option to make it possible to disable pulseaudio on Linux build.

This patch wraps the related find_package and find_library cmake directives into a conditional block, while it does not remove the "required" attribute from them.

In this way:
- if the option is ON then pulseaudio is still mandatory, and the build fails when it is not found.
- if the option is OFF then pulseaudio is not even searched for (and eventually previously cached variables are cleaned).

I admit I'm not completely sure whether this is the standard/best way to achieve that, so RFC, please :)

As a side-effect of editing src/CMakeList.txt my editor has been kind enough to remove also few trailing space :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.0)
+cmake_minimum_required(VERSION 2.8.0)
 
 #######################################################################################################################
 # Project name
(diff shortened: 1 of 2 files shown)


csete:
Thanks for this patch.

If it's not too much trouble, could we call the cmake option something like AUDIO_BACKEND=pulse or AUDIO_BACKEND=gr-audio instead? On/off option suggests that one can enable them individually, while in this case it is either pulseaudio or gr-audio.

andreamerello:
Your point seems very reasonable to me; I theoretically agree with you.. But I'm unsure if there is a clean way to place multi-value selection in cmake options (If you are aware of any way to do this, please, tell me). I think that usually cmake uses string option without any constraint, but in this way the user could write anything in there (for example as it's done for build type).

Well, It seems also that there is indeed a trick to achieve something close to that, but at a first look it seems quite an ugly hack to me, and it doesn't not seems very robust.. Nevertheless I may give it a try :)
http://blog.bethcodes.com/cmake-tips-tricks-drop-down-list

If this hack won't work well, then I'm certainly open to change my patch using the "string" version. IMHO both my initial version and the "string" version are suboptimal, and I have no a strong opinion towards any of them.

BTW maybe the best way would be to allow even both backend to be enabled at the same time, and to eventually let the user decide in the GUI which one to use.. In this case we could have two ON/OFF options (both OFF can maybe still be useful to use the program as a spectrum analyzer, or maybe if it allows to save demodulated audio on file).. But I guess this would need also changes in the program GUI code..

csete:
Hmm, well I have no idea. I just thought it was trivial since we have options like -DCMAKE_BUILD_TYPE  which also takes a string.

Please take a look at it, but if it is too much trouble we can use it as it is in your original proposal.

andreamerello:
Yes, doing  as  -DCMAKE_BUILD_TYPE is actually easy, and it's exactly what I was talking about when I said "I think that usually cmake uses string option without any constraint" :)
But I thought that you was suggesting a kind of "true multi-value choice" (and that is also what I would really like to have). 

.. And in fact the tricky thing I was talking about was for making it a "proper" multi-value selection.. Something like changing the labels "ON" and "OFF" to "PULSEAUDIO" and "GRAUDIO" (so that it would be really a choice: the user couldn't write by mistake, let's say, "ALSA".).. But unfortunately this seems not trivial..

Well, if you feel it's OK, then I will go for the -DCMAKE_BUILD_TYPE way . As I said I have not a strong opinion about this whole thing: I just hate disabling pulseaudio by hand, anything else is fine :)

csete:
> Well, if you feel it's OK, then I will go for the -DCMAKE_BUILD_TYPE way . As I said I have not a strong opinion about this whole thing: I just hate disabling pulseaudio by hand, anything else is fine :)

Yes, that would be fine with me. If the user enters a wrong string it would default to pulseaudio, or perhaps an error message can be shown.

andreamerello:
Done. In the "wrong string" case I chose the "error message" way.

csete:
Very nice, thanks!

PR #320, opened 2016-01-01
Allow utilization of whole bandwidth

michaellass:
Currently the frequency bounds are applied to the receiver frequency. Instead they have to be applied to the hardware rf frequency to allow demodulation outside of this range as long as the selected frequency is within the available bandwidth. To allow using the whole bandwidth, this pull request contains two changes:

1. Remove the 90% limitation for the filter offset. This seems to be rather arbitrary and makes certain frequency ranges inaccessible although the necessary data is available.
2. Change calculations of frequency limits so that the LNB LO and the filter offset are taken into account.

Also see: https://groups.google.com/d/topic/gqrx/fIZrzhOlFfc/discussion

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -548,18 +548,14 @@ bool MainWindow::loadConfig(const QString cfgfile, bool check_crash)
     uiDockAudio->readSettings(m_settings);
 
     {
-        double      f1, f2, step;
-
         int64_val = m_settings->value("input/frequency", 14236000).toLongLong(&conv_ok);
 
-        // If frequency is out of range and ignoreLimits() is FALSE, set hardware
-        // frequency to the center of the range.
-        double hw_freq = (double)(int64_val-d_lnb_lo) - rx->get_filter_offset();
-        if (!uiDockInputCtl->ignoreLimits() &&
-                (rx->get_rf_range(&f1, &f2, &step) == receiver::STATUS_OK) &&
-                (hw_freq < f1 || hw_freq > f2))
+        // If frequency is out of range set frequency to the center of the range.
+        qint64 hw_freq = int64_val - d_lnb_lo - (qint64)(rx->get_filter_offset());
+        if (hw_freq < d_hw_freq_start || hw_freq > d_hw_freq_stop)
         {
-            int64_val = (qint64)((f2 - f1) / 2.0 + rx->get_filter_offset()) + d_lnb_lo;
+            int64_val = (d_hw_freq_stop - d_hw_freq_start) / 2 +
+                        (qint64)(rx->get_filter_offset()) + d_lnb_lo;
         }
 
         ui->freqCtrl->setFrequency(int64_val);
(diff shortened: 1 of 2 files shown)


src/applications/gqrx/mainwindow.cpp
```diff
@@ -668,8 +670,8 @@ void MainWindow::updateFrequencyRange(bool ignore_limits)
         qDebug() << QString("New frequnecy range: %1 - %2 MHz (step is %3 Hz but we use 1 Hz).").
                     arg(startd*1.0e-6).arg(stopd*1.0e-6).arg(stepd);
 
-        qint64 start = (qint64)startd + d_lnb_lo;
-        qint64 stop  = (qint64)stopd  + d_lnb_lo;
+        qint64 start = (qint64)(startd + rx->get_filter_offset()) + d_lnb_lo;
+        qint64 stop  = (qint64)(stopd + rx->get_filter_offset())  + d_lnb_lo;
```
csete: Ok, I think this is where the bug in question is but using the filter offset here is not correct either. I think we want:

qint64 start = (qint64)(startd - 0.5 * rx->get_input_rate()) + d_lnb_lo;
qint64 stop  = (qint64)(stopd + 0.5 * rx->get_input_rate())  + d_lnb_lo;

No?

src/applications/gqrx/mainwindow.cpp
```diff
@@ -777,6 +779,8 @@ void MainWindow::setFilterOffset(qint64 freq_hz)
     rx->set_filter_offset((double) freq_hz);
     ui->plotter->setFilterOffset(freq_hz);
 
+    updateFrequencyRange(uiDockInputCtl->ignoreLimits());
```
csete: This is no longer necessary c.f. my comment above.

csete:
> Remove the 90% limitation for the filter offset. This seems to be rather arbitrary and makes certain frequency ranges inaccessible although the necessary data is available.

Most SDR hardware have much reduced performance (some are even useless) outside ~80% of their passband. I have no problem with removing the limit but it will not make the extra range usable. I believe several other SDR application even crop the spectrum not to show the full range for the same reason.

src/applications/gqrx/mainwindow.cpp
michaellass: This was actually my first ideas as well. But this on its own would just allow setting a hardware frequency outside of the limits.

Currently, changing the main frequency dial changes the hardware frequency. To change the offset one has to use the designated dial or to click into the spectrum. If we allow full access to the whole spectrum through the main dial we would have to implement some means of automatically changing the offset when the chosen frequency is outside of the hardware limits. This would of course make this patch more invasive but in particular it would make the behavior of the main dial less predictable.

So I chose to keep the current behavior: To deviate from the hardware frequency one has to enter an offset frequency or click into the spectrum. Changing the main dial does not interfere with the offset in any way, so its min and max values have to depend on the currently set offset.

michaellass:
Regarding the 90%: I just started playing around with SDR and only have the HackRF One available here. I did a quick test to see if the outer areas of the captured spectrum are usable:

I set the sample rate to 20MSa/s, so bandwidth is 20MHz. Then I set the receiver offset set to -9.92MHz and tuned into a local fm radio station. Demodulating FM right at the edge of the spectrum was perfectly possible (with more narrow filter widths I could go even further).

The HackRF One probably is an exception in this regard and I understand the intention of the 90% limit now. Still I think the possible range should not be limited arbitrarily.

src/applications/gqrx/mainwindow.cpp
csete: I came to think about the same thing after I posted my comment but I'm still not happy with this solution.

It feels simply wrong that every time we change the filter offset, which could be 10-20 times per second, we call updateFrequencyRange(), which in turn calls rx->get_rf_range(), which calls an external API call that we have no control over. Besides, it always returns the same limit, so it is waste of CPU. On the other hand I realize that the same argument can be used against lnb_lo :-(

At the very least we should update the comment "This function must also be called when the LNB LO has changed." to include the filter offset.

csete:
The problem is aliasing. If you have a signal just outside your passband, you will see a mirror image of it just inside the passband. The amplitude of the mirror drops (hopefully quickly) the farther always you are from the edge. There is no point is working sop close to the edge if you can avoid it. 

But ok, I really don't care.

csete:
By the way, the 0.9 factor is also there for the plotter widget so that should also be changed then.

src/applications/gqrx/mainwindow.cpp
michaellass: That is a valid point... Currently I have no alternative idea how to fix this bug, though.
To avoid the API call we could buffer the rf range locally. Assuming it only changes when changing device settings... Or do I miss something here?

csete: That would work. The RF range should not change unless the device changes.

michaellass:
I'll rework this to address the points discussed. Thanks for the review!

csete:
I can see there is a new commit, but for some reason I haven't received any notification. Is it ready?

michaellass:
It should be ready. The hardware rf range is now only fetched after toggling ignore_limits (and therefore also on config load) and is set to "dc-to-light" if ignore_limits is true or the range cannot be determined.

PR #322, opened 2016-01-04
adjust clickable area to match nLevels changed in commit f9a33e12695c…

pcunning:
This is probably not the best way to fix this (as it already noted it needed to be fixed).
…96d7f06c6fc4e7597a624cad42aa


csete:
Thanks.

Issue #282, opened 2015-09-23
AOS command should open squelch or have ability to set squelch from remote control


khaytsus:
When I'm scanning and recording using my gqrx-scan script, I've found that if I have the squelch set too tight in GQRX I will sometimes record empty or choppy audio.

For example, my scanning script has detected a signal and I've set -65 is the threshold for it to send AOS to GQRX, but GQRX is set to -60 squelch.  In this case, it doesn't record any audio, only silence.

This happens often as different bands have very different squelch levels...  70cm I typically see about a -65db noise floor but 2m I often see more like -40db.  A workaround is to mute gqrx's audio output and set the squelch so it's always open, but that's not really ideal.

It would be nice if AOS either opened the squelch, or alternatively there was a way to set the squelch via the remote control TCP interface.

csete:
I think the right thing to do would be to add squelch level to the interface because I don't like the idea of AOS having a side effect. Can you try to add it? I am busy at the moment and don't know when I will have time to do it

khaytsus:
I started on this a bit a few days ago, but haven't done C++ in a few years, and never QT, there's a lot of glue to put in random places..  I've been following the pattern of change frequency.  If I do get something that works I'll fork and issue a pull request.  :)

csete:
Sounds great. If you have specific questions or get stuck just ask.

khaytsus:
I never got very far with this, I never could get all of the glue tied together.  I followed some very similar functionality but I just made more and more of a mess ;)

khaytsus:
Pretty sure #328 fixed this long ago :)

PR #328, opened 2016-01-21
Read and write squelch via remote control

alexf91:
Allows read and write of squelch level over TCP.
I also rewrote some of the remote control code to allow spaces in commands.
#282

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -7,7 +7,9 @@ Supported command:
  F - Set frequency [Hz]
  m - Get demodulator mode
  M - Set demodulator mode (OFF, RAW, AM, FM, WFM, WFM_ST, WFM_ST_OIRT, LSB, USB, CW, CWL, CWU)
- l - Get signal strength [dBFS]
+ l STRENGTH - Get signal strength [dBFS]
+ l SQL   - Get squelch threshold [dBFS]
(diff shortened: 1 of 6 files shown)


csete:
Please never mix several topics into one pull request. This way it is difficult to make a proper review.

The QString based command processing is likely to be much slower that the current char based processing. So I am going to need a very good reason to even consider that change.

So, please submit the squelch in its own pull request. When you do, please also update the resources/remote-control.txt file. I assume the command is compatible with the hamlib rigctld interface.

alexf91:
Ok, I'll split it.
I changed the chars to QString because it's easier to handle parameters for commands.
QStrings are already used in the modeStrToInt and intToModeStr functions, so I thought it would be okay to use them.

csete:
Well, maybe I am overreacting. I didn't take a thorough look at it. We can leave it like this and I can take a closer look at it in a few days or next week as I am too busy at the moment.

I would still like to have the remote-control.txt file updated.

csete:
Ok, I had a chance to review and test it and I think it is fine. However, I would kindly ask you to make one last change, which is to allow the command "l" without any option. I.e. the command "l" without STRENGTH or SQL parameter would default to STRENGTH and that way keep backward compatibility with existing applications.

csete:
Thanks.

PR #333, opened 2016-01-27
Tweaks to gqrx executable under windows

guruofquality:
* Build the gqrx icon into the executable under windows.
* Converted gqrx.svg into gqrx.ico for embedding the icon.
* Also build the executable as win32 app, not console,
  so it can be opened from executable without a terminal.


csete:
Nice, thanks,

PR #335, opened 2016-01-27
Created compatibility file resources for /dev/null + random

guruofquality:
Random and null file are used to support GQRX receiver source
with file resouces when actual devices are not available.
This commit adds functions get_random_file() and get_null_file()
to the receiver class to get a path to a random and zero file resource respectively.
The functions will try and use /dev/random and /dev/null when available.
Otherwise the function will generate a temp file to be used as a suitable replacement.


csete:
Thanks.

PR #344, opened 2016-02-19
Control audio recorder via TCP socket

alexf91:
Adds support to control the audio recorder via remote control.
The commands are compatible to rigctl, using the 'u' and 'U' (get and set_func) commands.

Enable recording with "U RECORD 1" and disable with "U RECORD 0".
The status of the recorder is returned by "u RECORD".


src/applications/gqrx/mainwindow.cpp
```diff
@@ -1654,6 +1655,8 @@ void MainWindow::forceRxReconf()
  */
 void MainWindow::on_actionDSP_triggered(bool checked)
 {
+    emit receiverStatusChanged(checked);
```
csete: We could simply call setReceiverStatus(checked) instead and save the declaration of the signal.

alexf91:
Ok, I wasn't sure if this causes problems.
I'm not that familiar with Qt, but it makes sense.

PR #351, opened 2016-03-01
Fixes #338 - Frequency allocations view

### LINKED ISSUE #338: Feature Request: Display Frequency Allocation for displayed frequency band
Provide a widget for displaying the frequency allocations for the displayed band.

I have a branch with the prototype here: https://github.com/ajmas/gqrx/tree/allocations-view

A few things need to investigated:
- is there interest from the general community or the author for having this in Gqrx?
- is a dependency on a website for the information acceptable?
- can this be made to work without needing to depend on a website?
- how to ensure data freshness
- how to have the band in the main plot communicated to the widget, rather than just the centre frequency.

This feature depends on https://github.com/ajmas/EarthFrequencies for now.

ajmas:
Provides a new allocation view dock window, which uses data from the open data "www.earthfrequencies.com" project, to display frequency allocations data as provided by the ITU and member entities. The feature is proposed under issue 338.


src/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ endif(WIN32)
 #######################################################################################################################
 # Build the program
 add_executable(${PROJECT_NAME} ${${PROJECT_NAME}_SOURCE} ${UIS_HDRS} ${RESOURCES_LIST})
-qt5_use_modules(${PROJECT_NAME} Core Network Widgets Svg)
+qt5_use_modules(${PROJECT_NAME} Core Network Widgets Svg Webkit WebKitWidgets)
```
alexf91: CMake fails for me unless I replace "Webkit" with "WebKit".

alexf91:
I would also remove the "band: xxx" output when the frequency is changed.
This fills the terminal window with information that is already available in the GUI.

When the network connection is lost, I get this messages:
> "Method \"GetAll\" with signature \"s\" on interface \"org.freedesktop.DBus.Properties\" doesn't exist\n"
> "Method \"GetAll\" with signature \"s\" on interface \"org.freedesktop.DBus.Properties\" doesn't exist\n"
> "Method \"GetAll\" with signature \"s\" on interface \"org.freedesktop.DBus.Properties\" doesn't exist\n"

ajmas:
Removed the stdout. Not seeing any network logging when I don't have a connection. What platform are you running on, so I can try to reproduce?

alexf91:
I'm using Arch Linux with Qt 5.5.1
Looks like the second message is Qt related: https://bugreports.qt.io/browse/QTBUG-49174
It's quite reasonable to think that the first one is too.

ajmas:
It looks like it could well be. If I get the chance I will see if I can set up a Linux VM to do some testing there. 

BTW do you have any feedback on the feature and on possible improvements?

csete:
Sorry, but I don't think this will happen.

Issue #363, opened 2016-04-14
UDP streaming dropping samples in OS X


a1k0n:
This is weird. I kept losing sync on a DMR stream... well, long story short, it appears the UDP socket is dropping samples. I became suspicious as DSD kept de-synchronizing, but if I saved a demodulated .WAV out and ran that through DSD, it worked fine. It *also* works fine to play back a raw I/Q stream -- this only seems to happen during live capture from my RTL-SDR, which is even more weird.

I don't think it's losing entire UDP packets. After a bit of investigation, it turned out to only be multiples of 16 samples. Here's what I did:

I started gqrx, tuned to a frequency with a narrowband FM demodulator, started `nc -u -l localhost 7355 >blah` in a terminal, and quickly hit the UDP and Rec buttons. I saved out about 20 seconds of 48kHz samples. I then converted the .WAV back to 16-bit raw and killed netcat, so I had two 16-bit raw samples which should be identical, but unaligned. The UDP stream was missing about a half second of samples at this point.

I then ran a correlation to align them at the beginning, and found this:
<img width="903" alt="screen shot 2016-04-14 at 9 39 43 am" src="https://cloud.githubusercontent.com/assets/46170/14535895/dd0a2558-0224-11e6-83a5-8438f1802d26.png">

After 10485 samples, the next 16 samples are missing from the UDP stream but present in the WAV. And after another 527 samples, it skipped 224 samples. I haven't discerned any particular pattern.

Also, it seems the rounding is slightly different between the two samples; sometimes the 16-bit values are off by one.

Config: OS X 10.10.5, gqrx 2.5.3 installed from website, RTL2838UHIDIR.

csete:
Thanks for reporting this. This feature has existed for a long time but there hasn't been much testing or feedback, so it is not unlikely that there are bugs.

The wav writer is a gnuradio block and has it's own float to signed integer converter, so that probably explains the rounding error. I may have forgotten to add 0.5 in the UDP sample converter or something like that.

As for the dropped samples, the only thing I can think of here and now is perhaps high CPU load prevents the UDP reader to get all samples from the receiver output, or perhaps the read buffer is just too small. I'm pretty sure that code could use a thorough review but unfortunately I don't know when I will get some time to do that.

csete:
By the way, this error could also explain why the built-in AFSK decoder is no longer working; it uses the same mechanism for getting samples out of the GNU Radio receiver.

a1k0n:
[...]

PR #367, opened 2016-04-21
Set UDP packet size limit to 512 bytes

a1k0n:
Workaround for #363. Substantially more reliable on OS X with this, though the reason packets are lost remains elusive.

Not honestly sure whether this should be user-configurable, or to change the default in GNU radio, or what. What do you think @csete ?

### DIFF
--- a/src/interfaces/udp_sink_f.cpp
+++ b/src/interfaces/udp_sink_f.cpp
@@ -50,7 +50,13 @@ udp_sink_f::udp_sink_f()
 {
 
     d_f2s = gr::blocks::float_to_short::make(1, 32767);
+#ifdef GQRX_OS_MACX
+    // There seems to be excessive packet loss (even to localhost) on OS X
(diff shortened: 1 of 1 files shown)


csete:
Thanks for the patch.

I'd like to avoid user configurable parameters as much as possible. If we wrap this change into

#ifdef GQRX_OS_MACX
    d_sink = gr::blocks::udp_sink::make(sizeof(short), "localhost", 7355, 512);
#else
    d_sink = gr::blocks::udp_sink::make(sizeof(short), "localhost", 7355);
#endif

then we don't break anything for existing users. We can in time test and see if it works just as well on linux and use the same packet size. If you resubmit with the ifdef or something similar I will merge it.

a1k0n:
Ok, #ifdef'd. Please take another look.

csete:
Excellent, thanks.

PR #373, opened 2016-05-05
Added some instruction for cmake prefix path

bogdanr:


csete:
If gnuradio and cmake were installed through the package manager of the OS, then cmake should be able to find gnuradio without having to specify the patch. If not, something is wrong with the packaging.

csete:
And if I am to make a guess, then it is the double gnuradio/gnuradio that causes cmake not to find the .cmake files. The .cmake files should probably be in /usr/lib64/cmake/gnuradio/

PR #380, opened 2016-05-21
Recognize input decimation during IQ recording/playback

### LINKED ISSUE #379: IQ recording/playback ignores input decimation
During recording and playback of IQ files, input decimation is ignored.
The recorded file name contains the original sample rate instead of the decimated one. This causes playback at the original sample rate.
Also, the time displayed is wrong by the decimation factor.

alexf91:
Fix #379


csete:
Thanks!

PR #390, opened 2016-07-05
[alsa-devices] Let ALSA users pick a sound card

Cat-Ion:
Currently ALSA users can't pick an output sound card. With this patch, it's possible to enter (and save) usual ALSA strings like "hw:0,0" to do so.

### DIFF
--- a/src/qtgui/ioconfig.cpp
+++ b/src/qtgui/ioconfig.cpp
@@ -153,6 +153,8 @@ CIoConfig::CIoConfig(QSettings *settings, std::map<QString, QVariant> &devList, 
     }
 
 #else
+    ui->outDevCombo->addItem(settings->value("output/device", "Default").toString(),
+                             settings->value("output/device", "Default").toString());
     ui->outDevCombo->setEditable(true);
 #endif // WITH_PULSEAUDIO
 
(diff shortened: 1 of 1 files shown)


src/qtgui/ioconfig.cpp
```diff
@@ -254,12 +256,20 @@ void CIoConfig::saveConfig()
 
     idx = ui->outDevCombo->currentIndex();
 
-    if (idx > 0)
+    if (idx > 0
```
csete: I would prefer not putting an #ifdef inside an expression. Perhaps something like this insted:

```
#if defined(WITH_PULSEAUDIO) || defined(GQRX_OS_MACX)
    if (idx > 0)
    {
        qDebug() << "Output device" << idx << ":" << QString(outDevList[idx-1].get_name().c_str());
        m_settings->setValue("output/device", QString(outDevList[idx-1].get_name().c_str()));
    }
#else
    if (idx > 0 || ui->outDevCombo->currentText() != "Default")
    {
        qDebug() << "Output device:" << ui->outDevCombo->currentText();
        m_settings->setValue("output/device", ui->outDevCombo->currentText());
    }
#endif
    else
    {
        m_settings->remove("output/device");
    }

```

csete:
Thanks.

Cat-Ion:
Thanks for merging :)

Issue #388, opened 2016-07-05
Small negative frequency offsets are displayed incorrectly


Cat-Ion:
If the frequency offset is negative and less than 1 kHz, e.g. -0.99 kHz, the widget in the receiver options shows it as 0.990 kHz instead.

csete:
Hmm, I thought there was already an issue about this because it has been a known issue since the beginning, but I can not find it. Perhaps we only discussed it on the mailing list. Anyway, I am hoping someone will send a patch fixing this.

csete:
Fixed with commit #86439028

PR #392, opened 2016-07-06
Draw small negative frequencies correctly

Cat-Ion:
This would be a fix for #388. Having to check this in the drawing method's a bit ugly, but I think we'd have to change the digitinfo struct to use a separate sign and value otherwise.

### DIFF
--- a/src/qtgui/freqctrl.cpp
+++ b/src/qtgui/freqctrl.cpp
@@ -707,8 +707,12 @@ void CFreqCtrl::drawDigits(QPainter &Painter)
             else
                 Painter.setPen(m_DigitColor);
 
-            Painter.drawText(m_DigitInfo[i].dQRect, Qt::AlignHCenter|Qt::AlignVCenter,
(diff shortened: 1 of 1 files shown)


csete:
It doesn't seem to be working correctly.

1. Start with offset 0 Hz.
2. Step down with 100 Hz steps -> shown offset is positive
3. At -1 kHz the minus sign appears.
4. Step up with 100 Hz step -> minus sign is shown correctly
5. After you cross 0 Hz and offset becomes positive, the minus sign is still shown.

Cat-Ion:
Oh, that must be because the 0 doesn't actually change, so it doesn't redraw it. I'll take a look, thanks.

Cat-Ion:
How about this? :)

csete:
It works now, thanks.

I merged it manually because I wanted to squash the commits into one commit. Keeping many intermediate commits makes sense when each individual commit is meaningful on its own. In this case all commits are necessary to fix the issue.

PR #396, opened 2016-07-21
converted tabs to spaces

tasnad:
I converted all leftover tabs in dsp/filter/decimator.cpp to spaces,
as my gcc v.6.1.1 gave a warning in line 73.


csete:
Thanks.

PR #409, opened 2016-09-17
When resetting FFT Zoom, reset slider too

pvachon:
Make sure that when the FFT zoom is reset using the reset button, that the slider UI element (and associated label) are also reset simultaneously.

This is a pretty trivial change, but it's a UI bug that has been making me crazy.


csete:
Thanks.

PR #416, opened 2016-09-27
Add button to lock pandapter and waterfall ranges

alexf91:
Adds a button to the FFT dock to lock the pandapter and waterfall ranges together.
When enabled, the value of the slider that was modified last is copied to the other slider.

### DIFF
--- a/src/qtgui/dockfft.cpp
+++ b/src/qtgui/dockfft.cpp
@@ -245,6 +245,12 @@ void DockFft::saveSettings(QSettings *settings)
     else
         settings->setValue("waterfall_max_db", intval);
 
+    // pandapter and waterfall locked together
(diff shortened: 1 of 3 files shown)


csete:
Thanks for the patch. It has been on my todo list to add this option, but I'm not sure about using a button.

It's often a bad idea to set an upper limit on the width of widgets since the sizes must scale according to the user's theme and screen resolution. Furthermore, on Mac OS X buttons must have a size at least 50x32 pixels, otherwise they will be treated a "different" kind of buttons. That's why buttons have a minimum size swet in the UI files and overridden in the constructor in case we are using linux.

So, I think we have to think a bit more about how to do this.

csete:
Ok, I think I can make it look right by tweaking the table layout. Merging...

PR #418, opened 2016-09-29
Bookmarks: sort tags too.

kukabu:


kukabu:
@csete, Sorry I found the ugly bug when added bookmarks. Please revert it.

PR #419, opened 2016-09-30
Remote control: state is saved now.

### LINKED ISSUE #350: Feature Request: save state of "Remote control"
Hello

Please, add status of "Remote control" into configure file

kukabu:
Fixes #350.

### DIFF
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -57,7 +57,8 @@ RemoteControl::~RemoteControl()
 /*! \brief Start the server. */
 void RemoteControl::start_server()
 {
-    rc_server.listen(QHostAddress::Any, rc_port);
+    if (!rc_server.isListening())
+        rc_server.listen(QHostAddress::Any, rc_port);
 }
 
 /*! \brief Stop the server. */
@@ -100,6 +101,11 @@ void RemoteControl::saveSettings(QSettings *settings) const
         return;
 
     settings->beginGroup("remote_control");
+
+    if (rc_server.isListening())
+        settings->setValue("enabled", true);
+    else
+        settings->remove("enabled");
 
     if (rc_port != 7356)
         settings->setValue("port", rc_port);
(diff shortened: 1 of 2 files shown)


src/applications/gqrx/remote_control.cpp
```diff
@@ -74,35 +74,55 @@ void RemoteControl::stop_server()
 /*! \brief Read settings. */
 void RemoteControl::readSettings(QSettings *settings)
 {
-    bool conv_ok;
+    if (!settings)
+        return;
 
-    rc_freq = settings->value("input/frequency", 144500000).toLongLong(&conv_ok);
-    rc_filter_offset = settings->value("receiver/offset", 0).toInt(&conv_ok);
```
csete: Why are you deleting these?

src/applications/gqrx/remote_control.cpp
```diff
@@ -74,35 +74,55 @@ void RemoteControl::stop_server()
...
+    settings->beginGroup("remote_control");
```
csete: Although the beginGroup() / endGroup() change is useful, they don't have anything to do with the topic and should therefore be a separate commit and pull request. Including them here only makes the present pull request more difficult to understand.

src/applications/gqrx/remote_control.cpp
kukabu: it's unnecessary when RC server running we get the current value from main code.

src/applications/gqrx/remote_control.cpp
kukabu: ok, I will separate commits.

kukabu:
ok, I try to find more simple way to start RC in main code.

kukabu:
@csete, is it ok?

csete:
I will take a look at it after v2.6 is out, hopefully by the end of this
week.

csete:
Looks fine, lets use this.

kukabu:
Thanks!

PR #428, opened 2016-10-03
Keep zoom-slider and waterfall zoom synchronized

alexf91:
Not perfect, because the plotter widget allows zooming well beyond the range of the slider, but they are kept synchronized within reasonable ranges.


csete:
> Not perfect, because the plotter widget allows zooming well beyond the range of the slider, but they are kept synchronized within reasonable ranges.

I think it is fine. People will rarely use zoom levels > 50x IMO. Otherwise we could limit the zoom level on the FFT plot.

Note that we have a shortcut to reset the zoom level: Right click on spectrum. It does not reset the slider. Probably an easy fix.

Issue #429, opened 2016-10-04
Remote control: only the last connection working


kukabu:
Hello

Run multiple the remote control connections to the gqrx. All connection are allowed and accepted but the gqrx works only with the last connection.

I tested with the gqrx 2.5.3 and git version.

alexf91:
Indeed.
There are a couple of ways to deal with this issue:
* Allow multiple connections
* Refuse new connections
* Close old connections

#430 provides a fix for this by closing old connections.

kukabu:
Thanks.
As workaround it's well but I think allow multiple connections is better way.

kukabu:
hm... refuse new connections as workaround is better, a new connection can be an accident or a mistake and in this case, we run the risk of breaking something.

alexf91:
Refusing new connections might cause some trouble if the old connection is lost without a proper close/shutdown of the socket.
AFAIK there's no reliable, portable way to detect a lost connection, except for sending something.
Many protocols use heartbeat messages to work around this problem, but Gqrx only responds to commands.

kukabu:
I think SO_KEEPALIVE with custom TCP_KEEPCNT/TCP_KEEPIDLE/TCP_KEEPINTVL will be enough it our case.

The GQRX is interactive application and user can be restarted server when found troubles (warning message may be useful hints for this).

csete:
Gqrx can only manage a single connection and the correct behavior is to ignore new connections as long as there is an active one. I was sure this is how it worked, but I may remember it incorrectly, or perhaps things have changed in time.

If a connection becoming inactive is a concern, we can add a timeout that checks that a command has been received during the last period (e.g. 1 minute), if not, it terminates the connection.

DjSteker:
In setting you give access to the remote IPEndPoint to the application, that is, to the IP of the application client, it works for me in the format ::ffff:127.0.0.1  "is LocalHosr"
::ffff:192.168.1.1   "is WAN "
::ffff:192.168.1.2   "is Server GQRX localhost"
::ffff:192.168.1.3  "is Cliente TCP" 
¿?

rag2:
DJ,

[...]

PR #430, opened 2016-10-04
Close old socket when a new connection is accepted

alexf91:
When a new connection was accepted, the old one was overwritten without closing it properly.


csete:
As I wrote in issue #429, I would prefer if we ignored new connections as long as there is an active one.

alexf91:
Ok. Maybe I have time on the weekend to implement it.

csete:
Sounds good and no hurry from my part. I need to spend some time on other projects also and my goal is to release a new version of gqrx by the end of the year, so there is plenty of time.

PR #432, opened 2016-10-07
Remote control: next patchset

kukabu:


src/applications/gqrx/remote_control_settings.ui
```diff
@@ -48,7 +48,6 @@
         <number>65535</number>
        </property>
        <property name="value">
-        <number>7356</number>
```
csete: Did you make this change in Qt Creator / Designer or did you just remove the line in the .ui file?

src/applications/gqrx/remote_control_settings.ui
```diff
@@ -88,7 +87,6 @@
        </property>
        <item>
         <property name="text">
-         <string>::ffff:127.0.0.1</string>
```
csete: Same comment as above. The .ui files should only be edited in Qt Creator / designer.

resources/remote-control.txt
```diff
@@ -1,21 +1,38 @@
 Remote control protocol.
 
 Supported commands:
- f - Get frequency [Hz]
```
csete: As I have said in the previous review, I do want to see changes in a commit that are not related to the topic in question. In this case the commit message says "Implement the M ? command" and so the commit should only contain that.

If you wish to reformat the text file, fine with me, but do it in a separate commit. I'm going to accept it this last time because it is a small change, but please remember it for the next time.

src/applications/gqrx/remote_control.cpp
```diff
@@ -429,19 +429,24 @@ void RemoteControl::setMode(int mode)
 void RemoteControl::setNewRemoteFreq(qint64 freq)
 {
     qint64 delta = freq - rc_freq;
+    // it's an ugly but I hope this will be fixed with the issue #218
+    int FilterBW_half = std::min<int>(bw_half, 100000 /* WFM Wide */);
```
csete: Sorry, but this doesn't work well with narrow band modes or when the total bandwidth is few 100 kHz.

src/applications/gqrx/remote_control_settings.ui
kukabu: I remove it without Qt Creator / Designer. I recommit it.

resources/remote-control.txt
kukabu: I added description argument '?' to 'M' command. Without reformate the text is ugly.

src/applications/gqrx/remote_control.cpp
kukabu: Sorry, but with this patch the gqrx behaves better. If the total bandwidth is small we sets zero always as in original code.
Try it https://gist.github.com/kukabu/2bad6584291b9da24a36c4b091d0d46d (or something similar) with and without patch.

csete: I will give it a try later when I have time, but I doubt I will accept it in its present form. It behaves differently and for sure it may be better in some scenarios, but it is worse in other scenarios.

csete:
Hi @kukabu 

I have already complained about that this pull request contains too many commits that are not directly related to each other. Now you have added even more commits. Was that by mistake?

If you want to submit new changes then:

1. Create a new branch
2. Make the changes and commit
3. Open a new pull request from that branch

I know it will be more work for you but small pull requests I can easily review and we can better discuss them.

I can see you made some updates and I will try to review them shortly, but this pull request has become quite convoluted and I may end up cherry picking manually. But please no more commits on this pull request :-)

kukabu:
Thanks.
Yes, it's my mistake, some commits may be added directly but some depends from other.
Now I will wait a review and don't pull a new commits in this issue.

csete:
Okay, I have cherry picked 4 commits into the master branch with slightly modified commit message. The rest is in the [kukabu-temp branch](/csete/gqrx/tree/kukabu-temp) for later reference. I didn't take them because I see some problems as follows.

Regarding  CW/CWR/q/Q - these commits change the interface potentially breaking it for existing applications and that should be avoided. Please note that 100% hamlib compatibility has never been the scope of this interface, so are these changes really necessary? I.e. is there an application that does not work with gqrx unless these changes are made?

The 'close' command was only intended for debug purposes when we use the interface manually e.g. through telnet. Applications should never need to use it since they can simply close the connection programmatically. So, is it necessary to add q/Q?

As for the CW and CWR, we can add them as aliases but I don't like that this introduces an inconsistency between mode set and mode read. For instance, if an application sets "CWU" then reads the mode gqrx returns "CWR" instead of "CWU". So again, do we really need these aliases or should we perhaps wait until there is an application that needs it?

Concerning the passband in the remote control, does it work together with the filter settings? I.e. when we change the filter in the UI? I also find the mixture of hi/lo and passband confusing and this is also what led me to think that it is not finished.

Feel free to resubmit the missing items after a rework, but please try to organize the pull requests according to topics. You can do that using different branches.

kukabu:
Thanks

About CW/CWR/q/Q. I have plan to implement the mode switching for gpredic then we must to follow to hamlib specification. Also I think all software written by hamlib specification must work with gqrx, why force people to remake his working code? And more, the gpredict uses 'q' command. :).

The passband patch works correct. I tested it for a few weeks. Yes, if we change the filter in the UI (via selector or mouse) gqrx returns correct value of passband and vice versa. The full passband uses when we set it. I don't find a beautiful way to get hi/lo in main RC code.

csete:
> I have plan to implement the mode switching for gpredic then we must to follow to hamlib specification

Ok, I understand. However, the preferred process would have been to wait until it is actually needed and then add the command with an explanation why it is needed.

> Also I think all software written by hamlib specification must work with gqrx, why force people to remake his working code?

Agreed, but do you know of a program that does not work with gqrx today? I never said people should rewrite their application, after all I chose a hamlib-compatible interface to be able to use existing aplications. All I'm saying is that I don't want to have 100% of the Hamlib API implemented if applications only use 5%.So let's stick to adding new API only as needed.

> The passband patch works correct. I tested it for a few weeks.

Ok, I couldn't see it since it was mixed with other changes. I will try to extract the commits and apply them.

csete:
Okay, I think I managed to merge the passband commits without any issues.

kukabu:
Thank!

Issue #362, opened 2016-04-14
Feature request: CTCSS and DCS codes.


DMChatterton:
It would further enhance Gqrx if identification of the above codes were to be provided for NFM signals.
Many communications receivers have this capability.

teward:
I'd love to see this as well - some things I monitor I want to ID the codes/tones on a frequency, and I wish this were a feature that was available here.

royldean:
Ok, this is a really crude method, but I've learned that if you display the "audio" interface (lower right corner), you can zoom in on the 0-250 Hz area and see a very course "spike" where the tone is located.   I don't think thee is anyway to increase the resolution to make it easier (or more precise), however.

PR #436, opened 2016-10-19
Hamlib compatible

kukabu:
I hope is it ok?


csete:
It's OK, but I don't understand why we have to remove 'c' ... hopefully nobody uses it.

kukabu:
Thank.
The 'c' can be used in the future #362.

Issue #438, opened 2016-10-21
Possible typo


kukabu:
Hello

The ";" at the end of the line looks like a bug.
https://github.com/csete/gqrx/blob/master/src/qtgui/plotter.cpp#L1339

Cat-Ion:
Looks like it's just supposed to find the first `level` with `tagEnd[level] <= x`. But the following line should probably be indented better, and maybe replaced by `if(level == nLevels) { level = 0; }` for clarity.

kukabu:
maybe, but I'm embarrassed the indented in next line.

csete:
Fixed by #439

kukabu:
Thank.

PR #439, opened 2016-10-21
Cosmetic changes

Cat-Ion:
Addresses #438

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1336,8 +1336,11 @@ void CPlotter::drawOverlay()
 #endif
 
             int level = 0;
-            for (; level < nLevels && tagEnd[level] > x; level++);
-                level %= nLevels;
+            while(level < nLevels && tagEnd[level] > x)
+                level++;
+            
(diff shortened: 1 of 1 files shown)


csete:
Good idea to improve readability of the code, but why not go all the way and replace that confusing if statement with a while loop? Also, the comment seems the opposite of what it should be, no?

I would also much prefer if the commits were squashed into one commit.

Cat-Ion:
Replacing it with a while loop? I'm not sure I understand. Are you talking about the replacement for the modulo? But yes, the > should be replaced with <=, of course.

I... can try to do a force push to replace the commits? I've never done this before, but I'll try. Worst case we'll just close the PR and make a new one =P

csete:
```
while (level < nLevels && tagEnd[level] > x)
    level++;
```
Easier to read, no?

I'm okay with a new pull request if it works out better. Would much prefer one commit rather than three for this.

Cat-Ion:
Oh, you meant the for loop. Yeah, that's probably better.

Cat-Ion:
The push worked! Hooray. :)

PR #440, opened 2016-10-24
The 'm' command must returns a passband on new line.

kukabu:
Sorry. That's my fault.

```
Rig command: m
netrigctl_get_mode called
write_block(): TX 2 bytes
0000    6d 0a                                               m.              
read_string(): RX 3 characters
0000    46 4d 0a                                            FM.             
read_string(): RX 6 characters
0000    31 30 30 30 30 0a                                   10000.          
Mode: FM
Passband: 10000

Rig command: 
```

### DIFF
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -284,7 +284,7 @@ void RemoteControl::startRead()
     }
     else if (cmdlist[0] == "m")
     {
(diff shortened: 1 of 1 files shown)


csete:
By the way, shouldn't the passband always be positive, i.e. in LSB mode it will return a negative number.

kukabu:
Thank.

PR #441, opened 2016-10-24
Reimplement a hamlib compatible layer.

kukabu:
Hello

I think that you doesn't accept the pull #440 because it's break compatible.
This is a better fix for it.


csete:
We already changed when we introduced the passband parameter. In any case, I would prefer that than this change which IMO would make the code more and more confusing.

PR #444, opened 2016-10-28
Refactoring RemoteControl::startRead().

kukabu:
This function is very large and unclean.
Now one command - one function.


csete:
Looks good, thanks.

PR #453, opened 2016-11-18
Add .travis.yml for Travis CI (http://travis-ci.org/).

### LINKED ISSUE #451: Travis CI
Hello

I think it should be useful to add Travis CI integration.

kukabu:
Fixes #451.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+sudo: required
+
+dist: trusty
+
+language: cpp
+
+compiler:
+  - gcc
+  - clang
+
+env:
+  - AUDIO_BACKEND="Pulseaudio"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,28 @@
+sudo: required
+
+dist: trusty
+
+language: c
```
csete: Shouldn't this be cpp?

csete:
I'm going to merge it, but I'd like to point out that I still lack info about how this is useful. All I see is a configuration file that will perform a build almost the exact same way I do on my own computer before committing the code. So, what's the point?

On the other hand, we now have yet another file to keep up to date, yay!

Issue #393, opened 2016-07-10
Feature idea: Waterfall scaling, changing colors when changing squelch level, ...


Cat-Ion:
Keeping some of the old FFT data around for the waterfall would have a bunch of advantages. It would let us zoom around even for old data, and changing the min/max dB values could be applied to old data as well.

The obvious problem might be memory usage. Maximum would be 4 MiB per line for an FFT of size 2^20. For a fullscreen window on a 1080p monitor this might use some 3 GiB of RAM, if the waterfall display uses 2/3 of the screen. Could halve that by using e.g. a 16-bit int to store dB levels instead of full floats. On the other hand, PCs that do FFTs on a million samples at once might just be beefy enough to spare those 3 GiB anyway.

Each line would have some metadata associated with it, probably min/max frequency, a timestamp, and the FFT size. A ringbuffer sized by the number of lines in the waterfall should be appropriate for storing that, plus pointers to the FFT data, and not incur much overhead when checking for and deleting old data.

Performance shouldn't change, as the data would only have to be accessed when scaling the waterfall in some way.

I know @csete is too busy, but I wouldn't mind working on this when I get some time if I'm not the only one interested in this :)

csete:
I have a few ideas of my own, which would require double buffering and I think what you propose also requires it.

Currently, the waterfall has 8-bit resolution, which I believe is sufficient. Most common SDRs supported by gqrx have a dynamic range < 120 dB, so an 8-bit palette will still give better than 0.5 dB resolution.

Cat-Ion:
Yeah, but the dynamic range of gqrx is larger, going from 0 dB down to at most -300 dB, so that's about 1.2 dB resolution. If you use a small 10 dB range (which are admittedly rather grainy even without an intermediate step) you'll be down to just 8 steps. Might be worth just testing out, or leaving it configurable (at least with a #define). We might be able to use 10-bit samples or so, but would have to check the overhead of all those bitshifts.

csete:
Ok, you lost me here, but I'm pretty sure that gqrx does not have 300 dB dynamic range.

Cat-Ion:
You can set the reference level to anywhere from 0 to -100 dB, and the dB range can be 10 to 200 dB.

Cat-Ion:
oh, what we might be able to do is scan each FFT block and store the min/max and rescale to 0-255, so we'd have

    struct FftSample {
        qint64 freq_min, freq_max;
        qint64 timestamp;
        qint64 size;
[...]

PR #461, opened 2016-12-13
Waterfall caching and redrawing

Cat-Ion:
I've been doing some work on #393 and it's in a mostly working state. I still don't have a way of limiting used memory right now, but at least the backend part of that should be simple enough. You do get some visual artifacts when moving the waterfall, but that can't be avoided with the 8-bit storage for FFT bins.

Memory overhead right now is 1 byte per FFT sample, over the height of the waterfall, plus 32 bytes per line.

I've messed with the plotter class for a bit, mostly by putting draw() into a few smaller functions, and also optimized its drawing function to be much faster, as it was a major bottleneck for the redrawing.

While it probably shouldn't be merged at this stage (still need the max memory stuff), I'd love some feedback and maybe tests by other people.


csete:
Thanks @Cat-Ion -- I will take a look at it, though it'll probably be several weeks before I get a chance. I'm in the middle of moving and my backlog is already very long.

Cat-Ion:
Sure, no problem. As you can see, I still have lots to do. :)

Cat-Ion:
An overview of the changes I made:
- created FftBuffer class that stores FFTs as arrays of 8-byte values, along with frequency and power information. It uses a ring buffer with a size set to the height of the waterfall.
- Split up the somewhat large `CPlotter::draw()` function into `drawWaterfall()` and `draw2DSpectrum()`
- `drawWaterfall()` accumulates waterfall data (`accumulateWaterfallData()`), updates the buffer, scrolls the pixmap, and calls `drawWaterfallLine()`
- `drawWaterfallLine()` uses the `CPlotter`'s FFT array for the upper line or the buffer for all older lines, gets the appropriate data, and dumps it into an RGB buffer, which is drawn onto the waterfall pixmap.
- Changing the waterfall's axes now calls the new `redrawWaterfall()`, which redraws each line.
- removed `CPlotter::m_wfbuf` - using the RGB buffer now.
- added `CPlotter::m_wfAccum` for accumulating data as floats to match the format expected by the FFT buffer.

Advantages:
- The waterfall can be redrawn when you change the center frequency or spectral power limits.
- Having a few smaller functions instead of one large drawing function is easier to handle, IMO. At least I found it rather unwieldy when I was working on it, hence the splitting up.
- Drawing into an RGB buffer first is a huge performance gain.

Current disadvantage: Redrawing full large FFTs (above 32k or so) gets a bit choppy. This should get faster when I implement partial redrawing.

alexf91:
I know there are several issues addressing this pull request (e.g. #469) , but I'll just comment here.

I've been using this for a while now, and it works fine. I didn't see any crashes or other errors.
There are some minor performance issues with large FFT sizes, but these are acceptable on a regular PC.
However, some users are running gqrx on a Raspberry Pi or similar, so this might be a problem on such systems. I think making this feature configurable would be the way to go. From what I see, this is not hard to implement.

Regarding the code: Is there a reason why you are using malloc/free instead of new[]/delete[]?

PR #462, opened 2016-12-13
Implement Synchronous AM detection

gat3way:


csete:
Thanks @gat3way -- I'll try to take a look at it within a few weeks.

After a quick glance I'm a it concerned though about changing the numerical values of the demodulators. Please make sure that we keep backwards compatibility in the config file. We could probably also add AMS using a flag to the existing AM mode.

gat3way:
No probs, I will reimplement it as a demod option, I was going to refactor the dsp part anyway as the pll_trackingfilter block just mixes the recovered carrier with the input signal with no way to apply gain/filter prior to that which might be suboptimal. Will close that pull request and submit another when I am done with it (hopefully in 1-2 days)

kukabu:
I think it better way than AMS flag. Yes, we must be keep backwards compatibility but it's easy.
if you follow this logic, then the LSB/USB is SSB mode with flags, WFM STEREO, MONO, OIRT is WFM with flags and etc.

Also we have to add signal/slot to set AMS flag in remote control code, it makes code more difficult and confusing.

kukabu:
Uh, I forgot about bookmaks then even more spaghetti code

csete:
Initially, I'm fine with either way as long as we don't break things. In any case, we should probably continue discussion on PR #463 since that's the active one and this one has been discarded.

PR #463, opened 2016-12-14
AM-Sync implemented as AM demod option

gat3way:
Hello,

Left the mode enums intact and implemented synchronous AM detector mode as a demod option, hope it's OK now

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -199,6 +199,7 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
     connect(uiDockRxOpt, SIGNAL(fmMaxdevSelected(float)), this, SLOT(setFmMaxdev(float)));
     connect(uiDockRxOpt, SIGNAL(fmEmphSelected(double)), this, SLOT(setFmEmph(double)));
     connect(uiDockRxOpt, SIGNAL(amDcrToggled(bool)), this, SLOT(setAmDcr(bool)));
+    connect(uiDockRxOpt, SIGNAL(amSyncToggled(bool)), this, SLOT(setAmSync(bool)));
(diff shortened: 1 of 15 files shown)


csete:
Thanks for the update! I will do my best to review and test it in the coming weeks.

gat3way:
I did some (kind of controversial) improvements, most notably we now filter before the PLL refout block and I tweaked the loop bandwidth and the filter cutoff a bit to have a much more stable PLL lock. The end result being we now have a noticeably better AM-Sync mode, this is a real improvement, just as it has to be. The end result being it considerably helps with selective fading and we have a much better result even with weak signals. On the negative side, users now have to tune very precisely, as larger offsets from station's carrier frequency will get the carrier off the filter and we also need much more CPU power to satisfy the filter. Users with slower CPUs may have issues with that. That said, ideally speaking probably two or more am-sync modes need to be introduced where we compromise between demodulated audio quality and CPU resources/tuning offset.

kukabu:
I think it's a bad way. This way complicate the code for bookmarks and remote control.
with PR #462 way (AMS as mode)
- bookmark - must be works without any modification
- remote control - requires small changes in four functions

csete:
Good points @kukabu, perhaps it's best to add it as a new mode. In that case we should add it to the end of the list to keep backwards compatibility.

gat3way:
I am a bit confused, at the end we are doing that as a new mode or as a demod option?

csete:
Sorry for the confusion @gat3way. I think it's best to add it as a new mode appending it to the end of the list, otherwise it will be difficult to integrate it with bookmarks and remote control. Is it OK with you?

gat3way:
No problem, will do that in the following days.

csete:
Great, thanks. There is no hurry from my part as it will still be a few weeks before I am up and running at my new place.

csete:
It doesn't look like this will happen. Feel free to resubmit if you still wish to work on it.

PR #465, opened 2016-12-18
Added AMS mode to remote control.

kukabu:
Hello

@csete I wrote this patch because the '\dump_state' contains some magic numbers that may cause difficulty.
@gat3way I assuming that the https://github.com/csete/gqrx/blob/master/src/qtgui/dockrxopt.h#L66 will be 11 for AMS mode.

### DIFF
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -395,7 +395,7 @@ bool RemoteControl::setGain(QString name, double gain)
  *  \param mode The Hamlib rigctld compatible mode string
  *  \return An integer corresponding to the mode.
  *
- * Following mode strings are recognized: OFF, RAW, AM, FM, WFM,
+ * Following mode strings are recognized: OFF, RAW, AM, AMS, FM, WFM,
  * WFM_ST, WFM_ST_OIRT, LSB, USB, CW, CWL, CWU.
  */
 int RemoteControl::modeStrToInt(QString mode_str)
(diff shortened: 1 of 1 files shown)


csete:
No longer relevant as pull request #462 has been closed

kukabu:
@argilo Hello. I think now it's can be added.

argilo:
@kukabu You're correct that AMS should be added to the remote control protocol, so I'll re-open this.

I don't understand the changes to `cmd_dump_state`. Can you explain those in more detail?

src/applications/gqrx/remote_control.cpp
```diff
@@ -328,9 +328,6 @@ void RemoteControl::setReceiverStatus(bool enabled)
 /*! \brief Convert mode string to enum (DockRxOpt::rxopt_mode_idx)
  *  \param mode The Hamlib rigctld compatible mode string
  *  \return An integer corresponding to the mode.
- *
- * Following mode strings are recognized: OFF, RAW, AM, FM, WFM,
- * WFM_ST, WFM_ST_OIRT, LSB, USB, CW, CWL, CWU.
```
argilo, line 333: Why remove this?

argilo:
Ah, I now see that AMS is defined here:

https://github.com/N0NB/Hamlib/blob/5ae848c48f1810fa81deb8365cb89e552740c8b7/include/hamlib/rig.h#L1121

src/applications/gqrx/remote_control.cpp
kukabu: Sorry, I forgot why, but yes, it's not be removed

kukabu:
The dump_state isn't documented. I researched the hamlib source code and tried to describe the fields in the comments as detailed as possible.

argilo:
I pushed an updated version to your branch. Let me know how it looks.

kukabu:
Thanks. Looks fine.

argilo:
Thanks for checking. I also tested this, and CI passes, so I'll merge now.

PR #470, opened 2016-12-28
Replace deprecated mpsk_receiver by constellation_receiver

pinkavaj:
Hi,

mpsk_receiver is now deprecated in master branch of GnuRadio and removed from next branch. This patch replaces mpsk_receiver by apropriately configurad constellation_receiver.

### DIFF
--- a/src/dsp/rx_rds.cpp
+++ b/src/dsp/rx_rds.cpp
@@ -25,8 +25,7 @@
 #include <cmath>
 #include <gnuradio/io_signature.h>
 #include <gnuradio/filter/firdes.h>
-#include <gnuradio/digital/mpsk_receiver_cc.h>
-#include <gnuradio/blocks/complex_to_real.h>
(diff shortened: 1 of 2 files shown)


csete:
Thanks @pinkavaj, I'll take a look at it next time I'm working on gqrx.

Are you aware of any compatibility issue with the current GNU Radio 3.7? Also, please consider submitting this change to the upstream project also, unless they have already implemented it. Thanks again!

pinkavaj:
There should not be compatibility issues with GnuRadio 3.7, but I does not have simple way to check it.

I'm a bit confused about the upstream project, what do You mean by that, what project, You are not upstream of GQRX?

csete:
> I'm a bit confused about the upstream project, what do You mean by that, what project, You are not upstream of GQRX?

I meant the gr-rds upstream project, which as I recall was https://github.com/bastibl/gr-rds but they may be too different now.

csete:
Looks like it works fine. Thanks for the update.

Issue #471, opened 2017-01-03
Auto-tuning through remote control may need tweaking


csete:
https://groups.google.com/d/msg/gqrx/e7BuIWJ7W1w/vSwmIcGRDAAJ

@kukabu if you have time you are welcome to see if there is anything we need to tweak here.

kukabu:
Hello

I think I can fix it but only after holidays.

csete:
Sounds great. No hurry from my part. Enjoy your holidays!

PR #475, opened 2017-01-09
F471

kukabu:
Hello

Yet, I don't see another beautiful solutions.


csete:
Ok, I will take a look at it when I have some time, but would you please try a bit harder to write a meaningful title and description, i.e. something that actually describes what the pull request is about.

LongnoseRob:
@kukabu @csete 
I just pulled the PR to my fork of gqrk (which was synced before with csete:master) and tested it a little bit.
It works, when tracking sats with my airspy mini, gqrx and gpredict, gpredict does not make gqrx tune to the outer limits of the spectrum, where the falloff of the receiver is.
1. However I found that a  default setting of 80% for BW_WIN_VALUE works better with my setup.
2. the remote command u/U is already used for rremote control of the recording feature of gqrx, we should find a differnet way of setting this value by remote.
3. there should be a way to set the BW_WIN_VALUE in the GUI.

csete:
Thanks for the info. I hope to find some time soon to test this issue.

kukabu:
@LongnoseRob u/U is get_func/set_func as say in the hamlib documentation and I think we can use it for custom command.

csete:
I attempted to address issue #471 in commit 950996a4787274b491db92a930b0a1637e19ac9d. I don't know if there were any other features hidden in here, if yes, feel free to re-submit them separately.

PR #476, opened 2017-01-09
Added a few fake command for better hamlib compatible.

kukabu:
Hello

It's patch added some fake commands for better compatible with hamlib stuff.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -28,7 +28,17 @@ Supported commands:
  LOS
     Loss of signal (LOS) event, stop audio recording
  \dump_state
-    Dump state (only usable for compatibility)
+    Dump state (only usable for hamlib compatibility)
+ v
+    Get 'VFO' (only usable for hamlib compatibility)
+ V
+    Set 'VFO' (only usable for hamlib compatibility)
+ s
+    Get 'Split' mode (only usable for hamlib compatibility)
+ S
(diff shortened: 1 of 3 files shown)


csete:
I have said countless times that 100% hamlib compatibility is not the goal with the remote interface and that I do not want to bloat the code with adding anything more that what we really need. The only other option is if there is an existing application that does not work unless we implement something.

So, why do we need these? 

PS: This is the last time I am saying this and asking explicitly why we need the new API. In the future I will simply discard or ignore such pull requests if they do not include a proper explanation why the new API is added.

kukabu:
It's the minimum necessary for hamlib stuff.
Try it

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import sys
import Hamlib

def StartUp ():
    Hamlib.rig_set_debug (Hamlib.RIG_DEBUG_NONE)
    my_rig = Hamlib.Rig (Hamlib.RIG_MODEL_NETRIGCTL)
    my_rig.set_conf ("rig_pathname","127.0.0.1:7356")
    my_rig.open ()
    my_rig.set_freq (Hamlib.RIG_VFO_A, 145550000)
    my_rig.close ()

if __name__ == '__main__':
    StartUp ()
```

This script should set the 145.550MHz but not.

csete:
Ok, thanks for the info.

PR #477, opened 2017-01-09
Added a hidden configure option "gui/ResetLowerDigits" for the old behavior.

### LINKED ISSUE #445: Frequency widget digits set to 0 on retune
(git version)
 Changing frequency 10ths or 100ths MHz sets lesser digits to 0 
I don't know if this is intended feature or some bug.
AFAIR earlier versions behaved different - all lesser digits remained unchanged.
 F.e.: Let's tune to the middle of ham band 14.05MHz and change MHz to 21MHz -  I was used that gqrx tuned to 21.05 but now it's tuned to 21MHz.
 IMHO (as a ham) leaving them unchanged is preferred way,

kukabu:
Fixes #445

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -588,6 +588,7 @@ bool MainWindow::loadConfig(const QString cfgfile, bool check_crash,
             on_plotter_newFilterFreq(flo, fhi);
         }
     }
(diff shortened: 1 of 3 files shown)


csete:
Ok, adding a hidden config option could be the way to resolve this, but I don't like the implementation. also, I'd like to have some config record documentation set up before we start introducing hidden config records.

csete:
Can you please see if commit e7f3c94cda57296c1c8784d261588c5e160e0903 is sufficient? Thanks.

kukabu:
Thanks.

PR #487, opened 2017-02-16
Add checkbox to select frequency control widget reset behavior

alexf91:
Issue #445.

Adds a checkbox to select the behavior of the frequency control widget from the GUI.

I didn't really know how to name the button and where to put it. It's currently in the input-control dock and called "Reset frequency control".


csete:
Thanks. I was thinking about making it a menu item since it is a rarely used option that will just take up space in the UI. I am not 100% sure though...

What do you think?

alexf91:
I considered this as well, but I though this would be overlooked, because all other settings are in docks.
I would leave it in a dock and, if more GUI settings are added, make a separate dock or settings-window for these options.

csete:
All right, lets do this. I can always regret it later :-)

PR #495, opened 2017-03-02
Check if new and old configuration file are equal

### LINKED ISSUE #493: .conf files are deleted when changes are saved.
Using git master.

Steps to reproduce:

1. make some changes and save them to atest.conf
2. ls -l ~/.conf/gqrx
3. make some changes and save them to atest.conf
4. when asked to overwrite answer yes
5. ls -l ~/.conf/gqrx

If I can figure it out I will send a PR.

alexf91:
Fixes #493 
Keeps the old behavior of syncing before "save as", but checks if old and new configuration files are equal.


alexf91:
AFAIK QSettings objects don't allow changing the filename.
With this fix, we just check if the configuration is saved to the configuration file currently in use and sync to disk if this is the case.

csete:
Ok, thanks!

PR #499, opened 2017-03-14
Disable detected proxy settings

### LINKED ISSUE #497: GQRX doesn't listen for remote control connections if $http_proxy is set.
It's a weird one. I'm using GQRX on linux and it doesn't listen on control socket if launched from an environment that has http_proxy set.

alexf91:
Should fix #497 
Not sure if this should be merged, because it disables the proxy detection used by Qt. I'm not even entirely sure why this caused problems in the first place.

### DIFF
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -25,6 +25,7 @@
 #include <iostream>
 #include <QString>
 #include <QStringList>
+#include <QtGlobal>
 #include "remote_control.h"
 
 #define DEFAULT_RC_PORT            7356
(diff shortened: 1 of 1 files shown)


csete:
I'm confused... Are you saying that the reason why "gqrx does not listen" on a system with proxy is because Qt detects the proxy settings automatically? This makes no sense to me.

src/applications/gqrx/remote_control.cpp
```diff
@@ -50,6 +50,9 @@ RemoteControl::RemoteControl(QObject *parent) :
     rc_socket = 0;
 
     // Disable proxy setting detected by Qt
+    // Workaround for https://bugreports.qt.io/browse/QTBUG-58374
+    // Fix: https://codereview.qt-project.org/#/c/186124/
+    // Can be removed once Qt 5.9 is available
```
csete: We should probably wrap this in a Qt version check since there will be many years where 5.9 will be available alongside with many legacy versions distributed with linux.

alexf91:
Should be ok now.

> there will be many years where 5.9 will be available alongside with many legacy versions distributed with linux.

Unless everybody switches to Arch ;-)

PR #515, opened 2017-04-27
Add additional waterfall time spans

mhostetter:
I have really enjoyed the fixed time span waterfall feature. Although, I have found a desire/need for a few smaller time scales. This PR adds those.

**NOTE:** I'm not particularly familiar with this code base. So please let me know if I made any oversights. I did test this and everything seems to work fine.

### DIFF
--- a/src/qtgui/dockfft.cpp
+++ b/src/qtgui/dockfft.cpp
@@ -364,6 +364,8 @@ void DockFft::on_fftRateComboBox_currentIndexChanged(const QString & text)
 static const quint64 wf_span_table[] =
 {
     0,              // Auto
+    1*60*1000,      // 1 minute
+    2*60*1000,      // 2 minutes
     5*60*1000,      // 5 minutes
     10*60*1000,     // 10 minutes
     15*60*1000,     // 15 minutes
@@ -381,7 +383,7 @@ static const quint64 wf_span_table[] =
 /** Waterfall time span changed. */
 void DockFft::on_wfSpanComboBox_currentIndexChanged(int index)
 {
-    if (index < 0 || index > 12)
+    if (index < 0 || index > 14)
         return;
 
     emit wfSpanChanged(wf_span_table[index]);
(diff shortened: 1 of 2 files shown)


csete:
Thanks for the patch.

The changes look fine, but I don't quite understand the purpose. Can't you achieve similar time spans by adjusting the FFT rate? To be honest, I'm not even sure the algorithm works properly when the time span is shorter than the real-time span.

I don't mind adding more options but I prefer not having any options that have no effect or are redundant.

mhostetter:
That's what I thought of first, changing the FFT size and FFT rate, but couldn't get it to do exactly what I wanted. When in time span `Auto`, GQRX doesn't indicate what is the time scale of the pandapter. It just says `Res: -s`.

Secondly, what does the rate mean? Is `60 fps` 60 FFT lines or rows on the pandapter per second? If so, I would need to know how many rows there are to know the time scale.

I'm open to other solutions. In my case, `Auto` mostly works but sometimes I want a little more time duration on the pandapter, and 5 minutes is a bit too much.

csete:
Rate means how often FFT is redrawn on the screen. So, 10 FPS means the image is redrawn 10 times per second. If "Time span" is set to "Auto", each redraw generates a new line on the waterfall.

If the height of the waterfall is 600 pixels, then 10 FPS will correspond to 60 sec time span. If you select a longer time span, then two or more FFT lines will be combined into one line, but it doesn't work the other way. If it takes 60 seconds to fill the waterfall in real time, selecting 15 seconds will have no effect.

So, 1 and 2 minutes are OK I guess, but the 15 and 30 seconds choices will in most cases have no effect.

mhostetter:
Ok, thanks. Would it be acceptable for me to change the PR to just add the 1 and 2 minute options?

csete:
Yes, please do that and I will merge.

mhostetter:
I removed the 15 and 30 second options. Thanks for the feedback.

Issue #538, opened 2017-08-19
IQ file navigation error in large files


dk7ob:
Navigating in IQ file is faulty in large IQ files due to an integer overflow.
The attached patch fixes this.

[0001-IQ-rec-play-Fix-navigation-in-large-IQ-files.patch.txt](https://github.com/csete/gqrx/files/1236183/0001-IQ-rec-play-Fix-navigation-in-large-IQ-files.patch.txt)

csete:
Thanks for the patch. Merged it through the pull request by @alexf91

PR #539, opened 2017-08-19
IQ rec/play: Fix navigation in large IQ files

alexf91:
Fixed a play navigation error in large IQ files due to integer overflow.

Signed-off-by: Wolfgang Fritz <removed email address>

PR for #538


csete:
Thanks for the PR

Issue #547, opened 2017-08-26
Feature Request: Sent LNB LO via remote control


sdonnan:
I'm currently using the GQRX software with an RTL-SDR dongle as a pandapter for my radio. This setup works quite well. I set the LNB LO frequency in the configuration dialog for the first I/F of my radio, and then once I am running I can use the LNB LO display frequency adjustment in the Input Controls panel to compensate the displayed frequency on the UI.

The issue is that when I change the tuning of my radio I must manually recalculate and enter the display LNB LO frequency for the new config. This would be straightforward to automate by polling over the CAT interface to the radio and then setting the display LNB LO in GQRX via the remote interface. I am requesting the addition of a remote command to implement this functionality.

Note that I do not want to change the LNB LO used for tuning the RTL-SDR, I just want to change the LNB LO used for calculating the displayed frequency. Not sure what the best terminology is to distinguish between the two.

I am willing to take a stab at implementing this if you approve of the idea. I'd rely on you and others with experience interfacing other software to select the syntax of the command so I don't inadvertently create a compatibility problem.

alexf91:
I thought this was already possible, because I've been also using Gqrx to monitor my radio's IF during a contest once.

It looks like I've implemented this more than a year ago, but forgot to make a pull request.
It's not possible to merge now because of some refactoring in the remote control code.
I'm also not sure if it is buggy or not.

Here is my implementation:
https://github.com/alexf91/gqrx/tree/remote-lnb-settings

The signal connections should be fine, but the code in remote_control.cpp needs to be updated.
If you want to continue working on this, you could at least use it as a starting point.

Let's wait what @csete has to say about this.

csete:
Sounds like it could be useful, so just go ahead and do it, regardless of what I think ;-)

All I ask is that you check the hamlib commands to see whether there already is a command that could be used instead of introducing a new one.

I need to focus on some other projects for the rest of the year and I will probably be a bit slow with my responses.

sdonnan:
Sounds good. I'll start taking a look at the source. Might take me a couple weekends before I have anything to show for it.

@alexf91 Thanks for blazing the trail. Sounds like your branch might be a good starting point.

PR #550, opened 2017-08-26
Add LNB_LO command to remote control interface

sdonnan:
Addresses #547. Adds `LNB_LO` as a command. Updated documentation.

Example:
    
    LNB_LO
    0
    LNB_LO 1000000
    RPRT 0
    LNB_LO
    1000000

Comments:

The value is handled in the app as a `double`, but the remote command exposes an integer frequency to match up with the behavior of the `f` command.

When setting the frequency `RPRT 0` or `RPRT 1` is returned as other commands do even though this is not a hamlib command. Figured I'd try to stay consistent.

Using this remote command will __not__ change the tuner configuration at runtime (at least as tested with my RTL-SDR). This is desired behavior for me but may not be for others. I'd be willing to refactor the app a bit to separate the concepts of the tuner offset and the display offset but I decided the work is outside the scope of this PR.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -27,6 +27,9 @@ Supported commands:
     Acquisition of signal (AOS) event, start audio recording
  LOS
     Loss of signal (LOS) event, stop audio recording
+ LNB_LO [frequency]
+    If frequency [Hz] is specified set the LNB LO frequency used for
+    display. Otherwise print the current LNB LO frequency [Hz].
  \dump_state
     Dump state (only usable for hamlib compatibility)
  v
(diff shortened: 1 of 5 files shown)


csete:
> Using this remote command will not change the tuner configuration at runtime 

That's fine because LNB_LO is a parameter only affecting the display and is not supposed to change anything in the tuner or anywhere else.

@alexf91 can you please take a look at is and see if it looks fine to you?

alexf91:
It works, but I would have made the command a subcommand of the set_level/get_level commands.

* "l LNB_LO" returns the frequency
* "l ?" returns "SQL STRENGTH LNB_LO"
* "L LNB_LO <freq>" sets the frequency
* "L ?" returns "SQL LNB_LO"

This keeps the interface compatible to hamlib.

csete:
Thanks for the reply.

I don't see any compatibility issues since we are extending the protocol with something that does not already exist. We do not break anything and existing applications that use hamlib will still work.

alexf91:
Ok, then let's merge ;-)

PR #565, opened 2017-10-15
Add text encoding option for RDS-TMC data.

cymatics1717:
how about adding support for  RDS text encoding?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.0)
+cmake_minimum_required(VERSION 3.0.0)
 
 # Project name
 project(gqrx)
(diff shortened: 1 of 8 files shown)


csete:
Hi,

Thanks for the patch and sorry for the late reply. I have been occupied with other projects.

Normally, I require more explanation and justification before accepting new features, in particular when they affect the GUI.

If I understand this patch correctly, it adds a combo box with a list of encodings that the user can select from. I assume text encoding is something most user will select once or never. So, I am wondering whether there is a way to automatically detect and select the correct encoding instead of having a dedicated combo box that is never used.

I noticed the use of "auto" variables. It would be nice if they could be avoided, otherwise we have to require C++11 or later in cmake.

Also, please observe  and follow the existing coding style in the source file you are modifying.

csete:
No feedback.

Issue #100, opened 2013-08-09
Add support for external data decoders


csete:
Qt has a very nice option to interact with command line applications through stdin, stdout and stderr. We can take advantage of this and create a simple interface that can be used to interact with various data decoders, e.g. multimon-ng

alexf91:
I'm writing this here, because it's the issue referenced every time when someone requests a new decoder or plugins.

@csete @gat3way
After reading yesterday's discussion in #573, I looked into [Qt-Plugins](http://doc.qt.io/qt-5/plugins-howto.html). Apparently, Qt supports loading QObjects as plugins from shared libraries at runtime.

So far, I've implemented the minimum requirements necessary to load plugins and I've moved the AFSK1200 decoder into a plugin. It's far from stable, doesn't handle a lot of errors and is missing functionality to unload plugins or show windows after they are closed.
It doesn't make writing extensions easier, but it provides a way to write decoders that are integrated into Gqrx.

The two plugins I've written are built together with Gqrx and can be loaded in the File-menu.
One is the AFSK decoder and the other one just prints to stdout.

Plugins are located in the `plugins` directory, the current interface is defined in `src/plugins`. All other changes are in `mainwindow.{h,cpp}`.

Even if it's not going to get merged (at least not yet), I thought it's worth sharing.

https://github.com/alexf91/gqrx/tree/qt-plugins

csete:
Thanks @alexf91, but I have already rejected the idea of using plugins like this in gqrx many years ago. I didn't feel that the gqrx architecture (including the many dependencies) was robust enough for this kind of extension. Unfortunately, I was right, so dynamic loading of shared object is out of the question. Not on my watch, anyway.

Note that the Qt-plugin framework is (probably) just a Qt-fied version of the more generic [dlopen & co](http://pubs.opengroup.org/onlinepubs/7908799/xsh/dlfcn.h.html) functions. Using the Qt framework would make sense for plugins that require the GUI but would be pretty hopeless to require Qt dependency for e.g. a decoder that doesn't really require anything else than stdlib, wouldn't you agree?

Anyway, I proposed a very simple method to interface with command line tools. It would require a minimal amount of design and implementation in gqrx, would be very robust and work with already existing programs like direwolf and freedv_rx. It wouldn't even matter which language the decoder is written in. People could write decoders in GNU Radio Companion and use the generated python program from within gqrx. Really simple and powerful.

alexf91:
Fair enough, but I had to try.

[...]

PR #573, opened 2017-11-25
Improved noise blankers

gat3way:
PLEASE DO NOT MERGE!!!

I opened that pull request to start a discussion about this. I ported the LMS noise reduction and Spectral Noise Reduction algorithms from the wdsp library (it's GPL so there shouldn't be licensing issues).

The way they are implemented, they replace the current NB1 and NB2. I'd say both algorithms are much more advanced and powerful than what we currently have available in gqrx and after some tweaks it got close to the quality offered by sdrsharp's noise reduction. Please see the demo video I did:

https://www.youtube.com/watch?v=ISgfsq759wU

There are lots of reasons this should not get merged in its current form. First of all, it is invasive and changed the flowchart of nbrx receiver. The blankers are now applied post-demod (even though they take complex input, thus the stupid f2c/c2f blocks, I need to resolve this as it wastes precious CPU resources).

The settings to the algorithms are currently not exposed via the UI. The new NB1 has at least two settings that can be changed - gain and threshold (defaults should be fine, but better results can be achieved with some tweaking of course), the other one does not have anything that we can tweak at runtime (well there is a FFT size parameter and by increasing the bins number we can get even better quality at the expense of lots of CPU time, I left it at some sensible default - but that's something currently implemented in the constructor, it would take some time to make it dynamically tunnable if possible at all ).

From NB2's spectral noise reduction, we only implement the WDSP advised default gain and npe methods as they are performing better than the others anyway. Unfortunately this also requires a large (about 1MB) file with precalculated constants to be available (the file is named "calculus" and is committed under src/dsp). In the current code state, this file has to be in the current directory you run gqrx from, otherwise it would crash. Thus a mechanism to install it to a common location need to be implemented. Another possibility is to hardcode the values in an array (ugly and that would increase the binary size by 1MB, but still a possibility). 

There are lots of open questions with that pull request that need to be addressed and I honestly don't expect it to be merged in any form:) But still if you are interested, I guess with some efforts this can be made part of gqrx, lot of people (especially HF listeners like me) would find it useful.

P.S the code is still ugly and needs a cleanup, I will work on this in the following days.


gat3way:
P.S actually I think this should be closed, as there are actually licensing conflicts, wdsp is GPLv2 and gqrx is Apache and I guess even if we don't use exactly the wdsp source, this would still count as a derivative work. Ah, for some reason I thought gqrx is also GPL...

alexf91:
Gqrx is GPL. The Apache License only refers to widgets from the CTK project.

gat3way:
That's great to hear.

gat3way:
OK cleaned up lot of mess. The calculus file is not needed anymore, f2c/c2f blocks are gone.

csete:
Hi,

Thanks for testing and submitting this. I think it is interesting and worth exploring.

One of my question was where this came from, but you mention that it is from WDSP. I am aware of WDSP but never really used so I want to ask: Could we use WDSP as an external library? I.e. leave the implementation there and only add GUI and wrapper code in gqrx?

If WDSP can be installed as a library by the user, gqrx can load it at runtime and enable the respective code. So it would not even be a build time dependency but loaded like a plugin. If you think this could work I can set up a dedicated branch we can play in.

I'm not convinced this noise reduction is direct replacement of the current one. One of the current ones are just for removing pulse type noise like electric fence and as I recall works quite all right. I need to take a look at the WDSP documents if they exists.

I am going to play with it a little over the coming the days to see how it works.

PS: You didn't remove the binary file ;-)

gat3way:
That would be rather hard unless you switch from the gnuradio framework for the wdsp one. Using the high level wdsp API....I don't think so. They depend on their framework. Probably we can hack our way by setting up the structs they use and by calling directly the respective helper and work functions ourselves. But those expect fixed number of samples per run and the input and output formats are not what gnuradio can directly use (they always operate with complex streams where the complex values consist of two consecutive doubles - for real streams they discard every second value). So that would be even more hacky to use together with gnuradio than grabbing and modifying the code in question.

The current noise blankers actually do exist in WDSP as well (although their algorithm is not exactly the same but a bit more complex - the overall effect and input parameters are more or less the same - see the preemptive wideband noise blanker and the interpolating wideband noise blanker in the WDSP guide at https://github.com/TAPR/OpenHPSDR-wdsp/blob/master/WDSP%20Guide.pdf

Like the gqrx noise blankers, they operate on complex IQ data and they have configurable threshold. 

Good catch about the binary file, I am removing it now...

csete:
OK, lets see where it goes.

csete:
Here are two more demos:

AM: https://www.youtube.com/watch?v=30ccfgVVEaM
SSB: https://www.youtube.com/watch?v=DPaD4ek0LBA

gat3way:
Looks nice :)

gat3way:
Shall we proceed with merging that in gqrx? If yes - there is that important problem with integrating it with the UI layer (especially the parameters of NB1 and NB2). Here I have no idea on how to proceed really. 

Also the old NB algorithms should be still present I guess, replacing them is not that great idea. 

Also I found out there is an opensource general purpose SDR software, HPSDR that is built upon WDSP. I looked at their code and it turns out they use the "wideband" blankers (the ones in WDSP that work similarily to the current NB1/NB2 in gqrx) in a rather.....strange way. They apply them almost right after the SDR source so that they operate on the whole I/Q stream and affect the spectrum/waterfall displays as well apparently. Not sure if this is a good idea though, but probably good for easier tuning of the thresholds as you get visual feedback as well. This at the expense of CPU overhead for sure.

csete:
I don't think you should expect this to be merged into gqrx master anytime soon for several reasons.

First, I need to have a proper understanding of what we are dealing with here before I can commit to any decision or to have an idea what this should be called or how it should appear in the GUI. I work on other projects now and have basically no time for gqrx, so it will take some time to process all this.

On the more practical side, credits, copyrights and references to original work should be in place before we commit the code. It might be a good idea not the replace the old NB functions but add a new noise reduction file. We can remove the old ones later.

At this point I am confused about what you mean with the various noise blankers. Where they are applied depends on their type and purpose. They can be wide band applied at the input, IF or audio. As I recall this new one is audio, which is partly why I don't think it can not replace our existing ones. There are many noise types that need tp be removed before they can affect the AGC, which is in the IF.

So, as you can see, we (or at least I) need time to understand what we are dealing with.

gat3way:
Sorry for pushing and I understand your concerns. I have a suggesion but probably you've seen such an argument before - how about having an experimental branch? This way people would be encouraged to contribute while you can have your time on your other projects and only merge in the master when you decide stuff is mature enough and compliant?

The copyrights and credits part in that particular case are easy to handle as long as both projects stick to GPLv2. However it is exactly the project's ideological (plus the UI) realms which I am really lost into. So again, perhaps having an experimental branch is not that bad idea. I am sure lots of user contributions will eventually get there and this will only benefit the project.

csete:
We get code into gqrx when it is ready. I sometimes create dedicated development branches, and this PR may indeed start in such a branch. However, thiecode will not get more exposure by being in branch versus being in a pull request here. In fact, quite the opposite.

My comment about license was not conceptual but referring to a specific problem. You replaced the code in the original noise blanker sources with new code, but left the copyright header untouched. Things like this have to be resolved before we put the code in gqrx repo.

gat3way:
Fair enough. Meanwhile for the fun of it I've started an experimental branch to bring some features into gqrx. Many of them are quick hacks that need a lot of cleanup and polishing. Others contradict with the project philosophy (I know you guys aren't very keen on having decoders in gqrx for example). I don't expect most if not any of this to ever get merged into master and I am not going to submit pull requests for any of the changes made. Still I think it's worth and interesting working on something like this as a side project.

csete:
> I know you guys aren't very keen on having decoders in gqrx for example

It depends on what "in gqrx" means... If it means coded inside gqrx like the AFSK decoder, then no. If it means having an interface to interact with external decoders, then yes, see issue #100.

gat3way:
Yeah I've read that before. I kindly disagree though because I don't think a common interface can be designed so that users can have seamless operation in all cases. Sometimes it's pulseaudio that "glues" it all because the external decoder doesn't care about stdin/stdout (such as say fldigi). Sometimes yes piping stdin/stdout could be an option (say multimon-ng). Sometimes both aren't an option at all - like in e.g rtl_433's case. It just wants to operate on I/Q data and it does the demodulation itself, it doesn't like the idea of getting input from stdin which makes it impossible to easily tune to some signal while watching it on the waterfall (leave alone it supports only rtl-sdr hardware). Or another example (as in Issue #100) - DSD. It is hard to define such interface. It might output data from the MBE codec to the audio device and then the decoded frames on stdout. Pulseaudio could do the resampling and you would also get the stdout. But I am afraid there is just no standartized way to do that and people would have to fall back to manual configuration. And yes, I agree MBE is quite a bad problem with those patent issues.

csete:
I never claimed this interface would solve everybody's problem once and for all. It would merely provide a clean and simple interface to use external decoders. Some decoders would work as is, some would require new interface.

Anyway, this is off topic here.

spamcop:
I just want to thank gat3way for porting this, its a game changer in listening comfort.

csete:
Doesn't look like this will happen.

ulrichloose:
I've been using this great Noiseblanker for quite a while and I'm thrilled ... But now I've noticed that gqrx crashes if you accidentally set RAW I / Q. I don't need RAW I / Q, so I just set the entry to the same setting as Demod off. This is of course not correct ...

Maybe ra1nb0w could take another look at his code and fix it properly?

ra1nb0w:
@ulrichloose you should paste the trace error from the console ...

PR #581, opened 2017-12-31
SDRplay sample rate

willcode:
The SDRplay RSP1, RSP1A and RSP2 all have sample rates from 2M to 10M. This change clips user-selected sample rates to that range, and adds a set of preselected sample rates within that range.

### DIFF
--- a/src/qtgui/ioconfig.cpp
+++ b/src/qtgui/ioconfig.cpp
@@ -531,31 +531,23 @@ void CIoConfig::updateInputSampleRates(int rate)
     }
     else if (ui->inDevEdit->text().contains("sdrplay"))
     {
-        ui->inSrCombo->addItem("222222");
-        ui->inSrCombo->addItem("333333");
-        ui->inSrCombo->addItem("428571");
-        ui->inSrCombo->addItem("500000");
-        ui->inSrCombo->addItem("571429");
-        ui->inSrCombo->addItem("750000");
-        ui->inSrCombo->addItem("875000");
-        ui->inSrCombo->addItem("1000000");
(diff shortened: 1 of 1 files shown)


src/qtgui/ioconfig.cpp
```diff
@@ -531,31 +531,27 @@ void CIoConfig::updateInputSampleRates(int rate)
...
         ui->inSrCombo->addItem("6000000");
         ui->inSrCombo->addItem("7000000");
         ui->inSrCombo->addItem("8000000");
         ui->inSrCombo->addItem("9000000");
         ui->inSrCombo->addItem("10000000");
-        ui->inSrCombo->addItem("11000000");
-        ui->inSrCombo->addItem("12000000");
         if (rate > 0)
         {
+            if (rate < 2000000)
```
csete: We do not have this check for other devices to allow for some experimentation, so I suggest we leave it out.

csete:
Thanks fir the update. I'm not sure what happened... the invalid rates must also have come from an official source.

src/qtgui/ioconfig.cpp
willcode: I've also got the limit check in the gr-osmosdr sdrplay module I'm rewriting, to avoid errors. If a bad value is entered in gqrx, gr-osmosdr/get_sample_rate() returns a good value and gqrx uses the good value. However, the entry in the configuration dialog still shows the bad value, which may confuse the user. Up to you.

csete: Ok, then lets leave it in for now.

PR #591, opened 2018-02-01
Microtelecom Perseus support added

amontefusco:
Alexandru,
this patch is about integrating into gqrx the support for Microtelecom HF Perseus receiver.
I already added a specific class in gr-osmosdr found here:

https://github.com/IW0HDV/gr-osmosdr/tree/perseus


csete:
Thanks. Lets hope the gr-osmosdr changes can be merged upstream.

PR #605, opened 2018-03-03
Update gqrx.pro

szg0000:
Error message if pkg-config is missing.


csete:
Thanks. I have now merged the other patch from PR #604 so it should be fine.

PR #610, opened 2018-03-17
added MANIFEST.md for CGRAN parsing sake

777arc:
We are currently working on revitalizing CGRAN, the current version is here https://gnuradio.org/cgran/  and just like the old CGRAN it parses OOT's manifest files to get its information.  Feel free to modify stuff, I just copied whatever info I could find.


csete:
Ok, thanks for the patch.

PR #613, opened 2018-03-29
Add backspace key

Arinerron:
Backspace key has the same function as the left arrow key. When typing in a frequency using the number keys, I keep accidentally hitting the backspace key expecting it will move the digit it is editing one backwards (the same thing the left arrow does).


csete:
We can add this, but I suggest you make it a fallthrough case. There is no reason to repeat the same code.

Arinerron:
@csete Good point, fixed in latest commit.

csete:
Merged manually. I prefer to have such commits squashed into one.

PR #614, opened 2018-04-01
Make the frequency plot's tooltip's unit adapting to the displayed frequency

martonmiklos:
First of all many thanks for this great software!

kep
The unit of the frequency plot was hardcoded which made sometimes difficult to interpret the displayed frequency. This fix automatically generates the proper unit for the displayed frequency:
* 0-1kHz suffix with Hz
* 1kHz-1MHz suffix with kHz
* 1MHz-1GHz suffix with MHz
* above 1GHz suffix with GHz

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -256,8 +256,8 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
                 m_CursorCaptured = CENTER;
                 if (m_TooltipsEnabled)
                     QToolTip::showText(event->globalPos(),
-                                       QString("Demod: %1 kHz")
-                                       .arg(m_DemodCenterFreq/1.e3f, 0, 'f', 3),
+                                       QString("Demod: %1")
+                                       .arg(freqToString(m_DemodCenterFreq)),
                                        this);
             }
             else if (isPointCloseTo(pt.x(), m_DemodHiCutFreqX, m_CursorCaptureDelta))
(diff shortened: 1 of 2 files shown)


csete:
The purpose with the tooltip is to have a tool that we can use to read out the frequency with high 
resolution. In most cases users need sub-kHz resolution regardless of the RF frequency. This no longer works if you limit the number of decimal digits to 3.

If you want to make it adaptive, you also need to take the FFT span into account. However, I'd like to point out that I have deliberately left the display fixed because I didn't want to add more even more processing to an already CPU intensive graphics code. Any addition to this code will have to be accompanied by evidence that it does not severely impact performance on ARM boards.

martonmiklos:
> The purpose with the tooltip is to have a tool that we can use to read out the frequency with high
> resolution. In most cases users need sub-kHz resolution regardless of the RF frequency. This no longer works if you limit the number of decimal digits to 3.
I see your point, I could remove the decimal digit limitation or limit in the same way as it was before this PR (3 digits in kHz)

> If you want to make it adaptive, you also need to take the FFT span into account. 
Could you please spec it exactly how would you like to see. 

> However, I'd like to point out that I have deliberately left the display fixed because I didn't want to add more even more processing to an already CPU intensive graphics code. Any addition to this code will have to be accompanied by evidence that it does not severely impact performance on ARM boards.
All freqToString calls are in the mouseMoveEvent which means the recalculation only performed when the users moves the mouse. I only have an Odroid XU4 board to test.

csete:
Sorry, but I work on other SDR projects now and I can only spend minimal time on gqrx. I don't wish to spend that time on writing specs for something I did not ask for. If it is not clear how the FFT plot span affects the requirements on the tooltip resolution, I think we should just leave it as is.

martonmiklos:
Well I think your 3 sentence whining have required the same amount of time what would needed to describe what and how would you like to see to prevent more confusion. 

If you are so smart to be comfortable with interpreting frequencies in kHz even in the GHz range, that is fine, fortunately the SDRSharp authors thinks it differently.

csete:
I already told you what I want, but apparently that wasn't enough. My last message was just a polite way of saying that I am not interested in your quick hack that you haven't even tested properly.

Since you mention GHz range, try tuning to above 1 GHz then zoom in so that the FFT span is less than 1 MHz. As you will see, the tooltip will be stuck showing the same frequency with 1 MHz resolution. See what I mean?

PR #621, opened 2018-04-13
Add band plan and info to plotter

walthercarsten:
Hi, i added some band informations to the plotter output. It is based on a csv file [resources/bandplan.csv](resources/bandplan.csv) that provides the informations  of min frequency, max frequency, mode, step, color and name of the band. These informations are shown at the bottom of the frequency above the waterfall. I think it would "a nice to have" in a new version.

Best regards, Carsten


alexf91:
I would make it optional.

alexf91:
This is not working as intended. When the zoom is low (<0.5 in my case), the regions that are displayed are not correct.
This is the 10m band as it should be:
screenshot_10m_normal

And this is with a low zoom:
screenshot_10m_error

walthercarsten:
Where is the right place to add a checkbox for optional band informations?

csete:
Thanks for the submission. At this time I don't know when there will be a new feature release of gqrx, but I will be happy to consider it when we get to that point. So there is good time to work out the details.

As for how to show/hide, I think the view menu is the right place for this.

argilo:
I'm closing this off in favour of #767.

PR #623, opened 2018-04-18
Bookmarks handling

walthercarsten:
Hi, i added a more usefull bookmark generation in csv file and set the bookmarks table column width to  stretch mode.


csete:
Looks like your patch contains much more than what you intended to submit. For instance, it includes your band plan changes. Please submit patches that address only one issue at a time. I would also recommend to keep functional changes separate from stylistic changes, as the latter are more likely to be rejected.

Issue #524, opened 2017-06-27
Feature Request: Remote Control LNA Gain


hosler:
I would like to be able to control my dongle's gain via remote control

csete:
We can take a look, but I 'm not sure how to handle this. Different devices have different type of gain settings.

danhosler:
only toggling the gain on/off might be easier if you need to accommodate all devices.

alexf91:
I've started working on this.
I'll expose it through the "level" command (l, L) and call the different gain settings <GAIN_NAME>_GAIN.

csete:
What I was trying to say is that different devices have different gain names, so the TCP client would need to know which gains are available. Then suddenly we have data flow from gr-osmosdr to a TCP client - sounds like a bad idea...

alexf91:
I haven't really thought about the client side, but the available gain names can be queried by using the "l ?" command and filtering the return values for "_GAIN".

csete:
PR #635 merged and will be available in 2.12.

PR #635, opened 2018-05-28
Read and write gain using remote control

alexf91:
Read and write the gain settings via remote control, as discussed in #524.
The settings are exposed as <NAME>_GAIN and are accessed with the "l" (read) and "L" (write) commands.


csete:
Thanks for the patch; however, as you probably already know, I do not anticipate any new feature releases at this time. Should that change in the future, I will be happy to take a closer look at this.

csete:
We should probably update the documentation of the remote control protocol to include this.

csete:
Thanks for the update. Merging now...

PR #637, opened 2018-05-30
Replace deprecated qt5_use_modules macro

alexf91:
The `qt5_use_modules` macro was removed in Qt 5.11, causing the build to fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.0)
+cmake_minimum_required(VERSION 2.8.11)
 
 # Project name
 project(gqrx)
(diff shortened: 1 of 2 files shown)


csete:
Thanks for the heads up.

Is this the official way to do this? Seems a bit odd to replace a macro call with library listings...

alexf91:
The Qt documentation states it this way.
> The recommended way to use Qt libraries and headers with CMake 2.8.11 is to use the target_link_libraries command.
https://doc.qt.io/archives/qt-5.10/cmake-manual.html

 Using the macro is recommended for CMake older than 2.8.11, but for newer versions, `target_link_libraries` is recommended.

csete:
Ok, thanks. As far as I can see, there is no problem with requiring cmake 2.8.11

PR #656, opened 2018-09-23
Add buttons to reset, increase and decrease audio gain

Kezii:
Adds a button to set the audio gain to 0db, one to increase it by 10db and one to decrease it by 10db, under the slider
image

### DIFF
--- a/src/qtgui/dockaudio.cpp
+++ b/src/qtgui/dockaudio.cpp
@@ -115,7 +115,6 @@ int  DockAudio::audioGain()
     return ui->audioGainSlider->value();
 }
 
-
 /*! Set FFT plot color. */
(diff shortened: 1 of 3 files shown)


csete:
Hi,

Sorry for the late feedback.
In general, I'm trying to keep the number of redundant controls down to a minimum, so perhaps you can clarify why these controls are necessary?

Kezii:
Hi, thanks for answering

I made this to fix the minor annoyance that you can't quickly reset the gain slider, and end up with 0.1db of gain or something like that

It was only an idea

argilo:
Thanks for the proposal @Kezii. I tend to agree with @csete, though, that these additional controls are not useful enough to justify the extra space they would take up. Thus I'm going to close off this pull request.

PR #662, opened 2018-10-20
Patch to make CMake correctly resolve qt5 installed via homebrew

nickrobinson:
Mac users who have installed Qt5 using homebrew may get the following error when running cmake:

```
CMake Error at /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:15 (message):
  The imported target "Qt5::Core" references the file

     "/usr/local/.//mkspecs/macx-clang"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained
```

This fix will append to the cmake prefix path the directory where homebrew installs qt5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,12 @@ function(add_source_files list)
     set_property(GLOBAL APPEND PROPERTY ${list} "${SRCS}")
 endfunction(add_source_files)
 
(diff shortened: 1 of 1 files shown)


csete:
Thanks for the fix.

PR #663, opened 2018-11-03
Add start/stop DSP from remote control

Brandonn-Etheve:


kukabu:
The "d" and "D" commands are already in use, see man rigctl. I think it can be realized via "w" command.

alexf91:
> I think it can be realized via "w" command.

Or, to keep it consistent with the RECORD command, with "u" and "U". This way, the status of the DSP can be read and written.

kukabu:
Oh, I forgot it. I think this is better way. Thanks.

src/applications/gqrx/remote_control.cpp
```diff
@@ -643,9 +620,11 @@ QString RemoteControl::cmd_get_func(QStringList cmdlist)
     QString func = cmdlist.value(1, "");
 
     if (func == "?")
-        answer = QString("RECORD\n");
+        answer = QString("RECORD\nDSP\n");
```
alexf91: The newline between RECORD and DSP should be replaced with a space.

src/applications/gqrx/remote_control.cpp
```diff
@@ -662,7 +641,7 @@ QString RemoteControl::cmd_set_func(QStringList cmdlist)
 
     if (func == "?")
     {
-        answer = QString("RECORD\n");
+        answer = QString("RECORD\nDSP\n");
```
alexf91: Same here

Brandonn-Etheve:
I was wondering if i should stop the audio recording when one stop the dsp by remote control.

However i just noticed that turning off the dsp from gui did not turn off the audio recording (neither from the gui or by remote control with u RECORD).

In this situation you cant turn off audio recording with U RECORD 0 because this command first check that the dsp is running, witch is not the case... is this a bug?

scandioguarnaschelli:
i hope this is the right place for this message:
while the reception frequency can be changed with the mouse, the transmission frequency seems to be fixed once the transmission has been activated; for use with the satellites it is also necessary to modify the transmission tuning; is it possible to insert this function in the menu?

alexf91:
> i hope this is the right place for this message:
It is not.

Open an issue if you think there is a bug in Gqrx or if you require a certain feature.
Contact the mailing list if you need support: https://groups.google.com/forum/#!forum/gqrx

argilo:
Sorry for the long delay in following up on this pull request. I rebased it to resolve merge conflicts, and made a couple tweaks. I'll merge once CI passes.

PR #671, opened 2019-01-09
extend sample rates for bladerf micro, set default

ZeroChaos-:
libbladerf simply presents "bladerf" for both original bladerf and the
new bladerf micro.  this causes some pain for new users by default.  I
have extended the available sample rates to include the new sample rates
for the bladerf micro, but more importantly I have set the default
sample rate to the lowest value supported by both original and micro
hardware to avoid crashing with default settings

### DIFF
--- a/src/qtgui/ioconfig.cpp
+++ b/src/qtgui/ioconfig.cpp
@@ -384,11 +384,8 @@ void CIoConfig::updateInputSampleRates(int rate)
     }
     else if (ui->inDevEdit->text().contains("bladerf"))
     {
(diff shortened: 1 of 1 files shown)


csete:
I just want to avoid that the application crashes because the user selects an unsupported sample rate. That's what's happening with the Airspy and Airspy Mini. I haven't tried with Bladerf.

ZeroChaos-:
As requested, a simpler PR that disables the sample rates not support by both models.

csete:
Thanks

PR #677, opened 2019-02-05
MACOS INSTALLATION

Rotron:


csete:
Thanks, but the build instructions in the readme are kept generic on purpose. There are many ways to build gqrx on any given platform and I just don't have the capacity to document and maintain them all.

PR #678, opened 2019-02-20
Add shortcut. Pressing F moves focus to frequency widget.

andreiva:
For a long time I wanted to have a shortkey for setting frequency, so I didn't have to use mouse. It probably could be done better, but this solution works and it makes my life easier. I hope you can merge it like it is or make it better. 

-Andrei

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -2300,3 +2300,13 @@ void MainWindow::on_actionAddBookmark_triggered()
         ui->plotter->updateOverlay();
     }
 }
+
+void MainWindow::keyPressEvent(QKeyEvent *event)
+{
+    switch (event->key())
+    {
+    case Qt::Key_F:
+        ui->freqCtrl->setFrequencyFocus();
+        break;
+    }
+}
--- a/src/applications/gqrx/mainwindow.h
+++ b/src/applications/gqrx/mainwindow.h
@@ -32,6 +32,7 @@
 #include <QMessageBox>
 #include <QFileDialog>
 #include <QSvgWidget>
+#include <QKeyEvent>
 
 #include "qtgui/dockrxopt.h"
 #include "qtgui/dockaudio.h"
@@ -120,6 +121,9 @@ private:
 
     // dummy widget to enforce linking to QtSvg
     QSvgWidget      *qsvg_dummy;
+
+protected:
+    void    keyPressEvent(QKeyEvent *);
 
 private:
     void updateHWFrequencyRange(bool ignore_limits);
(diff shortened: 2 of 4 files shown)


csete:
Thanks, I will consider it, though we need to understand what side effect this might have. I'm not convinced that having a single letter shortcut without any modifier is a good idea.

andreiva:
Yeah, no problem. If I remember correctly I used F without modifier because CTRL+F was already taken.
If plain F is conflicting, it would be nice to get some easy combination. CTRL+SPACE perhaps.

ra1nb0w:
@andreiva is it still working? because if I try to use your patch and press ```F``` nothing happen.
thank you

andreiva:
> @andreiva is it still working? because if I try to use your patch and press `F` nothing happen.
> thank you

I don't think @csete has merged it yet. It worked on my machine, and it was handy..

ra1nb0w:
I am trying to use your patch (PR) with the latest commit in master but seems that nothing happen after pressing `F`

andreiva:
> I am trying to use your patch (PR) with the latest commit in master but seems that nothing happen after pressing `F`

Ah sorry, now I understand. Haven't compiled the project in a while, I'll try to do it soon.

ra1nb0w:
thank you :)

andreiva:
> thank you :)

Seems to work okay.
Try this:
```
git clone https://github.com/andreiva/gqrx gqrx-andreiva
cd gqrx-andreiva
git remote add csete https://github.com/csete/gqrx
git fetch csete
git rebase csete/master
```
Then build and run

ra1nb0w:
thank you for testing. now I understood why in macOS 10.15 it doesn't work. You need to enable Accesibility cabability. Probably not a good thing; maybe there is another way to catch the key but I am not a Qt expert.

andreiva:
> thank you for testing. now I understood why in macOS 10.15 it doesn't work. You need to enable Accesibility cabability. Probably not a good thing; maybe there is another way to catch the key but I am not a Qt expert.

I'm not a Mac user (Linux Mint) but it seems weird you have to enable Accesibility features to use a shortcut. Other shortcuts work, don't they? Is it because it's "f" without a modifier like CTRL/OPTION or is it because "f" is default system shortcut key in Mac just wondering.
Could you modify the code and try some other key?

ra1nb0w:
`QCursor::setPos` can't be used on macOS without enabling extra capability (_Accessibility_). See [this](https://stackoverflow.com/questions/60064492/find-out-if-qcursorsetpos-has-an-effect).

this [patch](https://gist.github.com/ra1nb0w/076b5a750a1a5615cbcd7c7bf1570ef8) should be enought to avoid the problem and preserve the functionality without duplicating code. Anyway, the controller should also implement left/right key to move on the frequency. Moreover, clicking _F_ more times freeze the input.

argilo:
@ra1nb0w has made an updated version of this, so I'll close this off and open a separate pull request for that version.

Issue #675, opened 2019-01-31
gqrx shouldn't exit with SIGABRT on user errors


l29ah:
Say,
```
[blah]
Using device #0 Generic RTL2832U OEM
usb_open error -3
Please fix the device permissions, e.g. by installing the udev rules file rtl-sdr.rules
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to open rtlsdr device.
zsh: abort (core dumped)  gqrx
```
shouldn't happen as it is not a programming error, yet it produces an useless core dump.

csete:
This is a weakness of the architecture, but please do send a pull request if you think it can be fixed in a clean way.

argilo:
This issue can be closed, since it was addressed in 34333e78b2e565745b6d62bf0efcc57cc3b1988f.

l29ah:
Thanks.

PR #686, opened 2019-05-04
Mitigate #675: trap exceptions and show a message instead of crashing

studiofuga:
Signed-off-by: Federico Fuga <fuga@studiofuga.com>


csete:
Okay, let's try it!

PR #687, opened 2019-05-13
Add second resampler for raw IQ demod

### LINKED ISSUE #684: Raw I/Q passthrough: I and Q data are identical
I thought the "Raw I/Q" option for the demodulator would be a true passthrough of the I/Q complex data, instead I notice that the two channels of the wav file are identical, while I thought I would have found the I data in one channel and the Q data in the other.

alexf91:
Fixes #684 

The second resampler is only used when IQ demod is selected. For all other modes only `audio_rr0` is used like before.


csete:
Thanks for the fix @alexf91!

Issue #689, opened 2019-06-06
Feature Request: Mute Button


LightningStalker:
One thing I find myself frequently doing is turning the volume down to -80dB in order to mute the af to hear other programs. It would be nice if there was an actual mute button on the interface to do this while keeping the gain setting.

csete:
Available since the merge of #690 by @alexf91

PR #690, opened 2019-06-06
Add mute button

alexf91:
Adds the mute button requested in #689 

Enabling the mute button sets the audio gain to -infinity, which results in a multiplier value 0 in the receiver.

mute_button


nickoe:
Wouldn't it be better if this mute button was just a shortcut to toggle the demodulator between off and latest setting?

alexf91:
@nickoe It's not the same thing. The mute button only mutes the audio output, but keeps the UDP stream running.

nickoe:
@alexf91 Ok, I didn't think of that use case.

Issue #657, opened 2018-09-24
gqrx does not compile against latest gnuradio 3.8 tech preview


lukaskuzmiak:
and I fixed that here: https://github.com/lukaskuzmiak/gqrx/commit/9752cc315da23f2fc99411a7edc360b9145cbf77

It seems you keep master working against gnuradio 3.7 and this will break it, do I create a pull request for master or some other branch (did not find one that would work for that).

Thanks!

lukaskuzmiak:
btw, this has one more issue, due to changes here: https://github.com/gnuradio/gnuradio/commit/6ae02d1d6d04a5f529ab3fea496c589f8d14229d#diff-9e599a674a0b50c90760539b4f037cef

GQRX fails to start with:

```
gr::log :WARN: file_source0 - file is too small
terminate called after throwing an instance of 'std::runtime_error'
  what():  file is too small
Aborted (core dumped)
```

This is because the way they check the file size does not work for `/dev/random`, I verified passing a file w/ size that is dividable by 8 (`sizeof(gr_complex)`) fixes this issue but not really sure what would be a good fix, suggestions welcome.

csete:
This will be a very annoying maintenance problem for many years :(

I could create a separate branch; however, that would require keeping it up to date all the time. So, I suggest that we check the GNU Radio version in cmake, and use #ifdefs in the code to make it build with both version 3.7 and 3.8.

What do you think?

lukaskuzmiak:
Makes sense to me, I'm honestly not sure how many breaking changes is 3.8 going to bring in the APIs and how "spaghetti code" will we need to produce to work around that. But so far it does not look too bad, so I agree #ifdefs sound better than a branch somebody needs to take care of.

englishm:
GNU Radio 3.8.0.0 was released today: 
https://github.com/gnuradio/gnuradio/releases/tag/v3.8.0.0

There is a migration guide for out of tree (OOT) modules here: https://wiki.gnuradio.org/index.php/GNU_Radio_3.8_OOT_Module_Porting_Guide

There's an open bug to update gr-osmosdr here:
https://osmocom.org/issues/3855

And the patches here appear to work (at least with the rc's, I haven't tested again yet):
https://github.com/igorauad/gr-osmosdr/tree/gr3.8

[...]

PR #696, opened 2019-08-17
Update to GNURadio versions >=3.8 (Changed file structure)

GHPS:
With GNURadio v.3.8 the structure of some C files has changed (see GR commit "blocks: replace gengen with C++ templates", 1043222e0de551f00683a1cc9060f194adc4fc85 for details).

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -28,7 +28,7 @@
 
 #include <iostream>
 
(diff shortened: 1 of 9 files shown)


csete:
Thanks for the patch; however, we need to get the builds under control and also have gr-osmosdr ported before this can be merged into the master branch. There is a discussion in issue #657.

csete:
Gqrx should now build with both gnuradio 3.7 and 3.8.

PR #703, opened 2019-09-24
Port to Gnuradio 3.8

StefanBruens:
- Use new syntax for required GR components
- Use gnuradio::gnuradio-* instead of GNURADIO_ALL_LIBRARIES
- Update sources for renamed/consolidated headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,14 +99,8 @@ endfunction(add_source_files)
 
 # 3rd Party Dependency Stuff
 find_package(Qt5 COMPONENTS Core Network Widgets Svg REQUIRED)
-find_package(Boost COMPONENTS system program_options REQUIRED)
-set(GR_REQUIRED_COMPONENTS RUNTIME ANALOG AUDIO BLOCKS DIGITAL FILTER FFT PMT)
-find_package(Gnuradio REQUIRED)
+find_package(Gnuradio "3.8" REQUIRED COMPONENTS analog audio blocks digital filter fft)
 find_package(Gnuradio-osmosdr REQUIRED)
-
-if(NOT GNURADIO_RUNTIME_FOUND)
-    message(FATAL_ERROR "GnuRadio Runtime required to compile gqrx")
-endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     if(NOT LINUX_AUDIO_BACKEND)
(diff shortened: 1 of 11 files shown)


alexf91:
Porting is not so simple. GNU Radio 3.7 will likely be around for years, because many applications are poorly maintained and won't be port to 3.8. It appears to be relatively easy to support both version in Gqrx.

Here are more discussions:
Existing PR: https://github.com/csete/gqrx/pull/696
Discussion on porting and support for 3.7/3.8: https://github.com/csete/gqrx/issues/657

BTW, what is my commit doing in your PR?

StefanBruens:
When you have a ported gr-osmosdr, it actually *is* so simple.

I have an almost complete set of OOT GR modules updated to GR 3.8, porting is not really hard after the first package. The packages I have omitted are mostly ones which also depend on e.g. Qt4, which is EOL for years.

Unfortunately, Github seems to be unable to track common commits, my branch was branched from the v2.11.5 tag.

alexf91:
> When you have a ported gr-osmosdr, it actually is so simple.

Sure, but we should wait for distributions before switching Gqrx to 3.8-only. It makes little sense to switch to 3.8 if popular distributions like Ubuntu can't run new versions. If we only have to include different headers and have some different names, then we can support both versions.

alexf91:
@StefanBruens I have rebased your branch: https://github.com/alexf91/gqrx/tree/gnuradio3.8

csete:
Thanks for the patch; however, as we also discussed in issue #657, it would be nice to support both versions as long as the changes are relatively simple.

CMakeLists.txt
```diff
@@ -99,15 +99,9 @@ endfunction(add_source_files)
 
 # 3rd Party Dependency Stuff
 find_package(Qt5 COMPONENTS Core Network Widgets Svg REQUIRED)
-find_package(Boost COMPONENTS system program_options REQUIRED)
```
csete: These Boost components are required by Gqrx

src/CMakeLists.txt
```diff
@@ -58,12 +58,17 @@ endif(WIN32)
 #######################################################################################################################
 # Build the program
 add_executable(${PROJECT_NAME} ${${PROJECT_NAME}_SOURCE} ${UIS_HDRS} ${RESOURCES_LIST})
-qt5_use_modules(${PROJECT_NAME} Core Network Widgets Svg)
```
csete: Why change this? Even is the other way is preferred, is the change related to the porting?

csete:
I have now merged the patch from @alexf91 so gqrx should now build with both gnuradio 3.7 and 3.8.

PR #705, opened 2019-10-06
Build for GNU Radio 3.7 and 3.8

alexf91:
Modifications to source files and CMake files to support builds with GNU Radio 3.7 and 3.8 as discussed in #657 and some pull requests.

There are a few clarifications and possible changes necessary. Some information can be found in the [porting guide](https://wiki.gnuradio.org/index.php/GNU_Radio_3.8_OOT_Module_Porting_Guide).

- [x] Confirm that Boost is discovered and linked correctly.
- [x] Rename the preprocessor macro (e.g. to version number as hex like QT_VERSION)
- [x] Test on more platforms
- [x] Check if the required CMake version has to be changed
- [x] Support qmake

So far I have tested it on Debian Buster for GNU Radio 3.7 and Debian Sid and Arch for GNU Radio 3.8.
Here are the Docker images and scripts I used for testing: https://github.com/alexf91/gqrx-gnuradio-build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,14 +105,18 @@ endif()
 
 # 3rd Party Dependency Stuff
 find_package(Qt5 COMPONENTS Core Network Widgets Svg REQUIRED)
-find_package(Boost COMPONENTS system program_options REQUIRED)
-set(GR_REQUIRED_COMPONENTS RUNTIME ANALOG AUDIO BLOCKS DIGITAL FILTER FFT PMT)
-find_package(Gnuradio REQUIRED)
 find_package(Gnuradio-osmosdr REQUIRED)
 
+set(GR_REQUIRED_COMPONENTS RUNTIME ANALOG AUDIO BLOCKS DIGITAL FILTER FFT PMT)
+find_package(Gnuradio REQUIRED COMPONENTS analog audio blocks digital filter fft)
 
-if(NOT GNURADIO_RUNTIME_FOUND)
+if(NOT Gnuradio_FOUND)
     message(FATAL_ERROR "GnuRadio Runtime required to compile gqrx")
+endif()
+
+if(Gnuradio_VERSION VERSION_LESS "3.8")
+    add_definitions(-DGNURADIO_37)
+    find_package(Boost COMPONENTS system program_options REQUIRED)
 endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,8 +114,20 @@ if(NOT Gnuradio_FOUND)
     message(FATAL_ERROR "GnuRadio Runtime required to compile gqrx")
 endif()
 
+
+# Pass the GNU Radio version as 0xMMNNPP BCD.
+math(EXPR GNURADIO_BCD_VERSION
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -105,16 +105,20 @@ endif()
 
 # 3rd Party Dependency Stuff
 find_package(Qt5 COMPONENTS Core Network Widgets Svg REQUIRED)
-find_package(Boost COMPONENTS system program_options REQUIRED)
-set(GR_REQUIRED_COMPONENTS RUNTIME ANALOG AUDIO BLOCKS DIGITAL FILTER FFT PMT)
-find_package(Gnuradio REQUIRED)
 find_package(Gnuradio-osmosdr REQUIRED)
 
+set(GR_REQUIRED_COMPONENTS RUNTIME ANALOG AUDIO BLOCKS DIGITAL FILTER FFT PMT)
+find_package(Gnuradio REQUIRED COMPONENTS analog audio blocks digital filter fft)
```
alexf91: This combines the instructions from 3.7 and 3.8.
The first line is only used by the 3.7 CMake files and the CMake components are only used for 3.8. After this step, the Gnuradio_FOUND variable should be set.

CMakeLists.txt
```diff
@@ -105,16 +105,20 @@ endif()
...
 
-if(NOT GNURADIO_RUNTIME_FOUND)
+if(NOT Gnuradio_FOUND)
     message(FATAL_ERROR "GnuRadio Runtime required to compile gqrx")
 endif()
 
+if(Gnuradio_VERSION VERSION_LESS "3.8")
+    add_definitions(-DGNURADIO_37)
+    find_package(Boost COMPONENTS system program_options REQUIRED)
```
alexf91: Boost is automatically discovered for 3.8. Discovering it again seems to cause problems in the child scope.
This needs some clarifications.

src/CMakeLists.txt
```diff
@@ -73,6 +73,15 @@ target_link_libraries(${PROJECT_NAME}
...
 )
 
+if(NOT Gnuradio_VERSION VERSION_LESS "3.8")
+    target_link_libraries(${PROJECT_NAME}
+        gnuradio::gnuradio-analog
+        gnuradio::gnuradio-blocks
+        gnuradio::gnuradio-digital
+        gnuradio::gnuradio-filter
+    )
+endif()
```
alexf91: For 3.7 the linked libraries are defined in GNURADIO_ALL_LIBRARIES, which should be empty for 3.8.

alexf91:
@csete Any thoughts how or if this should go forward?
I think the CMake build is fine, but I haven't looked into integrating this into the qmake build.

csete:
@alexf91, thanks for working on this. Looks good!

Regarding qmake, we can leave as a 3.7 option only. Personally, I only use the gqrx.pro file to load a project into Qt Creator, so as long as the list of files stays up to date, it will be suffucient for me.

alexf91:
Let's merge this and hope for the best.
I have tested it with the GNU Radio and Pulseaudio backends and with cmake and qmake.

I have no way to test it on OS X, but I don't see a reason why it wouldn't work.

csete:
Okay, so I am going to merge it. Sorry for the delay and thanks for the patch!

Issue #504, opened 2017-03-27
hackrf pipe error


derekchal:
Recently I have not been able to start gqrx on my desktop machine.
I get a pipe error shown below:

_linux; GNU C++ version 6.2.0 20161005; Boost_106100; UHD_3.11.0.git-28-gc66cb1ba

gr-osmosdr v0.1.4-86-ge9dde9af (0.1.5git) gnuradio 3.7.12git-48-gc73a73fd
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy soapy redpitaya 
Resampling audio 96000 -> 48000
BookmarksFile is /home/derek/.config/gqrx/bookmarks.csv
gr-osmosdr v0.1.4-86-ge9dde9af (0.1.5git) gnuradio 3.7.12git-48-gc73a73fd
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy soapy redpitaya 
Using HackRF One with firmware 2017.02.1 
terminate called after throwing an instance of 'std::runtime_error'
  what():  hackrf_set_freq(1e+10) has failed (-1000) Pipe error
Aborted_

It works correctly with a rtl dongle and a funcube dongle. 

Has anybody seen this error. I could well believe that I have updated something which causes a conflict.

csete:
If you update your firmware in the device you have to make sure that libhackrf is also up to date and that gr-osmosdr / gqrx uses this latest library.

Please use the mailing list for user support matters and keep github issues for bugs and feature tracking
http://gqrx.dk/user-support

phocean:
You should reopen this issue. I am having the same problem, whereas all my libs and my firmware are in sync.
Moreover, it should not fail silently with a core dump.
It works if I open it on another device input, then switch back to the hackrf... until the next time I relaunch the tool.

csete:
@phocean If somebody provides a proper error report using officially released packages and using gnuradio 3.7.10.x, I will consider. Note that the original poster used an arbitrary mixture of git snapshots and I don't even have a clue which operating systems are being used.

Even if there is an error and we reopen this issue, most likely nothing will happen. Gqrx is just a GUI and there is no device specific code in it. So the relevant developers who might be able to fix it wont even know. In the past I have spent a lot of time fixing and coordinating these issues but I don't have time for it anymore.

phocean:
No problem, I understand perfectly. I will try to post more details.

[...]

PR #718, opened 2020-01-14
Reorder readSettings to prevent use of uninitalized values

argilo:
When using a HackRF, Gqrx often crashes on startup with the following error:

`what():  hackrf_set_freq(1.71799e+10) has failed (-1000) Pipe error`

From the error, it's apparent that Gqrx is requesting a frequency (~17 GHz!) which is not valid for the HackRF. In this case, the HackRF returns an error and gr-osmosdr in turn throws an exception.

I did some digging to find out where the invalid frequency comes from, and it turned out to be an uninitialized value: when `MainWindow::updateFrequencyRange` is first called, `d_hw_freq_start` and `d_hw_freq_stop` are uninitialized. Those variables are not set until `MainWindow::updateHWFrequencyRange` is called.

To fix the problem, I've reordered `DockInputCtl::readSettings` so that `DockInputCtl::ignoreLimitsChanged` (which calls `MainWindow::setIgnoreLimits`, which calls `MainWindow::updateHWFrequencyRange`) occurs before `DockInputCtl::lnbLoChanged` (which calls `MainWindow::setLnbLo`, which calls `MainWindow::updateFrequencyRange`).

By running Gqrx in valgrind, I have verified that this eliminates thousands of usages of uninitialized values. The total number of errors reported by valgrind dropped from 4222 to 7.

I suspect this bug may have been the root cause of #504.

### DIFF
--- a/src/qtgui/dockinputctl.cpp
+++ b/src/qtgui/dockinputctl.cpp
@@ -61,16 +61,16 @@ void DockInputCtl::readSettings(QSettings * settings)
     setIqBalance(settings->value("input/iq_balance", false).toBool());
     emit iqBalanceChanged(ui->iqBalanceButton->isChecked());
 
+    bool_val = settings->value("input/ignore_limits", false).toBool();
+    setIgnoreLimits(bool_val);
+    emit ignoreLimitsChanged(bool_val);
+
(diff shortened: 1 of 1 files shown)


csete:
Excellent, thank you Clayton!

Issue #383, opened 2016-06-02
Gqrx not display line in FFT


rfrail3:
Pulling the 2.5.3 tag version from the repo and compiling without problem in a Debian 7.0 box. The ftt upper line don't appear, but the bottom waterfall yes, as normal.

I tried other releases, from this one until v2.4 but all have the same problem. Maybe is related with some library? It's a bit strange because the bottom right box with the Audio line works ok.

![screen](https://cloud.githubusercontent.com/assets/215429/15757323/5d165328-2905-11e6-84b8-17ca855924cd.png)

```
$ /usr/local/bin/gqrx 
linux; GNU C++ version 4.7.2; Boost_104900; UHD_003.007.002-0-unknown
gr-osmosdr 0.1.3 (0.1.3) gnuradio 3.7.5
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy 
Using Volk machine: avx_64_mmx_orc
BookmarksFile is /home/user33/.config/gqrx/bookmarks.csv
gr-osmosdr 0.1.3 (0.1.3) gnuradio 3.7.5
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy 
Using device #0 Realtek RTL2838UHIDIR SN: 00000001
Found Rafael Micro R820T tuner
```

csete:
You are not the first one to report this extremely weird issue and I have no idea what the error could be. 

Has there never been any plot in the upper range or did it suddenly disappear?

Try playing with the FFT settings and reset the configuration using gqrx -r

rfrail3:
Solved with "gqrx -r"!
Before that, I played with the FFT settings but it didn't had any difference.

Thanks,

csete:
Ok, cool. If it suddenly disappears again and you can reproduce it, do let me know!

reox:
I had the same issue just now. I used gqrx from debian repo (v 5.4.0 20160609) and did not see the FFT. The waterfall was working.
Also `gqrx -r` helped me in this case.
I probably had a configuration file of an older version of gqrx on the system and I suspect some interference there. 
I tested it on another machine with no previously installed gqrx and the fft showed up just like that.

I'm not really sure if this is just a configuration file problem but in my case it looks like it.

delta-alpha:
[...]

PR #719, opened 2020-01-18
Ensure d_fftAvg is initialized

argilo:
From time to time, I run into a bug where the FFT / panadapter does not display. This appears to be the same issue described in #383. I also frequently see participants run into this bug during Capture The Flag competitions involving SDR problems. So far I've been working around the bug by deleting `.config/gqrx/default.conf`.

I dug into this to see what the problem was, and found that the issue can be reliably produced by setting the FFT averaging to 50 in `.config/gqrx/default.conf`:

```
[fft]
averaging=50
```

If the value is different from 50, the problem does not occur. 50 happens to be the default value of the averaging slider:

https://github.com/csete/gqrx/blob/4e6cfd1b7d41d83da97cd88faea4e75d5d4f88f7/src/qtgui/dockfft.ui#L369-L377

The bug occurs because `MainWindow::setIqFftAvg` is the only thing that sets `d_fftAvg`:

https://github.com/csete/gqrx/blob/4e6cfd1b7d41d83da97cd88faea4e75d5d4f88f7/src/applications/gqrx/mainwindow.cpp#L1631-L1635

And the only thing that calls `MainWindow::setIqFftAvg` is `DockFft::on_fftAvgSlider_valueChanged`:

https://github.com/csete/gqrx/blob/4e6cfd1b7d41d83da97cd88faea4e75d5d4f88f7/src/qtgui/dockfft.cpp#L425-L430

So if Gqrx starts up without the value of the slider being changed then `d_fftAvg` is left uninitialized, resulting in undefined behaviour. Valgrind reports hundreds of thousands of warnings in this case, because the following line of code uses the uninitialized `d_fftAvg`:

https://github.com/csete/gqrx/blob/4e6cfd1b7d41d83da97cd88faea4e75d5d4f88f7/src/applications/gqrx/mainwindow.cpp#L1287

To fix the problem, I've set a default value of 0.25 for `d_fftAvg`, which corresponds to the default slider value of 75, which is defined here:

https://github.com/csete/gqrx/blob/4e6cfd1b7d41d83da97cd88faea4e75d5d4f88f7/src/qtgui/dockfft.cpp#L36

In addition, i set the default value of the slider in `dockfft.ui` to match `DEFAULT_FFT_AVG` so that the slider is always updated when the config file has a non-default value.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -63,6 +63,7 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
     ui(new Ui::MainWindow),
     d_lnb_lo(0),
     d_hw_freq(0),
+    d_fftAvg(0.25),
     d_have_audio(true),
     dec_afsk1200(0)
 {
--- a/src/qtgui/dockfft.ui
+++ b/src/qtgui/dockfft.ui
@@ -373,7 +373,7 @@
              <number>100</number>
             </property>
             <property name="value">
-             <number>50</number>
+             <number>75</number>
             </property>
             <property name="orientation">
              <enum>Qt::Horizontal</enum>


csete:
Thank you so much for the detailed analysis and the patch! Yes, it has been a long outstanding and annoying issue. Glad to see it fixed :)

PR #720, opened 2020-01-19
Prevent use of uninitialized values

argilo:
Valgrind reports that `m_Sql` and `m_pixperdb` are used before initialization during Gqrx startup. The `m_Sql` value is read prior to initialization if Gqrx is started without `[receiver] sql_level` set in default.conf, and the `m_pixperdb` value is read prior to initialization if a `sql_level` value is present in default.conf.

I've fixed both errors by setting initial values in the `CMeter` constructor.


csete:
Thanks!

PR #725, opened 2020-01-26
colormaps: Add White Hot, Black Hot and White Hot compressed

devnulling:
This PR adds 3 new color maps:

- White Hot
- Black Hot
- White Hot Compressed 

White Hot Compressed is useful for making high quality spectrum painting screen caps as shown by @drmpeg 's great write up here: https://gist.github.com/drmpeg/31a9a7dd6918856aeb60 

gqrx_colors


csete:
Nice! It looks like we can generate these algorithmically and I think we should do that whenever possible.

devnulling:
Sounds good, I'll push and squash a fixup for this.

PR #727, opened 2020-01-29
Fix FFT overlap calculation

argilo:
Gqrx displays the FFT overlap percentage in the FFT Settings pane. I noticed two errors in the calculation:

1. The calculation doesn't take into account the fact that the FFT interval is rounded down to the nearest millisecond. (For instance, when a rate of 60 fps is requested, the FFT display is updated every `floor(1000 / 60)` = 16 milliseconds, and the actual rate is 62.5 fps.)
2. The computed value is the length of the FFT as a fraction of the FFT interval; it should be the other way around.

I've corrected both of these problems and verified that the overlap percentage now stays between 0% and 100%. (Prior to the change, the percentage often exceeds 100%.) I also checked that the calculation works correctly when decimation is used.

### DIFF
--- a/src/qtgui/dockfft.cpp
+++ b/src/qtgui/dockfft.cpp
@@ -540,16 +540,17 @@ void DockFft::on_cmapComboBox_currentIndexChanged(int index)
 /** Update RBW and FFT overlab labels */
 void DockFft::updateInfoLabels(void)
 {
-    float   rate;
+    float   interval_ms;
+    float   interval_samples;
     float   size;
     float   rbw;
     float   ovr;
-    float   sps;
 
     if (m_sample_rate == 0.f)
         return;
 
-    rate = fftRate();
+    interval_ms = 1000 / fftRate();
+    interval_samples = m_sample_rate * (interval_ms / 1000.0);
     size = fftSize();
 
     rbw = m_sample_rate / size;
@@ -560,10 +561,9 @@ void DockFft::updateInfoLabels(void)
     else
         ui->fftRbwLabel->setText(QString("RBW: %1 MHz").arg(1.e-6 * rbw, 0, 'f', 1));
 
-    sps = size * rate;
-    if (sps <= m_sample_rate)
+    if (interval_samples >= size)
         ovr = 0;
     else
-        ovr = 100 * (sps / m_sample_rate - 1.f);
+        ovr = 100 * (1.f - interval_samples / size);
     ui->fftOvrLabel->setText(QString("Overlap: %1%").arg(ovr, 0, 'f', 0));
 }


csete:
Why should the overlap be limited to 100%? The intention was to express to which extent each sample is reused in subsequent FFT calculations and for sure there is no upper limit to that.

For each FFT update, we need fft_size*fft_rate samples per second. If the sample rate is lower than this we reuse samples from the previous update. If every sample is reused once, the overlap is 100%. If every sample is reused twice, the overlap is 200%, and so on.

Perhaps the word "overlap" is misleading, but I think it is also used the same way by other SDR software.

argilo:
When I googled "FFT overlap" I found several sources defining it the other way, e.g. https://www.spectraplus.com/DT_help/overlap_percentage.htm

argilo:
That said, if other SDR software uses the same definition as Gqrx does, that would be a good argument for keeping it.

argilo:
From what I've seen in screenshots of SDR Console (sdr-radio.com) it also appears to use the FFT overlap calculation I'm proposing: the percentage by which each FFT overlaps the next, varying between 0 and 100%.

argilo:
For instance: https://irp-cdn.multiscreensite.com/46d0be53/dms3rep/multi/Pluto+QO-100+2020-01-16-170753.png

550 kHz sample rate, line rate 40/s, FFT size 64k and the indicated overlap is 79%, which is 1 - (550000/40) / 65536.

remcycles:
Fixing # 1 definitely makes sense.  I'm biased regarding # 2 since GQRX is the first SDR app I learned and the only one I still use, but the existing overlap logic makes more sense to me.  With the proposed change, when an overlap is displayed as 100% I'd have no idea if the current window was overlapping more than just the previous window or not.

argilo:
> when an overlap is displayed as 100% I'd have no idea if the current window was overlapping more than just the previous window or not

That's not correct; you can tell whether more than two windows overlap because the overlap would be greater than 50% in that case. An overlap of 100% would correspond to an infinite number of windows overlapping.

argilo:
Baudline also defines overlap as a percentage of the FFT length:

baudline-overlap

The command line option works the same way:

`-overlap number      Sliding overlap percentage {1 ... 100}`

argilo:
Another source: https://en.wikipedia.org/wiki/Spectrum_analyzer#FFT_overlapping

> Overlapping rate is approximately 80%. An analyzer that utilises a 1024-point FFT process will re-use approximately 819 samples from the previous FFT process.

So far every source I have found defines overlap in terms of two consecutive FFTs, so it seems that Gqrx's current definition is unconventional.

csete:
Okay, I stand corrected with respect to the definition of FFT overlap; however, the existing calculation correctly expresses what was originally intended but should not be called "overlap". Having said that, I'm also fine with changing the calculation to be consistent with the correct meaning of overlap except that I'm not sure it will be consistent with the actual FFT performed in gqrx.

> https://www.spectraplus.com/DT_help/overlap_percentage.htm

As far as I can tell from the description, this program actually overlaps buffers whereas in gqrx we just make sure there are enough samples in one buffer to make an FFT. That's why I'm concerned that changing the overlap calculation might also require changing the FFT algorithm. Am I wrong?

> So far every source I have found defines overlap in terms of two consecutive FFTs

If this is so, then I don't understand this:

> you can tell whether more than two windows overlap because the overlap would be greater than 50% in that case. An overlap of 100% would correspond to an infinite number of windows overlapping.

---

I guess I'm just a bit confused and need to read a bit and understand this. I just don't have very little time and brain capacity right now, so please bear with me... I really want to have this issue clarified and resolved.

argilo:
> this program actually overlaps buffers whereas in gqrx we just make sure there are enough samples in one buffer to make an FFT

That's a good point. Gqrx does not make any effort to ensure that FFTs are performed at regular intervals. But https://github.com/csete/gqrx/pull/728 will move it a lot closer to doing so. Still, the average amount of overlap between consecutive FFTs will be equal to the indicated value, so I think it's still a sensible way to display it.

> I guess I'm just a bit confused and need to read a bit and understand this. I just don't have very little time and brain capacity right now, so please bear with me... I really want to have this issue clarified and resolved.

It seems @remcycles (and perhaps also you) were confused by my comment that "the overlap percentage now stays between 0% and 100%". That's not because the new calculation is less expressive than the old. Here's the relationship between old and new:

old | new
----|----
0% | 0%
100% | 50%
200% | 67%
300% | 75%
400% | 80%
500% | 83%
... | ...
∞ | 100%

The new formula expresses the average percentage of samples that are carried over from one FFT to the next, or the average amount of one FFT that overlaps the next.

csete:
Okay, thanks for the clarification. Let's try this and fix the FFT later.

remcycles:
> That's not correct; you can tell whether more than two windows overlap because the overlap would be greater than 50% in that case.

I see the logic in that (looking at a diagram helped).  The change (and de facto standard) make sense.  And it's a good idea to follow the commonly accepted definition, so I approve.

On the other hand, part of me still prefers the old figure.  From the first article you linked to, describing a 75% overlap:
> Notice that 4 times as many FFT's are performed in the second case; this will increase the time resolution by a factor of 4.

It's not rather intuitive how to do the mental math to convert between these two figures (75% overlap and 4 times the amount of processing).  The old Gqrx behavior reports a value that's much easier to map to extra CPU work and time resolution, which seems important for SDR systems.

argilo:
> It's not rather intuitive how to do the mental math to convert between these two figures (75% overlap and 4 times the amount of processing).

Subtract the overlap from one, then invert. It's a bit harder but not too bad.

> The old Gqrx behavior reports a value that's much easier to map to extra CPU work and time resolution

Agreed that it's a bit simpler, but still you need to add one to the displayed value since it would be 300% for that example.

I think consistency with other software is a more important factor though.

Issue #726, opened 2020-01-27
Waterfall & FFT sometimes update very infrequently


argilo:
When using an rtl_tcp source, the waterfall and FFT sometimes update very infrequently:

![gqrx-slow-waterfall](https://user-images.githubusercontent.com/583749/73150221-41691f80-4093-11ea-8ea7-9804cd2dbabe.png)

The effect is also noticeable (but to a lesser degree) for other input types when the FFT rate is set high.

I had a peek at the code, and this is happening because `rx_fft_c::work` stores samples as they arrive in a circular buffer that is equal in size to the FFT length, and the GUI polls this buffer at regular intervals. But samples aren't necessarily sent into the `work` function at a steady rate. With the rtl_tcp source they tend to arrive in large bursts, and so the GUI fetches the same FFT data many times in a row until the next burst of samples finally arrives.

I'm not sure yet how best to solve this problem.

argilo:
A workaround to get full 60 fps waterfall & FFT with a local RTL-SDR is to set the device string to `rtl=0,buflen=32768`. With a buffer length of 32768 bytes, samples are received 16384 bytes at a time. At the default sample rate of 1800000, a new burst of samples arrives every 9.1 ms, which is sufficient for ~110 fps.

The RTL-SDR's default buffer length is 262144 bytes (=131072 samples), limiting the frame rate to ~13.7 fps.

argilo:
Increasing the length of the circular buffer in `rx_fft_c` might be a reasonable way to let Gqrx cope with samples that arrive in bursts. `rx_fft_c::get_fft_data` could be modified to consume approximately one frame's worth of samples on each invocation (being careful to always leave at least enough samples for one FFT).

argilo:
Closing, since this was fixed in c41d9890b674d5bce468fca2c3631c9bad38c99e.

PR #728, opened 2020-01-29
Increase circular buffer length so FFT updates smoothly

argilo:
This is an attempt at solving #726 and making the higher FFT/waterfall update rates work properly. The main change I've made is to increase the size of the circular buffers used for FFTs. I've increased their sizes to store one second of samples, so that the FFT and waterfall will continue updating even if input samples arrive in bursts up to one second long.

I've changed `get_fft_data` to keep track of how much time has passed since the last invocation and consume a corresponding amount of samples. I added a fudge factor of 1.001 to ensure that samples are consumed slightly faster than they arrive so that the circular buffer does not fill up over time and add unnecessary latency. There's a check to ensure that the circular buffer always finishes with at least enough samples for one FFT.

~Since the audio FFT rate is not adjustable, I changed it from 25 fps to 62.5 fps (16 ms) to match the highest I/Q FFT rate.~

### DIFF
--- a/src/dsp/rx_fft.cpp
+++ b/src/dsp/rx_fft.cpp
@@ -26,11 +26,12 @@
 #include <gnuradio/gr_complex.h>
 #include <gnuradio/fft/fft.h>
 #include "dsp/rx_fft.h"
+#include <algorithm>
 
 
-rx_fft_c_sptr make_rx_fft_c (unsigned int fftsize, int wintype)
+rx_fft_c_sptr make_rx_fft_c (unsigned int fftsize, double quad_rate, int wintype)
 {
-    return gnuradio::get_initial_sptr(new rx_fft_c (fftsize, wintype));
+    return gnuradio::get_initial_sptr(new rx_fft_c (fftsize, quad_rate, wintype));
 }
 
 /*! \brief Create receiver FFT object.
@@ -38,11 +39,12 @@ rx_fft_c_sptr make_rx_fft_c (unsigned int fftsize, int wintype)
  *  \param wintype The window type (see gr::filter::firdes::win_type).
  *
  */
-rx_fft_c::rx_fft_c(unsigned int fftsize, int wintype)
+rx_fft_c::rx_fft_c(unsigned int fftsize, double quad_rate, int wintype)
     : gr::sync_block ("rx_fft_c",
           gr::io_signature::make(1, 1, sizeof(gr_complex)),
           gr::io_signature::make(0, 0, 0)),
       d_fftsize(fftsize),
+      d_quadrate(quad_rate),
       d_wintype(-1)
 {
 
@@ -50,10 +52,12 @@ rx_fft_c::rx_fft_c(unsigned int fftsize, int wintype)
     d_fft = new gr::fft::fft_complex(d_fftsize, true);
 
     /* allocate circular buffer */
-    d_cbuf.set_capacity(d_fftsize);
+    d_cbuf.set_capacity(d_fftsize + d_quadrate);
 
     /* create FFT window */
     set_window_type(wintype);
+
+    d_lasttime = std::chrono::steady_clock::now();
 }
 
 rx_fft_c::~rx_fft_c()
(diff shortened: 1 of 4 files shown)


argilo:
I've reordered the code to first consume samples and then linearize the circular buffer and perform the FFT. This should be a bit more efficient (since the circular buffer will be small at the time it is linearized) and will reduce latency (since the FFT will use samples that were more recently received).

csete:
I will check this as soon as I have some time.

argilo:
When you get around to testing it, using an RTL-SDR with the device string `rtl=0,buflen=1024000` should let you easily compare the old & new behaviour.

kantooon:
I have a modification of the rx_fft code that does not use the circular buffer at all and drops the samples directly into the FFT buffer. It seems faster than the circular buffer on my machine at high FFT sizes, but it requires somewhat different logic in the rest of the application, hence not submitted here. Perhaps it might be useful for you too: https://github.com/qradiolink/qradiolink/blob/64409acd6a4524e8aa498ad03425df4f53abb081/gr/rx_fft.cpp#L95
(note that only complex FFT code is used in that source file)

szpajder:
I'm currently testing this patch and it works great. Thanks @argilo!

I often face this issue when using GQRX with a remote SDR with SoapyRemote over WiFi. Right after I start streaming the FFT looks fine, but only after a couple of minutes it breaks like on your screenshot in #726 and the only way to restore it is to restart the program. Now it's running for several hours straight and the FFT is supersmooth. Great work.

There is also a possibly related issue with the power level meter at the top of the window. Initially it works correctly, but after a couple of minutes it starts to act erroneously - the reading varies wildly between -10 and 0 dBFS in a way which is completely unrelated to the signal level. This also happens only when using a device over network.

csete:
Thanks for the feedback.

What is interesting for me to know is how, if at all, this affects the real-time performance with directly connected devices, both in terms of latency and synchronization between audio and video. We have been lucky so far and no special synchronization code has been necessary.

src/dsp/rx_fft.cpp
```diff
@@ -105,8 +109,13 @@ void rx_fft_c::get_fft_data(std::complex<float>* fftPoints, unsigned int &fftSiz
         return;
     }
 
+    std::chrono::time_point<std::chrono::steady_clock> now = std::chrono::steady_clock::now();
+    std::chrono::duration<double> diff = now - d_lasttime;
+    d_lasttime = now;
+
     /* perform FFT */
-    do_fft(d_cbuf.linearize(), d_cbuf.size());  // FIXME: array_one() and two() may be faster
+    d_cbuf.erase_begin(std::min((unsigned int)(diff.count() * d_quadrate * 1.001), (unsigned int)d_cbuf.size() - d_fftsize));
```
argilo: @csete With respect to your question about latency, I specifically added a fudge factor of 1.001 here to ensure that latency would not build up even if the input sample clock is running fast or the audio sample clock is running slow. I haven't observed any noticeable latency between audio & panadapter.

src/dsp/rx_fft.cpp
```diff
@@ -105,8 +109,13 @@ void rx_fft_c::get_fft_data(std::complex<float>* fftPoints, unsigned int &fftSiz
...
+    do_fft(d_cbuf.linearize(), d_fftsize);  // FIXME: array_one() and two() may be faster
```
argilo: With respect to performance, one possible issue is that increasing the circular buffer size also increases the cost of the `linearize()` here. I haven't found it to be a problem, but it would probably be worth testing performance at high samples rates with & without this change. It may also make sense to reduce the extra capacity in the circular buffer -- I arbitrarily chose one second, but I'm not sure whether there are scenarios where samples arrive in bursts that large. Perhaps it's not unrealistic for TCP connections though...

src/dsp/rx_fft.cpp
argilo: I'd actually expect this change to improve synchronization. Currently, when a burst of input samples arrives the corresponding demodulated audio goes into a buffer, whereas the panadapter immediately jumps ahead to the last input sample.

szpajder: Correct. When the waterfall problem occurs, it is often accompanied with an audio delay (like half a second or so). With the patch, sync has been perfect for me so far (again: when using a remote device, as this is my primary way of using GQRX; haven't tried it with a local device yet).

src/dsp/rx_fft.cpp
argilo: I see that `linearize()` can be eliminated by replacing it with `array_one()` and `array_two()` so I'll take a stab at that.

argilo: I added a second commit which removes the `linearize()` calls. There is no appreciable difference in performance, even for the largest FFT size, because they only accounted for a very small fraction of the execution time.

argilo:
I also reverted the change to the audio FFT rate since it's not directly related to the other changes I made here, and we probably don't want to increase the speed of the audio waterfall, since it's normally small.

I actually didn't know there was an audio waterfall until @devnulling mentioned it in https://github.com/csete/gqrx/issues/741. :smile:

argilo:
> What is interesting for me to know is how, if at all, this affects the real-time performance with directly connected devices, both in terms of latency and synchronization between audio and video. We have been lucky so far and no special synchronization code has been necessary.

Synchronization between audio and video is working beautifully for me with various local devices. I tested with an RTL-SDR (with both large and small buffer sizes), a HackRF (at 20 Msps) and a USRP B200 (at 16 Msps and 250 ksps). I also tried with various FFT sizes.

csete:
Okay, it seems to work well for me too. I have applied the commits manually to resolve the conflicts with some of the later changes.

PR #729, opened 2020-01-31
Fix various bugs with waterfall time span

argilo:
I noticed several issues with the waterfall time span feature:

1. The resolution display shows "Res: - s" when Gqrx starts, instead of the current time resolution.
2. The indicated time resolution is incorrect when the time span is set to Auto.
3. When the time span is set to Auto, the indicated time resolution does not update if the FFT rate is changed.
4. The indicated time resolution does not update when the window is resized.
5. The time span selection is not saved in the settings file.

This pull request should hopefully address all these issues.


csete:
Thanks for the fixes!

PR #732, opened 2020-02-01
Misc improvements to the DSP chain

smunaut:
Ideally the rotator would be replaced with a freq_xlating_fir filter doing pre-decimation before the actual RX chain, but that's a more intrustive change and the 'rotator' only would still need to be a supported option anyway (for low bandwidth SDRs), so this is a good first step.

When the second step becomes reality, the update_ddc method will be a bit more filled, properly selecting which path to use and reconfiguring the filter decimation and taps.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -28,10 +28,6 @@
 
 #include <iostream>
 
-#if GNURADIO_VERSION < 0x030800
-#include <gnuradio/blocks/multiply_const_ff.h>
-#endif
-
 #include <gnuradio/prefs.h>
 #include <gnuradio/top_block.h>
 #include <osmosdr/source.h>
@@ -119,10 +115,8 @@ receiver::receiver(const std::string input_device,
     iq_sink->set_unbuffered(true);
     iq_sink->close();
 
-    rx = make_nbrx(d_quad_rate, d_audio_rate);
-    lo = gr::analog::sig_source_c::make(d_quad_rate, gr::analog::GR_SIN_WAVE,
-                                        0.0, 1.0);
-    mixer = gr::blocks::multiply_cc::make();
+    rx  = make_nbrx(d_quad_rate, d_audio_rate);
+    rot = gr::blocks::rotator_cc::make(0.0);
 
     iq_swap = make_iq_swap_cc(false);
     dc_corr = make_dc_corr_cc(d_quad_rate, 1.0);
@@ -351,7 +345,7 @@ double receiver::set_input_rate(double rate)
     d_quad_rate = d_input_rate / (double)d_decim;
     dc_corr->set_sample_rate(d_quad_rate);
     rx->set_quad_rate(d_quad_rate);
-    lo->set_sampling_freq(d_quad_rate);
+    update_ddc();
     tb->unlock();
 
     return d_input_rate;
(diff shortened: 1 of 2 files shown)


argilo:
The simplification of `receiver::connect_all` is great! I tested various options and everything still seems to work properly.

While switching to a rotator *should* be more efficient, it's actually much slower in GNU Radio 3.8 because the rotator was broken in VOLK 2.0.0 (https://github.com/gnuradio/gnuradio/issues/2748, https://github.com/gnuradio/volk/pull/278) and instead of patching VOLK, GNU Radio worked around the problem by disabling VOLK in the rotator block (https://github.com/gnuradio/gnuradio/pull/3062).

I'm not sure what the best course of action is for Gqrx.

smunaut:
Pfff ... I didn't know about that GR issue ... wtf ... freq shift is a basic function, that should have been fixed within the day ....

At least on my machine using even a non-volk rotator still improves performance at least.

argilo:
Running a HackRF at 8 Msps and default Gqrx settings uses about 255% CPU on the `master` branch, and 360% CPU on the `dsp-improvements` branch.

argilo:
If I revert https://github.com/gnuradio/gnuradio/pull/3062 and build GNU Radio with a patched VOLK instead, CPU utilization is about 221%.

smunaut:
Huh that's weird ... I don't get this at all here.
Reading an IQ file source, I have 115% current code, then ~ 110% with dsp-improvements and then ~100% with patched volk / gr.
But I probably have -mnative in my compiler flags so even the 'generic' rotator probably has been somehwat SIMD'd by the compiler on its own.

csete:
I like the change and will be testing this.

If I recall correctly, gqrx used the freq_xlating_fir in the beginning, but then I switched to a separate decimator which worked better on ARM, but it's long time ago.

smunaut:
I think making a 'ddc' subblock would make sense which would automatically pick the right filter architecture. ( single or double filter stage and using FIR or FFT filters depending on which ratio of input/output sample rate is needed ). With high bandwidht SDR and narrow channels multi stage helps a lot. (with the first stage being a _very_ loose filter)

csete:
@smunaut that would be great :)

I have now tested this and I also get some performance improvement, something like 410% -> 377% (GNU Radio 3.8 and Volk 2.0.0 built from source). But I am unsure whether we should merge this now or wait until something is fixed... Please advise :)

argilo:
> But I am unsure whether we should merge this now or wait until something is fixed...

There's a small risk that Ubuntu 20.04 will end up with worse performance, but it seems to hinge on compiler settings. I don't think it's worth holding up progress in Gqrx because of an issue that GNU Radio should fix on their end, so I think this should be merged.

smunaut:
The GR Debian maintainer routinely includes "out-of-band" patches in the packages so I'm sure we could get it to add a patch to fix Volk 2.0.0 and to re-enabled VOLK rotator in gnuradio package.

argilo:
> The GR Debian maintainer routinely includes "out-of-band" patches in the packages so I'm sure we could get it to add a patch to fix Volk 2.0.0 and to re-enabled VOLK rotator in gnuradio package.

Good point. Debian testing is already using VOLK 2.1.0, so I could suggest that they revert https://github.com/gnuradio/gnuradio/pull/3062/commits/544bd152649245c2a5329d3512b473b3e5d9e201 in their gnuradio package.

csete:
Okay, I'm merging.

argilo:
> I think making a 'ddc' subblock would make sense which would automatically pick the right filter architecture.

I did some experimenting in GNU Radio Companion, and it looks like just replacing the current rotator block with a decimating FFT filter followed by a rotator (like GNU Radio's "Frequency Xlating FFT Filter" Python block) would already be a big improvement. It don't expect that would be much work to implement.

PR #733, opened 2020-02-01
Fix FM deemphasis by opting in to new-style taps

argilo:
The audio produced by Gqrx's FM demodulators is 20-30 dB quieter than it should be. I dug into this and found that this is caused by a bug in FM deemphasis. When the code was copied over from GNU Radio, the `oldstyle=False` parameter for `iir_filter_ffd` was omitted. Since `oldstyle` defaults to True, the filter taps were misinterpreted. Here's the corresponding code in GNU Radio:

https://github.com/gnuradio/gnuradio/blob/9ed2fde11337edab2629195b45de2d1338a5f703/gr-analog/python/analog/fm_emph.py#L128

After this change, the audio level for the FM demodulators is comparable to the other demodulators.

### DIFF
--- a/src/dsp/rx_demod_fm.cpp
+++ b/src/dsp/rx_demod_fm.cpp
@@ -63,7 +63,7 @@ rx_demod_fm::rx_demod_fm(float quad_rate, float max_dev, double tau)
     d_fftaps.resize(2);
     d_fbtaps.resize(2);
     calculate_iir_taps(d_tau);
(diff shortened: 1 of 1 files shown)


argilo:
It looks like this bug probably crept in with 970591e1b44c3817c96eba6ec6ce2fcc59d8102e, which was meant to replicate the changes made in https://github.com/gnuradio/gnuradio/pull/777.

csete:
Nice catch, thanks!

PR #734, opened 2020-02-02
Speed up FFT & waterfall by deferring log calculations

argilo:
I did some profiling to see what opportunities for optimization there might be. A lot of CPU time is spent in `MainWindow::iqFftTimeout`, especially when using large FFT sizes, so this is a useful function to focus on.

About a third of the CPU time in `MainWindow::iqFftTimeout` is spent in `log10f`, which converts the FFT bin powers to dB. When the FFT is larger than the plot width, only a fraction of these dB values are actually displayed to the user. So we can save a lot of processing time by computing the dB value only for the FFT bins that are displayed. To do this, I've moved the dB calculation to the end of `CPlotter::getScreenIntegerFFTData`.

One side effect of this change is that the FFT averaging is now computed over the FFT bin powers instead of the dB values. This is probably a more sensible way to average anyway.

After making this change, the CPU time spent in `MainWindow::iqFftTimeout` dropped from 71 ms to 47 ms per invocation, when using an FFT size of 1048576.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -110,10 +110,9 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
 
     d_fftData = new std::complex<float>[MAX_FFT_SIZE];
     d_realFftData = new float[MAX_FFT_SIZE];
-    d_pwrFftData = new float[MAX_FFT_SIZE]();
     d_iirFftData = new float[MAX_FFT_SIZE];
     for (int i = 0; i < MAX_FFT_SIZE; i++)
-        d_iirFftData[i] = -140.0;  // dBFS
+        d_iirFftData[i] = 0;
 
     /* timer for data decoders */
     dec_timer = new QTimer(this);
@@ -382,7 +381,6 @@ MainWindow::~MainWindow()
     delete [] d_fftData;
     delete [] d_realFftData;
     delete [] d_iirFftData;
-    delete [] d_pwrFftData;
     delete qsvg_dummy;
 }
 
@@ -1300,7 +1298,7 @@ void MainWindow::iqFftTimeout()
 
         /* calculate power in dBFS */
         pwr = pwr_scale * (pt.imag() * pt.imag() + pt.real() * pt.real());
-        d_realFftData[i] = 10.0 * log10f(pwr + 1.0e-20);
+        d_realFftData[i] = pwr;
 
         /* FFT averaging */
         d_iirFftData[i] += d_fftAvg * (d_realFftData[i] - d_iirFftData[i]);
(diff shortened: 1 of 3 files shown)


src/applications/gqrx/mainwindow.cpp
```diff
@@ -110,10 +110,9 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
 
     d_fftData = new std::complex<float>[MAX_FFT_SIZE];
     d_realFftData = new float[MAX_FFT_SIZE];
-    d_pwrFftData = new float[MAX_FFT_SIZE]();
```
argilo: I noticed that this variable was unused, so I removed it.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1586,6 +1584,8 @@ void MainWindow::setIqFftSize(int size)
 {
     qDebug() << "Changing baseband FFT size to" << size;
     rx->set_iq_fft_size(size);
+    for (int i = 0; i < size; i++)
+        d_iirFftData[i] = 0;
```
argilo: I noticed that the averaged FFT data was not cleared when changing FFT sizes, which results in spurious data being displayed until the old data fades away. This fixes the problem by zeroing the averages when the FFT size is changed.

argilo:
i got a further 17ms speedup just by moving an expensive division outside the loop that calculates the X coordinates of the FFT bins, and the execution time of `MainWindow::iqFftTimeout` is now under 30ms. I've added that change in a second commit.

csete:
Nice, but I think this might require changing the averaging algorithm to keep the existing behavior. Will try to capture a video tomorrow.

argilo:
> I think this might require changing the averaging algorithm to keep the existing behavior. Will try to capture a video tomorrow.

Thanks for testing. I'd like to preserve the existing behaviour, so I'll be happy to look into it if something has changed.

argilo:
I rebased on master, since a lot of changes were merged since I opened this pull request.

csete:
> One side effect of this change is that the FFT averaging is now computed over the FFT bin powers instead of the dB values. This is probably a more sensible way to average anyway.

I would be if the averaged data were used for computation, but here the only purpose of averaging is to enhance the visual appearance and we display only dB values, so this change affects the visual appearance.

I have tried to capture a [video](https://www.youtube.com/watch?v=euMAKJuUoJw) showing the visual difference after this change. The original is on the left side, the new one is on the right. Both are using the default averaging constant, however, since the averaging is now done before the log power calculations, there is now a very slow decay of CW signals that IMO doesn't look good.

Not sure how to solve this without averaging the dB values or introducing separate attack and decay constants.

argilo:
@csete It looks like you pasted the wrong YouTube link.

csete:
Haha, yes, here is the correct link: https://www.youtube.com/watch?v=56atwqBSVgI

argilo:
Thanks for that. I'll think about this and see if there's a way to keep the original behaviour while avoiding the expensive `log10f`s.

argilo:
I'm closing this for now, since I agree the change in behaviour is undesirable. I'll open a new PR if I come up with another way to improve performance for large FFT sizes.

argilo:
#770 should deliver a comparable performance improvement without any change to the averaging behaviour.

PR #735, opened 2020-02-02
Set RTL-SDR gain to midpoint by default

argilo:
When starting up Gqrx with an RTL-SDR for the first time, the gain is set to the maximum. (For the reasoning behind this, see aa4c842af3789f77736302666657088bc2b308c1.) While this is better than starting at zero gain, it is very likely that the user will experience distortion since the ADC will probably be overloaded. To improve the new-user experience, I propose to default to the midpoint instead. For R820T-based dongles this is 24.8 dB.


csete:
I guess I overreacted with the max gain :)

PR #736, opened 2020-02-02
Increase default audio gain to -6 dB

argilo:
Gqrx currently has a default audio gain of -20 dB. I find I almost always have to increase it, since the default produces quieter audio than other applications. To improve the new-user experience, I propose to change the default to -6 dB.


devnulling:
+1 this seems like a good change

PR #737, opened 2020-02-02
Lock panadapter and waterfall together by default

argilo:
in 50f3e4660566288e907b55032bafa8587661711c dB ranges for the panadapter and waterfall were separated. Then in https://github.com/csete/gqrx/pull/416 @alexf91 added a "Lock" button to lock the ranges to each other. Currently the lock is off by default.

I always turn the lock setting on, and I believe defaulting to on would be sensible for new users. Thus I'm proposing to make that the default. I'm open to dissenting opinions on this.

### DIFF
--- a/src/qtgui/dockfft.cpp
+++ b/src/qtgui/dockfft.cpp
@@ -270,9 +270,9 @@ void DockFft::saveSettings(QSettings *settings)
 
     // pandapter and waterfall ranges locked together
     if (ui->lockButton->isChecked())
-        settings->setValue("db_ranges_locked", true);
-    else
         settings->remove("db_ranges_locked");
+    else
+        settings->setValue("db_ranges_locked", false);
 
     if (QString::compare(ui->cmapComboBox->currentData().toString(), DEFAULT_COLORMAP))
         settings->setValue("waterfall_colormap", ui->cmapComboBox->currentData().toString());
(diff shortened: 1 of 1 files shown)


alexf91:
I agree with this. It bothers me every time I reset the configuration.

csete:
I keep it always unlocked, but I don't have any statistics on what people prefer. It is my impression though, that many people like to have the noise floor in the fft plot about 1/3 above the frequency axis, which would make te waterfall too noisy.

In any case, I don't think it's a good idea to change this kind of default settings because people who are using the default would suddenly experience a change in the application when they upgrade.

Feel free to try and convince me otherwise...

argilo:
> I don't have any statistics on what people prefer.

Likewise. All I know at this point is my own preference, plus yours and @alexf91's.

> It is my impression though, that many people like to have the noise floor in the fft plot about 1/3 above the frequency axis, which would make te waterfall too noisy.

I personally keep the noise floor just above the numbers in the frequency axis.

> I don't think it's a good idea to change this kind of default settings because people who are using the default would suddenly experience a change in the application when they upgrade.

I thought a change in the default might be worth it if a lot of people preferred using the lock. But your preference has convinced me that might not be the case.

This change is of course a matter of taste, and I'm happy to leave the default as it is.

argilo:
I'm closing this off, since support for changing the default isn't unanimous.

PR #739, opened 2020-02-02
Decrease default FFT range to (-120, -20)

argilo:
The default FFT dB range is very large (from -135 to 0), which tends to leave a lot of empty space at the top & bottom of the panadapter. I think it might make sense to narrow this a bit, so I'm proposing -120 to -20. This seems to give reasonable results for an RTL-SDR and a USRP B200:

Screenshot from 2020-02-02 18-23-19

Screenshot from 2020-02-02 18-22-44


csete:
I guess it's better for most devices...

PR #744, opened 2020-02-05
Fix FFT split bugs, and display audio FFT by default

argilo:
While investigating @devnulling's proposal to display the audio FFT by default in https://github.com/csete/gqrx/issues/741, I noticed a couple bugs in the FFT split functionality:

* The I/Q FFT split slider defaults to 50, while the default in dockfft.cpp is 35 and the default in plotter.cpp is 30. As a result, the split will be set to 30 if the value saved in the configuration file happens to be 50. I've corrected this by setting all three values to 35.
* The audio FFT slider and the I/Q FFT slider work in opposite directions. I've changed the direction of the audio FFT slider to match the I/Q FFT slider.

I also implemented the suggestion to display the audio FFT by default, and set the default to 65. If you don't like that change, we could make the default 100. I happen to like 65 though.

### DIFF
--- a/src/qtgui/audio_options.cpp
+++ b/src/qtgui/audio_options.cpp
@@ -93,17 +93,17 @@ void CAudioOptions::setUdpPort(int port)
 
 void CAudioOptions::setFftSplit(int pct_2d)
 {
-    ui->fftSplitSlider->setValue(100 - pct_2d);
+    ui->fftSplitSlider->setValue(pct_2d);
 }
 
 int  CAudioOptions::getFftSplit(void) const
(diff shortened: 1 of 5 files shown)


csete:
One of the reasons why I chose to hide the waterfall by default is because there is very little room for the audio spectrum plot with the default layout and window size. So I prefer to keep it with waterfall hidden (I will change that back after merging).

argilo:
> So I prefer to keep it with waterfall hidden (I will change that back after merging).

That's fine by me. :+1:

PR #745, opened 2020-02-05
Fix crash when switching device to USRP

### LINKED ISSUE #125: Incorrect gain settings (crash with usrp & bladerf)
When switching devices gqrx will try so set/get the gains of the previous device. In some cases this will lead to an exception such as te one discussed here:
https://groups.google.com/forum/#!msg/gqrx/h69vMymU8-c/D2OZXYYFm3EJ

USRP and Bladerf devices are known to be affected by this.

argilo:
Fixes #125.

When switching from an RTL-SDR to a USRP B200 in the device chooser, Gqrx crashes with the following error:

> set_rx_gain: gain "LNA" not found for channel 0. Available gains: PGA

This occurs because the RTL-SDR's LNA gain value is loaded from the settings file and then set, even though the USRP does not have an LNA gain. When asked to set a non-existent gain, UHD throws an exception.

To solve the problem, I modified `DockInputCtl::readSettings` to check the existence of a gain before attempting to set it. Since `MainWindow::loadSettings` updates the gain stages prior to calling `DockInputCtl::readSettings`, `DockInputCtl` is able to check which gains are valid for the current device.

### DIFF
--- a/src/qtgui/dockinputctl.cpp
+++ b/src/qtgui/dockinputctl.cpp
@@ -96,8 +96,8 @@ void DockInputCtl::readSettings(QSettings * settings)
 
             gain_name = gain_iter.key();
             gain_value = 0.1 * (double)(gain_iter.value().toInt());
-            setGain(gain_name, gain_value);
-            emit gainChanged(gain_name, gain_value);
+            if (setGain(gain_name, gain_value))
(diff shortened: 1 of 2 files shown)


csete:
Nice fix, thanks!.

PR #746, opened 2020-02-06
Fix errors in plotter sizes calculations

argilo:
I stumbled across a couple tiny bugs which prevent the plotter from occupying the available space, leaving a blank pixel on the right and bottom.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -850,7 +850,7 @@ void CPlotter::resizeEvent(QResizeEvent* )
         m_2DPixmap = QPixmap(m_Size.width(), fft_plot_height);
         m_2DPixmap.fill(Qt::black);
 
-        int height = (100 - m_Percent2DScreen) * m_Size.height() / 100;
+        int height = m_Size.height() - fft_plot_height;
         if (m_WaterfallPixmap.isNull())
         {
             m_WaterfallPixmap = QPixmap(m_Size.width(), height);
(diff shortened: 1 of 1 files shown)


src/qtgui/plotter.cpp
```diff
@@ -850,7 +850,7 @@ void CPlotter::resizeEvent(QResizeEvent* )
         m_2DPixmap = QPixmap(m_Size.width(), fft_plot_height);
         m_2DPixmap.fill(Qt::black);
 
-        int height = (100 - m_Percent2DScreen) * m_Size.height() / 100;
```
argilo: Here the integer division rounds the result down, typically resulting in a blank pixel at the bottom of the waterfall.

src/qtgui/plotter.cpp
```diff
@@ -1174,7 +1174,7 @@ void CPlotter::getScreenIntegerFFTData(qint32 plotHeight, qint32 plotWidth,
         for (i = minbin; i < maxbin; i++)
             m_pTranslateTbl[i] = ((qint64)(i-m_BinMin)*plotWidth) / (m_BinMax - m_BinMin);
         *xmin = m_pTranslateTbl[minbin];
-        *xmax = m_pTranslateTbl[maxbin - 1];
```
argilo: `xmax` is supposed to be the first unused column, not the last used column. The off-by-one prevents the rightmost column from being used in the `largeFft` case. (It's used once you zoom in and reach the `else` case below, which returns the correct `xmax` value.)

PR #747, opened 2020-02-06
Use /dev/urandom to avoid blocking

argilo:
Gqrx uses a random file as a dummy input in certain circumstances:

https://github.com/csete/gqrx/blob/490ee3be4f1c5e0319b7610912104e2d0a47fd4e/src/applications/gqrx/receiver.cpp#L80-L83

Currently `get_random_file` returns `/dev/random` if it exists. That's problematic on Linux, because `/dev/random` consumes bytes directly from the entropy pool and blocks once it's empty. This device can typically provide only a few bytes per second. As a result, Gqrx becomes unresponsive if the random input is ever used. (I hit this case a few times while testing other changes.)

To solve the problem, I propose to replace `/dev/random` with `/dev/urandom`, which is guaranteed not to block since it uses a pseudorandom number generator.

### DIFF
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -32,7 +32,7 @@ std::string receiver::get_random_file(void)
     static std::string path;
     if (path.empty())
     {
-        path = "/dev/random";
+        path = "/dev/urandom";
         QFileInfo checkFile(QString::fromStdString(path));
         if (!checkFile.exists())
         {


argilo:
@csete After looking at this code, I was of the mind to get rid of `get_random_file` and `get_null_file` and just leave `src`, `iq_sink` and `wav_sink` null while not in use. I'll do this in a follow-up PR if it makes sense, but I wanted to make sure I wasn't missing some important reason why the dummy inputs & output blocks are needed.

csete:
> @csete After looking at this code, I was of the mind to get rid of get_random_file and get_null_file and just leave src, iq_sink and wav_sink null while not in use. I'll do this in a follow-up PR if it makes sense, but I wanted to make sure I wasn't missing some important reason why the dummy inputs & output blocks are needed.

The only reason I can think of is that we need all the blocks to be available if the application is running, but to be honest, I don't remember anything about this code. It just reminds me how horribly broken gqrx is, haha.

PR #748, opened 2020-02-08
Stereo UDP streaming

alexf91:
Adds optional stereo UDP streaming. This topic occasionally comes up on the mailing list.
The UDP sink contains an interleaver that is used when stereo streaming is enabled. I'm not exactly sure that this is the best way, but it does the job. Maybe the logic should be contained in the receiver and not the UDP sink class.
I'm open for suggestions.

stereo_udp


src/applications/gqrx/receiver.cpp
```diff
@@ -1110,16 +1114,20 @@ receiver::status receiver::stop_audio_playback()
 }
 
 /** Start UDP streaming of audio. */
-receiver::status receiver::start_udp_streaming(const std::string host, int port)
+receiver::status receiver::start_udp_streaming(const std::string host, int port, bool stereo)
 {
-    audio_udp_sink->start_streaming(host, port);
+    stop();
+    audio_udp_sink->start_streaming(host, port, stereo);
+    start();
```
argilo: The addition of `stop()` and `start()` here has introduced a bug: pressing the UDP button will start up the receiver even if it was not running at the time.

argilo: I don't think there's any need to stop and start the flow graph, you just need to add `lock()` and `unlock()` around the parts that reconfigure the connections in the UDP sink.

argilo:
I did a quick test and stereo worked beautifully with `nc -l -u localhost 7355 | aplay -r 48000 -f S16_LE -t raw -c 2`. The stereo setting is also correctly saved and loaded.

The following fix for the UDP button bug seems to work well:
```diff
diff --git a/src/applications/gqrx/receiver.cpp b/src/applications/gqrx/receiver.cpp
index 1a51d4f73a..3c1fa737e8 100644
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -1116,18 +1116,14 @@ receiver::status receiver::stop_audio_playback()
 /** Start UDP streaming of audio. */
 receiver::status receiver::start_udp_streaming(const std::string host, int port, bool stereo)
 {
-    stop();
     audio_udp_sink->start_streaming(host, port, stereo);
-    start();
     return STATUS_OK;
 }
 
 /** Stop UDP streaming of audio. */
 receiver::status receiver::stop_udp_streaming()
 {
-    stop();
     audio_udp_sink->stop_streaming();
-    start();
     return STATUS_OK;
 }
 
diff --git a/src/interfaces/udp_sink_f.cpp b/src/interfaces/udp_sink_f.cpp
index d711ba9960..621074a26b 100644
--- a/src/interfaces/udp_sink_f.cpp
+++ b/src/interfaces/udp_sink_f.cpp
@@ -78,6 +78,7 @@ udp_sink_f::~udp_sink_f()
  */
 void udp_sink_f::start_streaming(const std::string host, int port, bool stereo)
 {
+    lock();
     disconnect_all();
 
     if (stereo)
@@ -93,15 +94,19 @@ void udp_sink_f::start_streaming(const std::string host, int port, bool stereo)
         connect(d_f2s, 0, d_sink, 0);
         connect(self(), 1, d_null0, 0);
     }
+    unlock();
+
     d_sink->connect(host, port);
 }
 
 
 void udp_sink_f::stop_streaming(void)
 {
+    lock();
     disconnect_all();
     connect(self(), 0, d_null0, 0);
     connect(self(), 1, d_null1, 0);
+    unlock();
 
     d_sink->disconnect();
 }
```

alexf91:
Thanks. There's still different behavior between the UDP button and the Rec button in the audio dock widget. The UDP button stays enabled when DSP is not running, but the Rec button is disabled. I'll fix this tomorrow.

argilo:
> There's still different behavior between the UDP button and the Rec button in the audio dock widget. The UDP button stays enabled when DSP is not running, but the Rec button is disabled. I'll fix this tomorrow.

It seems like that problem isn't directly related to this pull request, so it could be done separately. It would be nice if the Rec button worked regardless of whether the DSP was running or stopped, just like the UDP button.

PR #749, opened 2020-02-09
Explicitly set antenna in readSettings

argilo:
When using a device with multiple antenna inputs (e.g. USRP B200), Gqrx forgets to set the antenna input when switching from I/Q playback back to hardware input. To fix this, I've added an `emit`, like all the other settings already have.


csete:
Nice, thanks!

PR #750, opened 2020-02-09
Simplify I/Q and WAV recording code

argilo:
It's simpler to create I/Q and WAV file sinks at the time they're needed, rather than building them in the receiver's constructor and setting their outputs to /dev/null. This eliminates the need for `receiver::get_null_file`.

I verified that I/Q and WAV recording still work, and also tested the error handling by replacing the filenames with `""` in the code temporarily. All appears to work as intended.


csete:
This is nice, thanks.

PR #751, opened 2020-02-09
Handle the error when set_input_device fails

argilo:
If the configured input device is not plugged in when launching Gqrx, the program simply crashes. By catching the exception in `receiver::set_input_device` we can handle the situation gracefully and give the user an opportunity to select another device without having to launch Gqrx for a second time.


csete:
I'm merging it because it's working. However, I need to mention that I really hate exceptions and would prefer to avoid them as much as possible. In this case, set_input_device could return an error code instead of throwing an exception and it would even require fewer lines of code.

argilo:
> In this case, set_input_device could return an error code instead of throwing an exception and it would even require fewer lines of code.

I originally wrote things that way, but switched to an exception so that the full error message could be passed out to mainwindow.cpp and displayed to the user. Let me know if you have an idea for how to handle this more elegantly.

PR #752, opened 2020-02-10
Implement DDC using a frequency xlating FIR filter

argilo:
This is a follow-up to @smunaut's [suggestion to implement a DDC block](https://github.com/csete/gqrx/pull/732#issuecomment-581122507). Here I've implemented it using an FFT filter followed by a rotator, like GNU Radio's [Frequency Xlating FFT Filter block](https://github.com/gnuradio/gnuradio/blob/master/gr-filter/python/filter/freq_xlating_fft_filter.py).

The receiver requests a decimation rate that will keep the DDC's output above 1 Msps. Since the demodulators need at most 240 ksps, this leaves plenty of room for filter roll-off so that the number of taps in the FFT filter can be kept low.

if the requested decimation is 1, then the FFT filter is left out of the chain.

Prior to this change, I could use my BladeRF up to 20 Msps before overruns started to occur. After the change, I got up to 34 Msps.

@smunaut's suggestion of multiple decimation stages could improve performance further, but I thought I'd start with a simple design for now.

### DIFF
--- a/gqrx.pro
+++ b/gqrx.pro
@@ -103,6 +103,7 @@ SOURCES += \
     src/dsp/agc_impl.cpp \
     src/dsp/correct_iq_cc.cpp \
     src/dsp/filter/fir_decim.cpp \
+    src/dsp/downconverter.cpp \
     src/dsp/fm_deemph.cpp \
     src/dsp/lpf.cpp \
     src/dsp/rds/decoder_impl.cc \
@@ -157,6 +158,7 @@ HEADERS += \
     src/dsp/correct_iq_cc.h \
     src/dsp/filter/fir_decim.h \
     src/dsp/filter/fir_decim_coef.h \
+    src/dsp/downconverter.h \
     src/dsp/fm_deemph.h \
     src/dsp/lpf.h \
     src/dsp/rds/api.h \
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -50,6 +50,7 @@
 #endif
 
 #define DEFAULT_AUDIO_GAIN -6.0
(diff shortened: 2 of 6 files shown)


argilo:
Because the number of taps stays relatively small, an FIR filter turned out to have better performance than an FFT filter. By switching to an FIR filter, I was also able to use GNU Radio's Frequency Xlating FIR Filter block, which simplifies the code. I've updated this pull request to use that approach.

After the change, I'm able to use a BladeRF at its maximum rate of 40 Msps.

argilo:
With the latest change I can also operate a USRP B200 at 56 Msps (provided that the receive buffers are increased by appending `,recv_frame_size=16360,num_recv_frames=128` to the device string).

argilo:
The recent changes have cause some merge conflicts. I'll resolve those and push an update shortly.

argilo:
I've rebased on master and this still appears to be working correctly.

argilo:
Rebase again to resolve merge conflicts. I also added the new files to gqrx.pro.

csete:
Thanks for updating.

csete:
Ups... it seems that this change has introduced a latency in the FFT plot. Using Airspy at 2.5 MSPS. When changing the RF frequency the audio is updated immediately, however, the FFT plot is ~1 sec delayed.

csete:
Btw. I am using decimation=1.

argilo:
OK, feel free to revert this and I can investigate. I have an AirSpy so it should be no problem to replicate your setup.

csete:
No rush from my side. I will open a bug and wait to see if we can find a fix within a reasonable time before reverting it :)

argilo:
Sounds good. I've replicated the bug. Will get a pull request opened up as soon as I track down the problem.

PR #753, opened 2020-02-17
Version 2.13 release candidate 

argilo:
I've merged all the recent pull requests into a single branch, resolved merge conflicts, and updated the news.txt file. If this branch works well, then I propose to release it as version 2.13.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,11 +1,21 @@
-    2.12.2: In progress...
 
-     FIXED: FM de-emphasis causing audio to be 20 dB quieter than should be.
+      2.13: Released February 17, 2020
+
+       NEW: Stereo option for UDP streaming.
+     FIXED: FM de-emphasis causing audio to be 20 dB quieter than it should be.
      FIXED: Update waterfall time resolution when FFT settings are changed.
      FIXED: Update waterfall time resolution when window is resized.
      FIXED: Restore waterfall time span between sessions.
      FIXED: FFT buffer overlap calculation.
-  IMPROVED: Software tuning.
+     FIXED: Crash when launching without device connected.
+     FIXED: Crash when setting invalid RF gains.
+     FIXED: Audio panadapter / waterfall slider direction.
+  IMPROVED: DSP performance.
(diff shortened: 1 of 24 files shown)


argilo:
@alexf91 @devnulling @smunaut If you have a chance to test out this branch, I'd appreciate your feedback.

argilo:
I added one more significant performance boost: `drawPolygon` can be very slow (>100 ms) when zooming in on the dB axis (especially when the panadapter is large). The time seems to be spent on determining which pixels are inside the polygon. To get around this, I've replaced `drawPolygon` with `drawPolyline` and filled the polygon using one `drawLine` call per column. This reduces the drawing time to <10 ms.

argilo:
I sped up the panadapter even more by switching from `drawLine` to `fillRect`.

csete:
The title of this pull request suggests that it has already been decided that there will be a version 2.13 and this will be its contents. However, no such decision has been made.

For the time being, I am still the maintainer of gqrx and I prefer to make my own decisions about what to release in which version and when to release it. I will be happy to have a conversation about possible releases (as time permits) but I am not going to be dictated version numbers, release schedules or its contents.

I just told you in a private email that my entire life is upside down and the last thing I need right now is to worry about is rushing out a release because of Ubuntu's release schedule. I don't expect anybody to understand or empathize with my situation, but I expect everybody to hear and respect what I say: I will review the proposed changes when I can or want to, and I will do that following the engineering methods that I have been practicing for decades. This includes reviewing, understanding, and testing every single change, regardless of how much time it takes. I need a certain level of confidence and test coverage for as many platforms and use cases as possible.

> I added one more significant performance boost ...
> I sped up the panadapter even more ...

Please put these new changes in their own pull requests where they can be properly reviewed and tested on their own.

argilo:
I'm sorry for approaching this the wrong way. My intention just was to batch the recent pull requests for easier testing; I had done this locally since there are merge conflicts between them and i thought it would be helpful to others. I threw in a proposed version number in the hopes that it might take some load off your shoulders, and didn't mean to dictate anything.

I'll be happy to put the latest changes into their own pull request.

PR #754, opened 2020-02-19
Speed up FFT plotting

argilo:
I noticed that drawing the panadapter can be slow, particularly when zooming in on the dB axis. Most of the time is taken up by `drawPolygon`, which seems to spend a long time determining which pixels are inside the polygon. To avoid this, I've replaced `drawPolygon` with `drawPolyline` and filled the area below the line using `fillRect` calls (one per column). When the dB axis is fully zoomed and the panadapter is expanded to the full screen, this reduces the time to draw a single frame from ~100 ms to ~15 ms on my laptop, allowing for faster framerates.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -998,38 +998,18 @@ void CPlotter::draw()
                                 &xmin, &xmax);
 
         // draw the pandapter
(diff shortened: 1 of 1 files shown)


csete:
Nice improvement, thanks.

PR #756, opened 2020-02-19
Close 'temp_file' before use.

gvanem:
Building on Windows, I got this message from gr-osmosdr:
`gr::log :WARN: file_source0 - file is too small`

I figured the problem was in `src/applications/gqrx/file_resources.cpp`;
The `temp_file` should be closed before trying to use it. Otherwise it stays at 0 bytes.
Hence the warning (and a thrown exception somewhere).

### DIFF
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -44,7 +44,8 @@ std::string receiver::get_random_file(void)
                 QDataStream stream(&temp_file);
                 for (size_t i = 0; i < 1024*8; i++) stream << qint8(rand());
             }
-            std::cout << "Created random file " << path << std::endl;
+            temp_file.close();
+            std::cout << "Created random file " << path << std::endl;            
         }
     }
     return path;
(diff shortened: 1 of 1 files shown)


src/applications/gqrx/file_resources.cpp
```diff
@@ -44,7 +44,8 @@ std::string receiver::get_random_file(void)
                 QDataStream stream(&temp_file);
                 for (size_t i = 0; i < 1024*8; i++) stream << qint8(rand());
             }
-            std::cout << "Created random file " << path << std::endl;
+            temp_file.close();
+            std::cout << "Created random file " << path << std::endl;
```
argilo: Some white space was accidentally added to this line.
```suggestion
            std::cout << "Created random file " << path << std::endl;
```

gvanem: Removed AFAICS. I *do* have a proper editor.

argilo: Proper editor or not, line 48 still has 12 trailing spaces added to it in this commit. :smile:

gvanem: Oh, I blame the editor at Github. Fixed now

argilo: Looks good now. :+1:

csete:
Thanks for the fix, and thanks for reviewing @argilo :)

PR #758, opened 2020-02-21
usb_claim_interface error because osmosdr devices are not closed

### LINKED ISSUE #757: usb_claim_interface error because osmosdr devices are not closed
The osmosdr devices are closed in the destructor.
Because in gqrx there is a `receiver` singleton with device-`src` instance object (shared_ptr), this destructor is only called on program exit.

So it is not possible to reconfigure the osmosdr input device without restarting gqrx.

A pull request with fix is on the way...

kolbma:
usb_claim_interface error because osmosdr devices are not closed

And stopping crashing when usb device is not connected or configuration error.

I've also added a device entry for rtlsdr with **direct sampling** set to value 2 (some devices need this for HF).

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -293,10 +293,6 @@ MainWindow::MainWindow(const QString cfgfile, bool edit_conf, QWidget *parent) :
     ui->plotter->setTooltipsEnabled(true);
 #endif
 
-    // Create list of input devices. This must be done before the configuration is
-    // restored because device probing might change the device configuration
-    CIoConfig::getDeviceList(devList);
-
     // restore last session
     if (!loadConfig(cfgfile, true, true))
     {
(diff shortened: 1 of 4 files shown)


argilo:
It looks like you've made three separate changes here; I'd recommend putting each change into a separate pull request so that they can be evaluated individually.

The first change, replacing `src` with a vector, seems unnecessary since the root cause of the issue is a bug in GNU Radio, which has already been fixed there (https://github.com/gnuradio/gnuradio/pull/3184).

kolbma:
@argilo 
Well, the development was to see that destructor is not called. Then my idea was that the reset() doesn't free immediately (race) and so I've tried with the vector where I can free the shared_ptr. And somehow I got to see the destructor called after this the first time in debug. 
A side effect is that you could handle multiple interfaces concurrently when you extend this and also provide the UI.

argilo:
> Well, the development was to see that destructor is not called. Then my idea was that the reset() doesn't free immediately (race) and so I've tried with the vector where I can free the shared_ptr. And somehow I got to see the destructor called after this the first time in debug.

I don't think moving `src` into a vector changed anything. It's GNU Radio that holds onto a reference to the source block, not Gqrx. Note that this problem only occurs after the flow graph is started; it could be that you didn't start the flow graph after you moved `src` into a vector.

argilo:
> Not sure what is best to do... at the moment it works for me. I can change devices without restarting or crashing gqrx.

These issues should already be fixed by https://github.com/gnuradio/gnuradio/pull/3184 and https://github.com/csete/gqrx/pull/751.

PR #761, opened 2020-02-22
Clear averaged FFT data when changing FFT size

argilo:
This was originally part of #734.

When changing the FFT size, averaged FFT data is not cleared. This causes incorrect data to be displayed in the panadapter until the old data fades away:

fft-average

To fix this I've updated `setIqFftSize` to clear `d_iirFftData`. The -140.0 comes from here:

https://github.com/csete/gqrx/blob/1f85ca2a3827df793005e92180456abc630c97af/src/applications/gqrx/mainwindow.cpp#L114-L115


argilo:
I added a news item as well.

PR #763, opened 2020-02-24
Use Ubuntu 16.04 for Travis CI

argilo:
Currently Travis CI is performed on Ubuntu 14.04, which is rather outdated. It's also simpler to build with Ubuntu ~18.04~ **16.04**, since it has all the necessary dependencies in its package archive.

The clang builds [currently fail](https://travis-ci.org/argilo/gqrx/builds/654392032) due to the very slow compilation of parser_impl.cc, but that will be fixed by https://github.com/csete/gqrx/pull/762.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: required
 
-dist: trusty
+dist: xenial
 
 language: cpp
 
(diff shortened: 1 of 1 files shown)


argilo:
I'm also thinking about how to add GNU Radio 3.8 to CI. One possibility would be to wait until Travis makes Ubuntu 20.04 available, and then add it to the `dist` list.

csete:
The idea with the Travis builds was to check compatibility wit hthe oldest supported version of Ubuntu (or other OS for that matter). Can 16.04 work?

argilo:
In that case, it makes sense to continue using Ubuntu 16.04 until April 2021. I updated this pull request, and CI still passes: https://travis-ci.org/github/argilo/gqrx/builds/675686420

csete:
Thanks!

I forgot to say regarding adding GNU Radio 3.8 to CI that the GNU Radio project now has a [PPA](https://wiki.gnuradio.org/index.php/InstallingGR#Ubuntu_PPA_Installation), so we can use that.

PR #768, opened 2020-03-03
Combine rx_fft_f and rx_fft_c

vadimostanin:
rx_fft_f and rx_fft_c do same, but by sligtly different code.
Combined to template rx_fft<sizeofsize>
tested, works same


argilo:
Thanks for your contribution @vadimostanin! I agree it would be useful to combine these two classes, but unfortunately the code has changed enough since you made this pull request that the patch cannot be easily applied anymore. I'll close this off for now and consider merging these classes in the future.

PR #770, opened 2020-03-04
Speed up log power calculation using VOLK

argilo:
As noted in #734, the log power calculation in `MainWindow::iqFftTimeout` is very expensive when the FFT size is large. Currently the log power & FFT averaging loop has a running time of about 34ms on my laptop when the FFT size is 1048576, and most of this time is spent in `log10f`.

@ghostop14 helpfully pointed out that VOLK's `volk_32f_log2_32f` is a much faster way to calculate logarithms. I've rewritten the entire log power calculation as VOLK calls, which reduced the runtime from 34ms to 4ms. I left the FFT averaging in the loop because computing that with VOLK turned out to be slower.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -23,6 +23,7 @@
  */
 #include <string>
 #include <vector>
+#include <volk/volk.h>
 
 #include <QSettings>
 #include <QByteArray>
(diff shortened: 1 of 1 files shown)


src/applications/gqrx/mainwindow.cpp
```diff
@@ -1283,23 +1285,14 @@ void MainWindow::iqFftTimeout()
...
-            pt = d_fftData[fftsize/2+i];
-        }
-        else
-        {
-            pt = d_fftData[i-fftsize/2];
-        }
-
-        /* calculate power in dBFS */
-        pwr = pwr_scale * (pt.imag() * pt.imag() + pt.real() * pt.real());
-        d_realFftData[i] = 10.0 * log10f(pwr + 1.0e-20);
```
argilo: I intentionally left out the `+ 1.0e-20` part of the calculation when converting it to VOLK, because the `volk_32f_log2_32f` kernel returns -127 instead of -Inf when the input is zero, as documented here:

https://github.com/gnuradio/volk/blob/aae9fed1cd15f0f1695d5adad04c65cf8bbd3cc0/kernels/volk/volk_32f_log2_32f.h#L30-L31

PR #771, opened 2020-03-04
New feature FrequencyHistory

kolbma:
In https://github.com/csete/gqrx/issues/764 I have announced my development of a FrequencyHistory feature to go back and forward on frequency selections and receiver settings to this frequency.

Entries are added to history after 2 seconds without change.
There is a maximum of 20 entries. Older one is removed.

There are two new buttons in menubar and two entries in tools menu.

Next to frequency the filter settings, mode and squelch setting are restored.

freq-history

### DIFF
--- a/gqrx.pro
+++ b/gqrx.pro
@@ -92,6 +92,8 @@ VERSTR = '\\"$${VER}\\"'          # place quotes around the version string
 DEFINES += VERSION=\"$${VERSTR}\" # create a VERSION macro containing the version string
 
 SOURCES += \
+    src/applications/gqrx/freqhistory.cpp \
+    src/applications/gqrx/freqhistoryentry.cpp \
     src/applications/gqrx/main.cpp \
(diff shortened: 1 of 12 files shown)


kolbma:
YT vid showing the feature https://youtu.be/hjfgLxbfMYE

kolbma:
Ok. Travis fails because I use C++11 lambda.
Really compiling with such an old GCC?

argilo:
I've rebased this to resolve merge conflicts, but I haven't yet tested the changes.

ra1nb0w:
Nice feature. I also find useful a table with the last 20 frequencies that permit to choose not just the previous or the next frequency. From the video this seems not implements. What do you think?

kolbma:
Nice idea.  
Also the timing needs at least an adjustable setting or maybe some auto-detection on the current usage.  
But I can't work at the moment on this...

PR #772, opened 2020-03-04
Fixes preset filter selection

kolbma:
https://github.com/csete/gqrx/issues/766


csete:
Applied manually to aviod merge commit.
Thanks.

PR #773, opened 2020-03-05
Feature to provide menu for recent config files

kolbma:
This stores the last used conf files in a new config file recentconfig.cfg in the default cfg directory.
In file menu there is a new sub-menu **Recent settings** (menu_RecentConfig).
There are added up to 9 actions to load the last known configurations.

gqrx-recentconfig


argilo:
This branch was a mess, so I've squashed it into a single commit and rebased on master. If it still works properly and passes CI, I'll merge this.

PR #774, opened 2020-03-05
Feature Device scan for IOConfig

kolbma:
This adds a new button **Device scan** to IOConfig.
Updates input and output device list.

No need to restart on (un)plugin different devices. Just reconfigure and click OK.

gqrx-update-devlist

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -2,6 +2,7 @@
       2.14: In progress...
 
        NEW: Start/stop DSP via remote control.
+       NEW: Device scan button in Configure I/O devices dialog.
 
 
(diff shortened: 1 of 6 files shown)


kolbma:
The dialog needs longer to open. So at least there is the need for an hourglass pointer.
And if there is any change to hardware stuff (I have only one rtl_sdr to test and there is no problem), which makes gqrx stop to work on a cancel afterwards, it might be good, when the user knows that he has scanned for devices.

argilo:
Thanks for building this! I had a look over the code and tested this out today, and it seems to work well. I've rebased the pull request and will merge once CI passes.

PR #784, opened 2020-04-18
Apply de-emphasis after stereo demodulation

### LINKED ISSUE #783: WFM Stereo separation disappeared?
It appears that the commit 1bf1e1c2f97c635744070eca49db9bf1661062f6 has broken WFM Stereo decoding(the version of Stereo that is transmitted in the 87.5-108 MHZ band, not the OIRT one). The separation between left and right channels has become virtually zero - you might as well just stick to Mono mode. I've verified this by using test signals. In the previous commit - with a commit ID 9718d60c003b31a328b4457b400308c603e1518e, the situation with stereo audio was still normal. It seems that changing to the "new style taps" has broken something. 

As for the version of the GQRX that this is happening at, the situation is still occurring as of the current master branch state(as of 11 UTC on 16 April)

I am aware this isn't really much information, so if you have any further questions, please feel free to ask and I'll provide additional information as needed.

argilo:
Fixes #783.

The fix to FM de-emphasis taps in #733 made another bug apparent: the WFM stereo receiver applies de-emphasis to the entire baseband signal. This causes the L-R signal to be greatly attenuated, removing stereo separation.

To fix this, I've moved de-emphasis to the end of the WFM receive chain so that it is applied do the demodulated left & right channels. After this change, de-emphasis is needed in two places (rx_demod_fm.cpp for Narrow FM, and stereo_demod.cpp for WFM), so I moved it into its own hierarchical block.

### DIFF
--- a/gqrx.pro
+++ b/gqrx.pro
@@ -103,6 +103,7 @@ SOURCES += \
     src/dsp/agc_impl.cpp \
     src/dsp/correct_iq_cc.cpp \
     src/dsp/filter/fir_decim.cpp \
+    src/dsp/fm_deemph.cpp \
     src/dsp/lpf.cpp \
     src/dsp/rds/decoder_impl.cc \
     src/dsp/rds/parser_impl.cc \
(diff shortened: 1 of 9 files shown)


csete:
I don't have any test signal but it seems to work. Thanks.

PR #787, opened 2020-04-19
Set correct sample rate for IQ FFT

### LINKED ISSUE #786: Latency on FFT plot
Commit e6b031257f221ee0ca530eb71f79abe636f805fd seems to have introduced ~1 sec latency to the FFT plot.

* Airspy R2, 2.5 MSPS, decimation=1
* GNU Radio 3.8.0.0
* Volk 2.0.0

argilo:
Fixes #786.

The IQ FFT block is after the input decimator, but before the DDC decimator. Therefore, it operates at `d_decim_rate`, not `d_quad_rate`.

Passing in the wrong rate was causing `rx_fft_c`'s circular buffer to fill up, which introduced latency into the FFT.


csete:
That was fast, thanks!

PR #788, opened 2020-04-23
Tweaks to RDS demodulation for a more stable decoding

es5nhc:
This patch is to mitigate against a situation where even the big local blowtorch FM stations were decoding RDS only intermittently - one could notice how periodically(once in a couple of seconds if not more frequently) the decoding just cut out as if something was multiplying the downconverted RDS symbols with a sine wave. 

The new setup still isn't perfect but is a noticeable improvement over the current situation. It also appears that a number of stations here in Estonia do not have their RDS signal apparently exactly at 57 kHz, but a few hertz off. Some stations are more off than others. If you are interested, I can provide I/Q recordings of the bigger trouble-makers. I sure wonder if it could be more optimal to derive the exact RDS frequency from the stereo pilot tone.

### DIFF
--- a/src/dsp/rx_rds.cpp
+++ b/src/dsp/rx_rds.cpp
@@ -4,7 +4,7 @@
  *           http://gqrx.dk/
  *
  * Copyright 2011 Alexandru Csete OZ9AEC.
- *
+ * 
  * Gqrx is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 3, or (at your option)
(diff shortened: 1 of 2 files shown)


argilo:
I had a closer look at this, and I believe the main reason for Gqrx's poor RDS receive performance is that there is no symbol synchronization. There should be a Symbol Sync block (or equivalent) somewhere in the RDS receiver.

Also, the receiver throws away half of each Manchester-encoded symbol (with a Keep 1 in N block), further degrading receive performance. I'm taking a crack at fixing this problem in gr-rds, and will pull the fixes into Gqrx afterwards.

argilo:
> I sure wonder if it could be more optimal to derive the exact RDS frequency from the stereo pilot tone.

I don't think this is necessary, since the Constellation Receiver block already corrects for frequency offset. Also, some stations don't broadcast in stereo and therefore don't have a pilot tone.

PR #799, opened 2020-05-23
Use std::shared_ptr with GNU Radio 3.9

argilo:
In the upcoming version 3.9, GNU Radio is switching from `boost::shared_ptr` to `std::shared_ptr`:

https://wiki.gnuradio.org/index.php/GNU_Radio_3.9_OOT_Module_Porting_Guide#C.2B.2B_Modernization

As a result, Gqrx will need to do the same for its blocks. I've added preprocessor conditionals to check the GNU Radio version and use the appropriate shared pointer type. I've tested this change on GNU Radio's `maint-3.8` branch as well as `master`.

Note: Similar changes are also needed for the gr-iqbal and gr-osmosdr dependencies. I used these branches when testing Gqrx against GNU Radio's master branch:

https://github.com/argilo/gr-iqbal/tree/gr3.9
https://github.com/argilo/gr-osmosdr/tree/gr3.9

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,6 +1,7 @@
 
     2.13.2: In progress...
 
+       NEW: Preliminary support for GNU Radio 3.9.
      FIXED: Loss of precision in the plotter's frequency axis.
 
 
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -1002,9 +1002,15 @@ receiver::status receiver::start_audio_recording(const std::string filename)
 
     // if this fails, we don't want to go and crash now, do we
     try {
+#if GNURADIO_VERSION < 0x030900
         wav_sink = gr::blocks::wavfile_sink::make(filename.c_str(), 2,
                                                   (unsigned int) d_audio_rate,
                                                   16);
+#else
+        wav_sink = gr::blocks::wavfile_sink::make(filename.c_str(), 2,
+                                                  (unsigned int) d_audio_rate,
+                                                  gr::blocks::FORMAT_WAV, gr::blocks::FORMAT_PCM_16);
+#endif
     }
     catch (std::runtime_error &e) {
         std::cout << "Error opening " << filename << ": " << e.what() << std::endl;
--- a/src/dsp/correct_iq_cc.h
+++ b/src/dsp/correct_iq_cc.h
@@ -38,8 +38,13 @@
 class dc_corr_cc;
 class iq_swap_cc;
 
+#if GNURADIO_VERSION < 0x030900
 typedef boost::shared_ptr<dc_corr_cc> dc_corr_cc_sptr;
 typedef boost::shared_ptr<iq_swap_cc> iq_swap_cc_sptr;
+#else
+typedef std::shared_ptr<dc_corr_cc> dc_corr_cc_sptr;
+typedef std::shared_ptr<iq_swap_cc> iq_swap_cc_sptr;
+#endif
 
 /*! \brief Return a shared_ptr to a new instance of dc_corr_cc.
  *  \param sample_rate The sample rate
--- a/src/dsp/downconverter.h
+++ b/src/dsp/downconverter.h
@@ -33,7 +33,11 @@
 
 class downconverter_cc;
 
+#if GNURADIO_VERSION < 0x030900
 typedef boost::shared_ptr<downconverter_cc> downconverter_cc_sptr;
+#else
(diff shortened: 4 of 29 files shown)


dl1ksv:
Your gr3.9 tree of gr-iqbal and gr-osmosdr both work fine. I'm able to run flowgraphs in grc using gr-osmosdr with an rtl dongle and the latest checkout of gnuradio..

But when I try to run cmake on gqrx cmake fails to find the package gnuradio-osmosdr

-- Checking for module 'gnuradio-osmosdr'
--   No package 'gnuradio-osmosdr' found
-- Could NOT find GNURADIO_OSMOSDR (missing: GNURADIO_OSMOSDR_LIBRARIES GNURADIO_OSMOSDR_INCLUDE_DIRS) 

Maybe this depends on my setup:  

gnuradio in /usr/local/gnuradio/
gr-iqbal and gr-osmosdr in /usr/local/gnuradio/ ,too.

Changing  https://github.com/csete/gqrx/blob/bcf853134958af4fb16d2d722e851c8107fa4f78/CMakeLists.txt#L111
to
find_package(gnuradio-osmosdr REQUIRED PATHS /usr/local/gnuradio/lib/cmake/osmosdr)

makes cmake run without errors.
The PATHS entry is necessary otherwise the package won't be found.
Compiling works, too but linking fails with

/usr/bin/ld: CMakeFiles/gqrx.dir/applications/gqrx/receiver.cpp.o: in function `receiver::get_rf_range(double*, double*, double*)':
receiver.cpp:(.text+0x1c3b): undefined reference to `osmosdr::meta_range_t::meta_range_t()'
/usr/bin/ld: receiver.cpp:(.text+0x1d86): undefined reference to `osmosdr::meta_range_t::start() const'
/usr/bin/ld: receiver.cpp:(.text+0x1d94): undefined reference to `osmosdr::meta_range_t::stop() const'
/usr/bin/ld: receiver.cpp:(.text+0x1f64): undefined reference to `osmosdr::meta_range_t::start() const'
/usr/bin/ld: receiver.cpp:(.text+0x1f71): undefined reference to `osmosdr::meta_range_t::stop() const'
/usr/bin/ld: receiver.cpp:(.text+0x1f82): undefined reference to `osmosdr::meta_range_t::step() const'

and some more.

drmpeg:
Works good here except for linking with Volk. This is with GNU Radio 3.9 and a custom install prefix (/opt/gnuradio-3.9.0git). Here's the error:
```
[ 98%] Building CXX object src/CMakeFiles/gqrx.dir/qrc_textfiles.cpp.o
[100%] Linking CXX executable gqrx
/usr/bin/ld: cannot find -lvolk
/usr/bin/ld: cannot find -lvolk
collect2: error: ld returned 1 exit status
src/CMakeFiles/gqrx.dir/build.make:1072: recipe for target 'src/gqrx' failed
make[2]: *** [src/gqrx] Error 1
CMakeFiles/Makefile2:250: recipe for target 'src/CMakeFiles/gqrx.dir/all' failed
Makefile:132: recipe for target 'all' failed
make: *** [all] Error 2
```
Removing the volk target_link_libraries entries in gqrx/src/CMakeLists.txt fixes the problem.

dl1ksv:
Interesting. I think I have to figure out why find_package(Gnuradio-osmosdr REQUIRED) fails on my system, though osmosdr package is installed and accessible and usable inside grc.

I have no real idea, but my guess is that it depends on the installation location.
So @drmpeg  could you please tell me your installation path of gnuradio and osmosdr.

drmpeg:
@dl1ksv I'm using /opt/gnuradio-3.9.0git

I like to use /opt because it makes it easy to delete installs. /usr/local is too messy.

dl1ksv:
OK, I got it now.
The CMAKE_INSTALL_PREFIX for gqrx and gr-osmosdr must be the same.
I installed gr-osmosdr to /usr/local/gnuradio/  
and run the cmake  to build gqrx without setting the CMAKE_INSTALL_PREFIX  
This resulted in the errors.
Setting the CMAKE_INSTALL_PREFIX to  /usr/local/gnuradio/ gave 

-- Checking for module 'gnuradio-osmosdr'
--   No package 'gnuradio-osmosdr' found
-- Found GNURADIO_OSMOSDR: /usr/local/gnuradio/lib/libgnuradio-osmosdr.so
and gqrx builds and runs fine.

argilo:
The `wavfile_sink` API also changed, so I've accounted for that here. I'll merge for now, and track any additional GNU Radio 3.9 API changes as they occur.

PR #800, opened 2020-05-27
Synchronous AM demodulation

antonblanchard:
Forward port and update the patch from #462. I've hopefully fixed the backward compatibility issues in the process.

Also add a patch to make the carrier PLL loop BW configurable. There are three PLL options (fast, medium and slow). The values were chosen by experimentation with challenging shortwave signals, looking at them in a GRC mock up of the detector as well as listening to them in gqrx. Having said that, they are pretty arbitrary so could be fine tuned based on further feedback.

One thing missing from the demodulator API is a callback whenever the frequency is changed. We could use that the reset the PLL. Right now the PLL can take a long time to adjust to a change in frequency, especially on the slow setting.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -4,6 +4,7 @@
        NEW: Start/stop DSP via remote control.
        NEW: Device scan button in Configure I/O devices dialog.
        NEW: Added "Recent settings" menu.
+       NEW: Added synchronous AM demodulator.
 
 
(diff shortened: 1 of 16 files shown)


argilo:
I've updated this pull request to resolve the merge conflict. The new demodulator appears to work well apart from the minor limitations you mentioned. I'll merge this one CI passes.

PR #803, opened 2020-05-31
Allow scroll wheel direction to be inverted

argilo:
A "natural scrolling" option, whereby the scrolling direction for the mouse wheel & touchpad is inverted, is becoming more common across multiple platforms, and is now often the default. This option works well for text and web pages, but not for some other widgets like the frequency control and plotter in Gqrx.

The QWheelEvent object provides an [`inverted()`](https://doc.qt.io/qt-5/qwheelevent.html#inverted) method, which indicates whether scrolling is inverted, but the documentation warns:

> Many platforms provide no such information. On such platforms inverted always returns false.

This is the case at least on Ubuntu 18.04 and 20.04. As a result, I think it would be best to add a checkbox to Gqrx to allow the user to invert the scrolling direction, which is what I've done here. The new option appears at the bottom of the "Input controls" panel:

gqrx-invert-scrolling

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -3,6 +3,7 @@
 
        NEW: Stereo option for UDP streaming.
        NEW: Script to generate AppImage.
+       NEW: Allow scroll wheel direction to be inverted.
      FIXED: FM de-emphasis causing audio to be 20 dB quieter than it should be.
      FIXED: FM de-emphasis applied incorrectly in WFM stereo receiver.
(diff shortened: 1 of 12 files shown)


argilo:
While working on this change, I noticed that the "Reset frequency controller digits" setting does not take effect in the "Receiver Options" tab until Gqrx is restarted. I added a quick fix for that in a second commit. (Putting it in a separate PR would have caused merge conflicts, since the code is adjacent to the scroll wheel changes.)

PR #811, opened 2020-06-22
qualify placeholders with their full namespace

trabucayre:
With boost >= 1.73, placeholders must be qualified by using the full namespace.

Without the namespace, the build fails with:
gqrx-2.12.1/src/dsp/rds/parser_impl.cc: In constructor 'gr::rds::parser_impl::parser_impl(bool, bool)':
gqrx-2.12.1/src/dsp/rds/parser_impl.cc:42:72: error: '_1' was not declared in this scope
  set_msg_handler(pmt::mp("in"), boost::bind(&parser_impl::parse, this, _1));

### DIFF
--- a/src/dsp/rds/parser_impl.cc
+++ b/src/dsp/rds/parser_impl.cc
@@ -40,7 +40,7 @@ parser_impl::parser_impl(bool log, bool debug, unsigned char pty_locale)
 	pty_locale(pty_locale)
 {
 	message_port_register_in(pmt::mp("in"));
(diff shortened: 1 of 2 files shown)


argilo:
Unfortunately, this change broke the Travis CI build because `boost::placeholders` does not exist in Boost <1.60, and Ubuntu 16.04 has Boost 1.58: https://travis-ci.org/github/csete/gqrx/builds/733464745

I think the best path forward would be for gr-rds and Gqrx to switch to `std::bind` and `std::placeholders`.

PR #816, opened 2020-07-21
airspyhf: Add all supported sample rates

kbeckmann:


kbeckmann:
I have tested this with an Airspy HF+ Discovery. The output from SoapySDRUtil --probe shows the following:

```
  Sample rates: 0.768, 0.384, 0.256, 0.192 MSps
```

I have also tested all of the sample rates and they do work.

I do wonder how this works with a classic Airspy HF+ (the one with two ports). The code that populates the supported sample rates requests this from the device, so there is a chance we get other values from the HF+. The code that does that is located here for reference https://github.com/airspy/airspyhf/blob/master/libairspyhf/src/airspyhf.c#L872

It would be helpful if someone who has an Airspy HF+ could post the output from `SoapySDRUtil --probe`, then we know for sure if this patch will work with both devices or not.

csete:
I have tried SoapySDRUtil with my two-port HF+ and it only reports 0.768 MSPS. Perhaps I need to update the firmware to something recent?

kbeckmann:
> Perhaps I need to update the firmware to something recent?

Maybe. Here's my output from `airspyhf_info`:

```
AirSpy HF library version: 1.6.8

S/N: (removed)
Part ID: 0x00000002
Firmware Version: R3.0.6-CD
Available sample rates: 768 kS/s 384 kS/s 256 kS/s 192 kS/s
```

kbeckmann:
Maybe the proper way would be to probe for all sample rates and only add those that are actually supported dynamically.

csete:
I agree that would be the best assuming that the backend implements it properly, and it would be quite tricky to implement as it would require creating and destroying objects multiple times. Moreover, some recent transceiver devices (LimeSDR, USRP, BladeRF) support arbitrary sample rates so that case would have to be handled as well.

I will try to update the firmware on my device.

argilo:
My Airspy HF+ Dual Port initially had firmware 1.0.0 which only supported 768 kS/s, but once I updated to the latest firmware more rates were available:
```
$ airspyhf_info
AirSpy HF library version: 1.6.8

S/N: 0xC852948094D23F89
Part ID: 0x00000001
Firmware Version: R3.0.7-BB
Available sample rates: 912 kS/s 768 kS/s 456 kS/s 384 kS/s 256 kS/s 192 kS/s
```
All of the rates you've added here are supported and the change works well, so I'll merge this.

I did notice that if an invalid sample rate is selected, then the following error occurs and Gqrx becomes unresponsive:
```
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.
```
That seems to be a pre-existing bug though, so I'll look into it separately.

PR #825, opened 2020-10-01
Ensure that a minimum filter width is maintained

argilo:
When dragging the "Low cut" and "High cut" frequencies in the plotter, it is easy to accidentally drag the high cut frequency below the low cut frequency (or vice versa), which causes the receive flowgraph to crash:
```
sched: <block fir_filter_blk<IN_T,OUT_T,TAP_T> (14)> is requesting more input data
  than we can provide.
  ninput_items_required = 11570
  max_possible_items_available = 8191
  If this is a filter, consider reducing the number of taps.
```
To recover from this situation, it is necessary to correct the filter frequencies, then stop & restart the DSP.

To avoid this problem I've added bounds checks to ensure that the two frequencies always remain at least 200 Hz apart. This matches the minimum bandwidth of the CW demodulator.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -75,6 +75,8 @@ int gettimeofday(struct timeval * tp, struct timezone * tzp)
 #define FFT_MIN_DB     -160.f
 #define FFT_MAX_DB      0.f
 
+#define FILTER_WIDTH_MIN_HZ 200
+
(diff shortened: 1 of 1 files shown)


argilo:
@Flamewires This fixes the bug you encountered during GRCon.

Flamewires:
Awesome, I completely forget about this, thanks for remembering and fixing it!

csete:
Thanks!. Indeed, this has often been a problem when using narrow-band modes.

PR #827, opened 2020-10-14
Switch from boost::bind to std::bind

argilo:
#811 broke the Travis CI build (https://travis-ci.org/github/csete/gqrx/builds/733464745) because `boost::placeholders` does not exist in Boost <1.60. Switching to `std::bind` and `std::placeholders` resolves the problem (https://travis-ci.org/github/argilo/gqrx/builds/735583345), and I've opened an upstream pull request for this: https://github.com/bastibl/gr-rds/pull/35.

RDS still works correctly in Gqrx after this change.


argilo:
Merging to get CI working again.

PR #830, opened 2020-10-15
Fix linking of Volk with a custom install prefix.

drmpeg:


drmpeg:
Tested on both GNU Radio 3.7 and 3.8.

argilo:
I'm not very familiar with cmake, so I'm having trouble understanding what the purpose of this change is and how it works. Could you provide a more detailed description? Thanks!

drmpeg:
Sure, no problem. It has to do with using a custom install prefix. Let's say I use the following for the CMake stage of the build because that's where I've installed GNU Radio:
```
cmake -DCMAKE_INSTALL_PREFIX=/opt/gnuradio-3.7.14 ../
```
When the linker executes with the current CMake code in src/CMakeLists.txt, it's going to use the syntax:
```
-lvolk
``
to find the Volk library. But there's no path on the system for the Volk library and it's going to fail with:
```
/usr/bin/ld: cannot find -lvolk
collect2: error: ld returned 1 exit status
src/CMakeFiles/gqrx.dir/build.make:1685: recipe for target 'src/gqrx' failed
```
Now you could add a path with the environment variable LIBRARY_PATH (not LD_LIBRARY_PATH). But you can get CMake to do it all automatically. BTW, the current CMake files work for the default install directory /usr/local since the linker always searches /usr/local/lib.

For GNU Radio 3.7, I've added the lines:
```
if(Gnuradio_VERSION VERSION_LESS "3.8")
    find_package(Volk)
endif()
```
to the top level CMakelists.txt. This will set VOLK_LIBRARIES to the correct path (if Volk is found). I've also changed the target_link_libraries() in src/CMakeLists.txt to use ```${VOLK_LIBRARIES}``` instead of ```volk```. Then when the linker runs, CMake will add ```/opt/gnuradio-3.7.14/lib/libvolk.so``` to the list of libraries to link with (just like it did for all the other GNU Radio libraries).

On the GNU Radio 3.8, the syntax is different (```Volk::volk```), but it accomplishes the same thing. There's no need for a ```find_package()``` because Volk always gets included with gnuradio-runtime on 3.8.

argilo:
Thanks for the detailed explanation @drmpeg! This makes sense, and CI passed. I'll do a quick test to confirm that this still plays nicely with PyBOMBS and then merge.

argilo:
Tested against GNU Radio 3.7, 3.8, and 3.9 with PyBOMBS and everything was good.

PR #836, opened 2020-10-17
Add more Airspy HF+ sample rates

argilo:
#816 added more Airspy HF+ sample rates, but recent firmware versions have a couple more: 456 ksps and 912 ksps. I've added them here, and left 768 ksps as the default since it should be supported in all firmware versions.

@kbeckmann


kbeckmann:
Nice! I added the sample rates that were available to me at the time. Also I have the single port version, could be that these sample rates are only available on the dual port version. I could update the firmware and check with my device later today.

argilo:
I'll merge for now since it works on my end, but do let me know if you run into any problems with the single-port version.

kbeckmann:
With the latest firmware R3.0.7 I also get 912 kS/s and 456 kS/s. Nice!

PR #837, opened 2020-10-18
Fix compiler warnings

argilo:
While building on Debian unstable, I noticed some compiler warnings, which I've fixed here.

### DIFF
--- a/src/dsp/rx_noise_blanker_cc.cpp
+++ b/src/dsp/rx_noise_blanker_cc.cpp
@@ -47,11 +47,12 @@ rx_nb_cc::rx_nb_cc(double sample_rate, float thld1, float thld2)
       d_thld_nb2(thld2),
       d_avgmag_nb1(1.0),
       d_avgmag_nb2(1.0),
+      d_delay {0},
       d_delidx(2),
       d_sigidx(0),
       d_hangtime(0)
 {
-    memset(d_delay, 0, 8 * sizeof(gr_complex));
+
 }
 
 rx_nb_cc::~rx_nb_cc()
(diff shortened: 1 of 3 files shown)


src/dsp/rx_noise_blanker_cc.cpp
```diff
@@ -47,11 +47,12 @@ rx_nb_cc::rx_nb_cc(double sample_rate, float thld1, float thld2)
       d_thld_nb2(thld2),
       d_avgmag_nb1(1.0),
       d_avgmag_nb2(1.0),
+      d_delay {0},
       d_delidx(2),
       d_sigidx(0),
       d_hangtime(0)
 {
-    memset(d_delay, 0, 8 * sizeof(gr_complex));
```
argilo, line 54: The warning here is "cleaning an object of non-trivial type; use assignment or value-initialization instead". I moved this to the initialization list, since arrays can be initialized there in C++11.

src/qtgui/bookmarkstaglist.cpp
```diff
@@ -269,7 +269,7 @@ void BookmarksTagList::AddTag(QString name, Qt::CheckState checkstate, QColor co
     // Column 0
     item = new QTableWidgetItem();
     item->setFlags(Qt::ItemIsEnabled);
-    item->setBackgroundColor(color);
```
argilo, line 272: https://doc.qt.io/qt-5/qtablewidgetitem-obsolete.html#backgroundColor

src/qtgui/plotter.cpp
```diff
@@ -1270,7 +1270,6 @@ void CPlotter::drawOverlay()
     QFontMetrics    metrics(m_Font);
     QPainter        painter(&m_OverlayPixmap);
 
-    painter.initFrom(this);
```
argilo, line 1273: https://doc.qt.io/qt-5/qpainter-obsolete.html#initFrom

This is obsolete and no replacement is suggested. The plotter appears to work fine without it, so I just removed it.

PR #838, opened 2020-10-18
Fix loss of precision in xFromFreq and freqFromX

### LINKED ISSUE #818: Minimal tuning step achievable with mouse clicking is too coarse
Hi,
Using Gqrx v2.12.1-52-gbcf8531349 with Adalm Pluto for monitoring beacons on 5760 MHz. 
Don't know what causes it but the minimum tuning step is 510 Hz using mouse on waterfall or panadapter. One can click anywhere on signal, but vfo "sticks" to nearest 510 Hz points.

argilo:
Fixes #818.

At high frequencies (e.g. 5 GHz) and when zoomed in on the frequency axis, there is a noticeable loss of precision when clicking to tune. Also, the red bar showing the currently tuned frequency and the grey area showing the filter edges exhibit a similar loss of precision.

These problems occur because `CPlotter::xFromFreq` and `CPlotter::freqFromX` use single-precision floating-point, which only has 24 bits of precision. I've switched these calculations to use `qint64` instead.

@ly2ss If you have a chance to test this, it would be much appreciated!

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.13.2: In progress...
+
+     FIXED: Loss of precision in the plotter's frequency axis.
+
+
     2.13.1: Released October 17, 2020
 
      FIXED: Crash when invalid sample rate is specified.
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1521,12 +1521,12 @@ void CPlotter::makeFrequencyStrs()
     }
 }
 
-// Convert from screen coordinate to frequency
+// Convert from frequency to screen coordinate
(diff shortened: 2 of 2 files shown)


ly2ss:
All controls (filter edges too) behavior work correctly now. 
100 Hz tuning/clicking step is consistent across whole band.

Maybe smaller step 10 Hz (configurable?) would be nice (especially narrow modes). But it's out of subject scope.

(Personally I am used to tune gqrx within band clicking on stations and then clicking and holding left mouse I drag filter "window" to exact tone. This way I love to surf bands - no need to fiddle with frequency control)

I am happy with the way it works now. Thanks!

argilo:
Wonderful, thanks for testing! I'll merge this, and get it into a release later this week.

PR #851, opened 2020-11-02
Merge in the latest changes from gr-rds

argilo:
In https://github.com/bastibl/gr-rds/pull/37, gr-rds removed its dependence on Boost.Format. Here I'm merging the changes into gqrx as well, removing its second-last dependence on Boost.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -2,12 +2,13 @@
     2.13.4: In progress...
 
        NEW: Added man page for use by distributions.
+  IMPROVED: Removed Boost.Format dependency.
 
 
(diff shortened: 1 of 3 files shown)


kukabu:
Hello.
Maybe remove gr-rds from gqrx and add it as an external dependency?

argilo:
Back when RDS was added, it was decided to copy in the parts of gr-rds that were needed rather than taking on another external dependency:

https://github.com/csete/gqrx/pull/226#issuecomment-77867832

This has been working reasonably well, so I'm OK leaving it that way or the time being.

PR #652, opened 2018-08-11
Adding Support for DX Cluster in GQRX

dh2wq:
Hi,
I would like to add DXCluster Support to my main use gqrx as panadapter.
I mainly reused the Bookmark feature and tried to implement no collision with the bookmark feature.
Fir Reference this is an screenshot of ongoing WAE Contest this weekend.

gqrc dxcluster wae
Following image is showing my settings to connect it to my local Skimmer Server instance. 
gqrx dxc options dh2wq skimmer
Usually I'm based on programming HAL in small uControllers, so please have a close look to my implementation.
It was just tested on up to date Linux Mint and Ubuntu.

Best regards and thanks for this great piece of software. I really enjoy to use it.

PR #864, opened 2020-11-14
Add support for DX cluster

argilo:
I recently rebased @dh2wq's pull request #652 and attempted to push to `dh2wq:master`, but this pushed `csete:master` instead and closed off the pull request. I don't see any way to recover from that, so I'm opening a new pull request to continue work on this feature.

I have not yet tested the rebased code, but it at least compiles without error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ The following people and organisations have contributed to gqrx:
 * Nate Temple
 * Nick Robinson, KE5YWP
 * Nokia
+* Oliver Grossmann, DH2WQ
 * Pavel Milanes, CO7WT
 * Pavel Stano
 * Phil Vachon
--- a/gqrx.pro
+++ b/gqrx.pro
@@ -145,7 +145,9 @@ SOURCES += \
     src/qtgui/qtcolorpicker.cpp \
     src/receivers/nbrx.cpp \
     src/receivers/receiver_base.cpp \
-    src/receivers/wfmrx.cpp
+    src/receivers/wfmrx.cpp \
+    src/qtgui/dxc_spots.cpp \
+    src/qtgui/dxc_options.cpp
 
 HEADERS += \
     src/applications/gqrx/gqrx.h \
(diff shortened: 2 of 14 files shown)


argilo:
@ra1nb0w I've incorporated the additional fixes from your experimental branch.

I'll do some code review and testing, then merge this if things are looking good.

ra1nb0w:
Good. thank you @argilo for your effort

argilo:
One bug I've noticed: if the DSP is stopped, then the FFT display goes blank shortly afterwards.

argilo:
That seems to be a side-effect of `CPlotter::updateOverlay()` being called periodically. Even in earlier versions of Gqrx, the FFT goes blank if the window is resized while the DSP is stopped. I guess the FFT state needs to be kept when the DSP is stopped.

argilo:
Another bug: if gqrx exits abnormally (for instance, when Ctrl+C is pressed in the terminal) while the DX cluster is connected, then its icon remains in the dock permanently.

dh2wq:
Sorry for being late by replying but I do have very busy times in the moment.
I will also support by testing later this week.

Thanks for continuing the work. I do have only basic programming knowledge and might do stupid stuff...

argilo:
Part of the reason that the FFT display was going blank is that there was a 1-second interval timer to poll for expired DX spots. I've replaced that with a single-shot timer that fires at the exact time that a DX spot expires. That way, the FFT overlay will only be redrawn when there is actually a change in the DX spots.

I've been unable to reproduce the bug with abnormal closure, so it could have been an artifact of changes I was working on.

I noticed one more bug: if multiple DX spots arrive in a single TCP packet, then only the first one is displayed. I guess the processing should happen one line at a time instead of one packet at a time.

argilo:
I've updated the code to read spots one line at a time.

The next bug: gqrx crashes if the line is not in the expected format:

`ASSERT failure in QList<T>::operator[]: "index out of range", file /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h, line 552`

I'll add input validation to fix this.

argilo:
The crash is now fixed. In the meantime I got a couple more copies of the Gqrx icon stuck in my dock, so I'll see if I can track down why that's happening.

argilo:
I'll merge this now, since the feature appears to be working well enough. If there are bugs, we can sort those out as they're found.

PR #865, opened 2020-11-14
Fix PortAudio with GNU Radio 3.8 on macOS/Darwin platform

ra1nb0w:
this is needed if the file system is case-sensitive


argilo:
Thanks!

Issue #829, opened 2020-10-14
Building gqrx with qmake fails, with locally built libgnuradio-osmosdr.so


darrenfreeman:
Compiling from the latest git.

It appears that the qmake project requires the gnuradio-osmosdr package, that I don't have installed.

```
qmake ../
Info: creating stash file /opt/gqrx/build/.qmake.stash
Project MESSAGE: No prefix given. Using /usr/local
Project MESSAGE: Gqrx configured with pulseaudio backend.
Project ERROR: gnuradio-osmosdr development package not found
```

However it configures, and then builds fine, if I use cmake instead.

```
cmake ../
[many lines deleted]
-- Configuring done
-- Generating done
-- Build files have been written to: /opt/gqrx/build
```

I don't think this package is really needed if I don't have the relevant hardware. As evidenced by success with cmake.

csete:
gnuradio-osmosdr is a required dependency: https://github.com/csete/gqrx/blob/master/CMakeLists.txt#L111
Gqrx can neither build nor run without it.

A more likely explanation is that you actually do have gnuradio-osmosdr installed but qmake can not find it.

darrenfreeman:
I created a new build directory, and started again. Here is the relevant output from cmake. Indeed you are right, cmake is able to find my locally built and installed version. (I am using gnuradio etc. all built from source.)

```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Checking for module 'gnuradio-osmosdr'
--   No package 'gnuradio-osmosdr' found
-- Found GNURADIO_OSMOSDR: /usr/local/lib/libgnuradio-osmosdr.so  
```

So is the issue then with libgnuradio-osmosdr not installing a .pc file when it's built from source? Or should qmake be able to find it regardless?

argilo:
I attempted to add a a qmake build on MacOS to Travis CI and encountered the same error: https://travis-ci.org/github/argilo/gqrx/jobs/744354506. The qmake build works fine on Linux.

I'll take a look at this soon.

Perhaps this is related? http://osmocom.org/issues/4423

argilo:
Yes indeed, the removal of gnuradio-osmosdr.pc in gr-osmosdr 0.2.0 seems to be the problem.

I submitted a patch to gr-osmosdr to bring it back: http://lists.osmocom.org/pipermail/osmocom-sdr/2020-November/002109.html

argilo:
[...]

PR #875, opened 2020-11-19
Drop support for qmake

argilo:
Gqrx currently uses two build systems: qmake and CMake. This means that many changes to gqrx require two build files to be updated and tested. The GNU Radio ecosystem has standardized on CMake, so it's becoming difficult to fix problems like the one described in #829. Since Qt Creator has built-in support for CMake, I think we should now abandon qmake.

Please let me know if there are compelling reasons to maintain the qmake build.


argilo:
@csete I'd be interested in your feedback on this. It seems from the description in #290 that keeping qmake was only meant to be temporary, but I see you've been keeping it up to date in the years since. Do you need it for anything?

csete:
The only reason I keept the .pro file was to be able to load the project into Qt Creator for convenient editing. I haven't used it for building. If there is another way to load the project into Qt Creator then I think the qmake-based build can be removed.

kgarrels:
yes, please leave the .pro file in for a while.

I will try to open & build with cmake in QtCreator, and report back once successful.

Thanks,
kai

argilo:
I was able to load the project in Qt Creator using the `CMakeLists.txt` file, and the project built and ran correctly within the IDE. So it would appear there's no reason to keep the `gqrx.pro` file. I'll merge this now.

kgarrels:
it seems the .pro/qmake build also created the macos app bundle, let's see how we can recreate this with cmake.
Any suggestions?

argilo:
What MacOS app bundle are you referring to? Macports? Homebrew? Something else?

kgarrels:
The app bundle it is the "gqrx.app" directory, that includes the excutable, plus the app icon and a plist file.
It allows to launch the app from the finder etc.

It looks like this:
`Gqrx.app/Contents:
total 16
-rw-r--r--  1 kai  staff  714 20 Nov 09:07 Info.plist
drwxr-xr-x  3 kai  staff   96 20 Nov 09:07 MacOS
-rw-r--r--  1 kai  staff    9 20 Nov 09:07 PkgInfo
drwxr-xr-x  4 kai  staff  128 20 Nov 09:07 Resources

Gqrx.app/Contents/MacOS:
total 3280
-rwxr-xr-x  1 kai  staff  1678216 20 Nov 09:07 Gqrx

Gqrx.app/Contents/Resources:
total 1168
-rw-r--r--  1 kai  staff       0 20 Nov 09:07 empty.lproj
-rw-r--r--  1 kai  staff  596907 20 Nov 09:07 gqrx.icns
`

argilo:
Where did you get that from?

kgarrels:
https://doc.qt.io/qt-5/macos-deployment.html

argilo:
I still don't understand this though:

> it seems the .pro/qmake build also created the macos app bundle

Gqrx has not made MacOS releases since version 2.11. I don't think removing gqrx.pro has changed anything with respect to MacOS.

kgarrels:
I have been building locally with the .pro file on macos.

argilo:
OK, thanks for the explanation. I would assume it's possible to do this with CMake as well. If changes are needed in CMakeLists.txt, please let me know.

kgarrels:
yes, I already started investigating, cmake is a beast and I am not as fast as you.

Thanks again for driving all of this!

csete:
Although qmake can create app bundles for Mac, it has never been able to produce the gqrx.app bundle because of the complex dependency tree. I always used a script that took each library, added it to the app, and updated the paths. It has been an excruciatingly painful process, which is why I gave up many versions ago.

argilo:
I'd like to set up GitHub Actions (or similar) to produce gqrx binaries automatically. For AppImage this will probably be straightforward, but I'm not a Mac user and so I'll need help to produce an app bundle. @csete If you still have your script around, it might at least be useful as a starting point.

kgarrels:
> Although qmake can create app bundles for Mac, it has never been able to produce the gqrx.app bundle because of the complex dependency tree. I always used a script that took each library, added it to the app, and updated the paths. It has been an excruciatingly painful process, which is why I gave up many versions ago.

yes, I make a "naked" app gundle only for the binary, the dependencies are all install with macports.

kgarrels:
this shoud do the trick

`
# Build the program

if(APPLE)
    set(ICON_FILE ${CMAKE_SOURCE_DIR}/resources/icons/gqrx.icns)
    set_source_files_properties(${ICON_FILE} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")

    add_executable(${PROJECT_NAME} MACOSX_BUNDLE ${${PROJECT_NAME}_SOURCE} ${UIS_HDRS} ${RESOURCES_LIST} ${ICON_FILE})
    set(MACOSX_BUNDLE_BUNDLE_NAME gqrx)
    set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.csete.gqrx.Gqrx")
    set(MACOSX_BUNDLE_INFO_STRING "gqrx")
    set(MACOSX_BUNDLE_ICON_FILE "gqrx.icns")
    set(MACOSX_BUNDLE_LONG_VERSION_STRING ${VERSION})
    set(MACOSX_BUNDLE_SHORT_VERSION_STRING ${VERSION})
    set(MACOSX_BUNDLE_BUNDLE_VERSION ${VERSION})    
    #set_property(TARGET ${PROJECT_NAME} PROPERTY MACOSX_BUNDLE_INFO_PLIST ${CMAKE_SOURCE_DIR}/resources/Info.plist)    # message(${MACOSX_BUNDLE_INFO_PLIST})
    add_custom_command(
        TARGET ${PROJECT_NAME}
        POST_BUILD
        COMMAND plutil -replace NSHighResolutionCapable -bool true gqrx.app/Contents/Info.plist
        )
else()
endif(APPLE)

set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 11)
# The pulse libraries are only needed on Linux. On other platforms they will not be found, so having them here is fine.
target_link_libraries(${PROJECT_NAME}
@ -73,6 +95,7 @@ target_link_libraries(${PROJECT_NAME}
    ${VOLK_LIBRARIES}
)


if(NOT Gnuradio_VERSION VERSION_LESS "3.8")
    target_link_libraries(${PROJECT_NAME}
        gnuradio::gnuradio-analog
@ -91,4 +114,5 @@ if (WIN32)
endif (WIN32)

set(INSTALL_DEFAULT_BINDIR "bin" CACHE STRING "Appended to CMAKE_INSTALL_PREFIX")
install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION ${INSTALL_DEFAULT_BINDIR})
install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION ${INSTALL_DEFAULT_BINDIR} BUNDLE DESTINATION ${INSTALL_DEFAULT_BINDIR})
`

csete:
>  @csete If you still have your script around, it might at least be useful as a starting point.

Sorry for the late reply. I have sent my notes and scripts over email.

PR #895, opened 2020-12-02
Add button to lock audio fft and wf ranges

### LINKED ISSUE #892: Option to sync pandapter and waterfall range for audio tab
In the FFT the "Lock" button allows the IQ pandapter and waterfall to always have the same range. The audio pandapter and waterfall do not have this option and it would be nice to add it.

shuyuan-liu:
Implements #892. Added Lock buttin in audio options dialog to lock audio fft and waterfall ranges.

### DIFF
--- a/src/qtgui/audio_options.cpp
+++ b/src/qtgui/audio_options.cpp
@@ -24,8 +24,6 @@
 #include <QPalette>
 #include <QDebug>
 
-#include <iostream>
-
 #include "audio_options.h"
 #include "ui_audio_options.h"
 
--- a/src/qtgui/audio_options.cpp
+++ b/src/qtgui/audio_options.cpp
@@ -128,6 +128,8 @@ void CAudioOptions::on_pandRangeSlider_valuesChanged(int min, int max)
 {
     if (ui->audioLockButton->isChecked())
         ui->wfRangeSlider->setValues(min, max);
+
+    m_pand_last_modified = true;
     emit newPandapterRange(min, max);
 }
 
@@ -147,7 +149,28 @@ void CAudioOptions::on_wfRangeSlider_valuesChanged(int min, int max)
 {
     if (ui->audioLockButton->isChecked())
         ui->pandRangeSlider->setValues(min, max);
+
+    m_pand_last_modified = false;
     emit newWaterfallRange(min, max);
+}
+
+/** Lock button toggled */
+void CAudioOptions::on_audioLockButton_toggled(bool checked)
+{
+    if (checked) {
+        if (m_pand_last_modified)
+        {
+            int min = ui->pandRangeSlider->minimumValue();
+            int max = ui->pandRangeSlider->maximumValue();
+            ui->wfRangeSlider->setPositions(min, max);
+        }
+        else
+        {
+            int min = ui->wfRangeSlider->minimumValue();
+            int max = ui->wfRangeSlider->maximumValue();
+            ui->pandRangeSlider->setPositions(min, max);
+        }
+    }
 }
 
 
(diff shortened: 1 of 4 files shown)


argilo:
This seems like a useful feature. One thing I noticed is that when the ranges are locked, dragging the frequency axis or zooming in on the frequency axis using the mouse wheel does not have any effect on the waterfall. Would you like to try fixing that?

I also noticed that adjusting the frequency axis does not update the slider values, but this problem predates your pull request.

shuyuan-liu:
@argilo I can't replicate the issue. Dragging and zooming are working for me:
image

shuyuan-liu:
I'll try making the sliders update according to the axis, in another PR

src/qtgui/audio_options.cpp
```diff
@@ -147,9 +149,30 @@ void CAudioOptions::on_wfRangeSlider_valuesChanged(int min, int max)
...
+        }
+        else
+        {
+            int min = ui->wfRangeSlider->minimumValue();
+            int max = ui->wfRangeSlider->maximumValue();
+            ui->pandRangeSlider->setPositions(min, max);
+        }
+    }
+}
+
```
shuyuan-liu, line 175: Note: implementation copied from dockfft.cpp

argilo:
> I can't replicate the issue. Dragging and zooming are working for me

Oops, I mistyped. I meant dragging the amplitude axis, not the frequency axis.

argilo:
I see you've already fixed dragging the frequency axis, thanks! I'll have a look at this tonight.

shuyuan-liu:
Now when Lock is enabled, dragging / zooming the pandapter's vertical axis will also change tha range of the waterfall, and (as a side effect really) will aos change the slider values in the options dialog.

shuyuan-liu:
I decided I'd better block signals for the pandapter slider and handle the updating myself, otherwise updating the slider's values will emit a signal that updates the pandapter, which seems to create a signal loop.

argilo:
Looking good! One last thing: the state of the button is not saved. I think a `db_ranges_locked` setting should be added here:

https://github.com/csete/gqrx/blob/a98411f77503f2060ebf2babbaaf875088fdf7d1/src/qtgui/dockaudio.cpp#L315-L425

Here's the corresponding code for the FFT:

https://github.com/csete/gqrx/blob/a98411f77503f2060ebf2babbaaf875088fdf7d1/src/qtgui/dockfft.cpp#L265-L269

https://github.com/csete/gqrx/blob/a98411f77503f2060ebf2babbaaf875088fdf7d1/src/qtgui/dockfft.cpp#L346-L347

shuyuan-liu:
Added get / set functions to help save / load the lock button state. This is because the button is a private member of `CAudioOptions`, which means the `setChecked()` function of the button cannot be used directly from within `dockaudio.cpp`.

argilo:
Thanks for your hard work on this! The code looks good and the feature is working well, so I'll merge this now.

ulrichloose:
A small Question...

This new button should work as a toggle button? The name suggests that. If so, then there must be an error... The compilation of the latest gqrx produces on my mac a constantly "blue" Button without any feedback of the toggle state.

shuyuan-liu:
@ulrichloose 

Yes the new button is a toggle button and should work like the Lock button in the FFT settings. Is the FFT settings Lock button working for you?

Putting aside the fact that the new button is always blue on your build, does it actually work? I.e. when you click it once then drag the vertical axis of the audio FFT, does the audio waterfall change accordingly?

I tried building on Linux with both clang and gcc, and the new button works just like the old Lock button in FFT settings.

argilo:
I was able to reproduce this problem as well. On Linux the "Lock" button works properly, but on macOS it does not. (The "Lock" button in the FFT Settings panel works fine on both operating systems.)

argilo:
Fixed in 3c945faa83eb57fc3d6567f43001e8ed53695d69.

PR #901, opened 2020-12-05
FreeBSD fix

martymac:
This small patch enables audio backend selection (and fixes build) on FreeBSD.

I've also added messages to indicate the selected backend in build logs.


argilo:
Thanks!

PR #907, opened 2020-12-16
How about some more keyboard shortcuts?

szpajder:
This PR adds keyboard shortcuts for some receiver options and audio-related functions. I find them convenient especially on HF, where many modes and filter widths are in use.

Modes:

- `!` - Demod off
- `i` - Raw I/Q
- `a` - AM
- `n` - NFM
- `w` - WFM mono
- `shift+w` - WFM stereo
- `s` - LSB (memo: lower SSB)
- `shift+s` - USB (memo: upper SSB)
- `c` - CW-L
- `shift+c` - CW-U
- `o` - WFM oirt
- `shift+a` - AM sync

Squelch:

- `` ` `` - reset squelch
- `~` - auto squelch

Filter width:

- `<` - narrow
- `.` - normal
- `>` - wide

Audio functions:

- `r` - start/stop recording
- `m` - toggle mute
- `-` - decrease gain by 1 dB
- `+` - increase gain by 1 dB

### DIFF
--- a/resources/kbd-shortcuts.txt
+++ b/resources/kbd-shortcuts.txt
@@ -2,31 +2,31 @@ Keyboard shortcuts
 
 Main window actions:
 
- ctrl+d         Start/stop DSP
- ctrl+l         Load settings from a file
- ctrl+s         Save current settings to a file
- ctrl+shift+b   Bookmark the current frequency and mode
- ctrl+i         Open I/Q recorder dialog
- ctrl+w         Save waterfall as an image
- ctrl+c         Open DX cluster dialog
+ Ctrl+D         Start/stop DSP
+ Ctrl+L         Load settings from a file
+ Ctrl+S         Save current settings to a file
+ Ctrl+Shift+B   Bookmark the current frequency and mode
+ Ctrl+I         Open I/Q recorder dialog
+ Ctrl+W         Save waterfall as an image
+ Ctrl+C         Open DX cluster dialog
  F11            Toggle full screen mode
- f              Set focus to the frequency controller
- ctrl+q         Quit the program
+ F              Set focus to the frequency controller
+ Ctrl+Q         Quit the program
 
 Receiver modes:
 
  !              Demod off
- i              Raw I/Q
- a              AM
- n              NFM
- w              WFM mono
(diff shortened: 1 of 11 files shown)


csete:
Is it not somehow possible to handle multiple shortcuts in one callback?
I would also strongly recommend starting thinking about documenting the shortcuts somewhere. A simple text file accessible via the Help menu would be a good start, see e.g. https://github.com/csete/gqrx/blob/master/resources/remote-control.txt

szpajder:
I have added the help text for all shortcuts.

One could probably connect more than one shortcut to the same slot (callback), but how would that callback look like for dozens of shortcuts? A giant switch or if statement checking for all possible shortcuts and firing shortcut-specific methods? IMO that's more than ugly.

resources/kbd-shortcuts.txt
```diff
@@ -0,0 +1,47 @@
+Keyboard shortcuts
+
+Main window actions:
+
+ ctrl+d         Start/stop DSP
+ ctrl+d         Load settings from a file
```
argilo, line 6:
```suggestion
 ctrl+l         Load settings from a file
```

argilo:
I'm not sure why, but the `<` shortcut (narrow filter width) is not working for me. (Tested on Ubuntu 20.04.)

szpajder:
@argilo it should be better now, can you please try again?

For some reason I couldn't get shortcuts with `Qt::Key_<special_character>` codes to work correctly, so I thought this has to be done as a key sequence with Shift... which is probably not the best idea given that people may use various keyboard layouts. Anyway, now it works as I wanted initially, so it must have been a PEBKAC of some kind.

csete:
> One could probably connect more than one shortcut to the same slot (callback), but how would that callback look like for dozens of shortcuts? A giant switch or if statement checking for all possible shortcuts and firing shortcut-specific methods? IMO that's more than ugly.

Not necessarily one callback for handling all shortcuts, but one callback for handling mode changes, one for handling filter changes, and so on. Having dozens of callbacks with a single line of code is not pretty either, and certainly not cheap in terms of instructions.

PR #910, opened 2020-12-20
Window namespace change in GR 3.9

willcode:
Renames to move from filter::firdes to fft::window for GR 3.9.

No thought has been given to how to make this compile for both 3.8 and 3.9, so the PR is not usable yet.

Signed-off-by: Jeff Long <willcode4@gmail.com>


willcode:
The new namespace already existed. The old one was deprecated. So, it turns out that this works for both 3.8 and 3.9. The fact that the checks were completing was my first clue :-)

argilo:
> So, it turns out that this works for both 3.8 and 3.9

And 3.7! Thanks for this.

PR #911, opened 2020-12-20
src/main.cpp: CodeQuality

hb9fxq:
- CodeQuality: Code style fixes for main.cpp
- Project Setup: Added /CMakeLists.txt.user .gitignore


.gitignore
```diff
@@ -7,4 +7,4 @@
 /version.txt
 .idea/
 cmake-build*/
-
+/CMakeLists.txt.user
```
argilo, line 10: Could you put newlines at the end of these files? Thanks.

hb9fxq: Done

src/applications/gqrx/main.cpp
```diff
@@ -231,4 +228,4 @@ static void list_conf(void)
             std::cout << conf_files.at(i).toLocal8Bit().constData() << std::endl;
         }
     }
-}
+}
```
argilo, line 231: A newline is needed here as well.

hb9fxq: Got it :-)

PR #912, opened 2020-12-20
Update deprecated QString::SkipEmptyParts to Qt::SkipEmptyParts

shuyuan-liu:
Fixes compilation warnings like the following:

```
/home/shuyuan/Dev/gqrx/src/qtgui/dxc_options.cpp:118:36: warning: 'split' is deprecated: Use Qt::SplitBehavior variant instead [-Wdeprecated-declarations]
            spot = incomingMessage.split(" ", QString::SkipEmptyParts);
                                   ^
```

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -2391,7 +2391,7 @@ void MainWindow::on_actionAddBookmark_triggered()
         info.bandwidth = ui->plotter->getFilterBw();
         info.modulation = uiDockRxOpt->currentDemodAsString();
         info.name=name;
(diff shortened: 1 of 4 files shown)


shuyuan-liu:
It looks like the CI tests don't like `Qt::SkipEmptyParts`... It compiles just fine on my machine with Qt 5.15.2. 

[The Qt documentation on this](https://doc.qt.io/qt-5.15/qt.html#SplitBehaviorFlags-enum) says the new `Qt::SkipEmptyParts` was introduced in Qt 5.14. This is probably a very recent change.

argilo:
Yes, unfortunately this change will have to wait a few years until Linux distributions are using 5.14 or newer.

Issue #903, opened 2020-12-10
File recording is always in stereo, even for mono demodulators


darrenfreeman:
My archive of recordings is twice the size it should be, because gqrx always records in stereo. Even if, for example, I'm using SSB or narrow FM.

I think gqrx should automatically select the number of channels to record to file, based on the demodulator's capability. Mono demodulators should record in mono.

PR #917, opened 2020-12-21
DRAFT PR: Default the WAV audio recorder to 1 channel/mono (Issue #903)

hb9fxq:
Defaults the WAV audio recorder to 1 channel/mono, recoding stereo only for defined demods. (Issue https://github.com/csete/gqrx/issues/903)

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -845,6 +845,8 @@ receiver::status receiver::set_demod(rx_demod demod, bool force)
     if (!force && (demod == d_demod))
         return ret;
 
+    d_is_stereo_type_demodulator = false;
+
     // tb->lock() seems to hang occasioanlly
     if (d_running)
     {
@@ -888,11 +890,13 @@ receiver::status receiver::set_demod(rx_demod demod, bool force)
     case RX_DEMOD_WFM_S:
         connect_all(RX_CHAIN_WFMRX);
         rx->set_demod(wfmrx::WFMRX_DEMOD_STEREO);
+        d_is_stereo_type_demodulator = true;
         break;
 
     case RX_DEMOD_WFM_S_OIRT:
         connect_all(RX_CHAIN_WFMRX);
         rx->set_demod(wfmrx::WFMRX_DEMOD_STEREO_UKW);
+        d_is_stereo_type_demodulator = true;
         break;
 
     case RX_DEMOD_SSB:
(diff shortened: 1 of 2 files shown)


hb9fxq:
Probably needs rework, because I'm not sure about the intended behaviour of the recorder, when switching demodulators while active recording... that has to be defined: 

a) auto restart recorder in all cases
b) auto restart recorder in case of change in mono/stereo demodulator

Any suggestions?

src/applications/gqrx/receiver.cpp
```diff
@@ -888,11 +890,13 @@ receiver::status receiver::set_demod(rx_demod demod, bool force)
...
+        d_is_stereo_type_demodulator = true;
         break;
 
     case RX_DEMOD_WFM_S_OIRT:
         connect_all(RX_CHAIN_WFMRX);
         rx->set_demod(wfmrx::WFMRX_DEMOD_STEREO_UKW);
+        d_is_stereo_type_demodulator = true;
         break;
 
     case RX_DEMOD_SSB:
```
pinkavaj, line 902: `RX_DEMOD_NONE` produces 2 channels and needs `d_is_stereo_type_demodulator = true`  (it is raw IQ)

src/applications/gqrx/receiver.cpp
```diff
@@ -1000,11 +1004,11 @@ receiver::status receiver::start_audio_recording(const std::string filename)
     // if this fails, we don't want to go and crash now, do we
     try {
 #if GNURADIO_VERSION < 0x030900
-        wav_sink = gr::blocks::wavfile_sink::make(filename.c_str(), 2,
+        wav_sink = gr::blocks::wavfile_sink::make(filename.c_str(), d_is_stereo_type_demodulator ? 2 : 1,
```
pinkavaj, line 1007: mey-be rename `d_is_stereo_type_demodulator` to somethink like `d_demod_channels` and set directly to 1 or 2?

pinkavaj:
> Probably needs rework, because I'm not sure about the intended behaviour of the recorder, when switching demodulators while active recording... that has to be defined:
> 
> a) auto restart recorder in all cases b) auto restart recorder in case of change in mono/stereo demodulator c) leave recorder running in initial selected mono/stereo state
> 
> Any suggestions?

I would go for new file each time the demod is changed. The reasoning is as follow:

* Keepeng the old mode would break IQ and WFM when the old record is mono and would not save any space in case of switch from stereo to mono demodulator - not really usefull
* Restart the recording at least when the number of recording channels change will fix the previous issue
* But creating new file for each demodulator change would split the recordings in IMHO usefull way

Changing output format for audio UDP stream will probably break receiving app, so althought inconsistent probably should stay in stereo mode all the time (just duplicate the mono).

Considering all the fuss, I'm not 100% sure the saved space is worth the complexity, but I guess this should be best proven by the code.

PR #918, opened 2020-12-25
Keep custom filter width

### LINKED ISSUE #908: Changing filter shape resets custom filter width to "normal"
If I drag the filter edges on the pandapter to adjust the filter width, then change the filter shape in the Receiver Options tab, the filter width gets reset to what you would get when selecting width "normal". Despite this, the filter width dropdown box still shows the previous "User (...k)" filter width.

## Screenshots

Starting GQRX with Normal filter width and Normal filter shape:

GQRX, Normal filter width and shape

Drag on filter edges to increase / decrease filter width. The filter width value in the Receiver Options tab is updated properly.

Change filter width

Change filter shape to sharp / soft. The actual filter width gets reset, both on the pandapter and as I can tell from the audio output. However the filter width value in the Receiver Options tab is not updated.

Change filter shape, width gets reset

shuyuan-liu:
Fixes #908. Now the custom filter width is kept when changing filter shape. The custom filter width is stored separately for each demod, and switching between demod with the User filter width selected will recall the previously set filter widths for the new demod. These custom filter widths are not yet stored in the settings.

### DIFF
--- a/src/qtgui/dockrxopt.cpp
+++ b/src/qtgui/dockrxopt.cpp
@@ -31,20 +31,20 @@
 QStringList DockRxOpt::ModulationStrings;
 
 // Filter preset table per mode, preset and lo/hi
-static const int filter_preset_table[DockRxOpt::MODE_LAST][3][2] =
-{   //     WIDE             NORMAL            NARROW
-    {{      0,      0}, {     0,     0}, {     0,     0}},  // MODE_OFF
-    {{ -15000,  15000}, { -5000,  5000}, { -1000,  1000}},  // MODE_RAW
-    {{ -10000,  10000}, { -5000,  5000}, { -2500,  2500}},  // MODE_AM
-    {{ -10000,  10000}, { -5000,  5000}, { -2500,  2500}},  // MODE_NFM
-    {{-100000, 100000}, {-80000, 80000}, {-60000, 60000}},  // MODE_WFM_MONO
-    {{-100000, 100000}, {-80000, 80000}, {-60000, 60000}},  // MODE_WFM_STEREO
-    {{  -4000,   -100}, { -2800,  -100}, { -2400,  -300}},  // MODE_LSB
-    {{    100,   4000}, {   100,  2800}, {   300,  2400}},  // MODE_USB
-    {{  -1000,   1000}, {  -250,   250}, {  -100,   100}},  // MODE_CWL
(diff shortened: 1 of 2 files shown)


src/qtgui/dockrxopt.cpp
```diff
@@ -31,20 +31,20 @@
...
+    {{ -10000,  10000}, { -5000,  5000}, { -2500,  2500}, { -5000,  5000}},  // MODE_NFM
+    {{-100000, 100000}, {-80000, 80000}, {-60000, 60000}, {-80000, 80000}},  // MODE_WFM_MONO
+    {{-100000, 100000}, {-80000, 80000}, {-60000, 60000}, {-80000, 80000}},  // MODE_WFM_STEREO
+    {{  -4000,   -100}, { -2800,  -100}, { -2400,  -300}, { -2800,  -100}},  // MODE_LSB
+    {{    100,   4000}, {   100,  2800}, {   300,  2400}, {   100,  2800}},  // MODE_USB
+    {{  -1000,   1000}, {  -250,   250}, {  -100,   100}, {  -250,   250}},  // MODE_CWL
+    {{  -1000,   1000}, {  -250,   250}, {  -100,   100}, {  -250,   250}},  // MODE_CWU
+    {{-100000, 100000}, {-80000, 80000}, {-60000, 60000}, {-80000, 80000}},  // MODE_WFM_STEREO_OIRT
+    {{ -10000,  10000}, { -5000,  5000}, { -2500,  2500}, { -5000,  5000}}   // MODE_AMSYNC
 };
```
shuyuan-liu, line 48: A column is added in the filter preset table to store the user filter width settings. This makes the table no longer `const`, but saves the trouble of adding extra code to deal with user filter widths separately when changing filter width.

src/qtgui/dockrxopt.cpp
```diff
@@ -599,10 +604,22 @@ void DockRxOpt::on_filterCombo_activated(int index)
  * Note that the modes listed in the selector are different from those defined by
  * receiver::demod (we want to list LSB/USB separately but they have identical demods).
  */
-void DockRxOpt::on_modeSelector_activated(int index)
+void DockRxOpt::on_modeSelector_activated(int mode_index)
```
shuyuan-liu, line 607: Renamed `index` to `mode_index` to avoid confusion with `filter_index`.

PR #921, opened 2020-12-27
Changing filter shape resets custom filter width to "normal" #908

ulrichloose:
A simple solution for this behavior and small error fix in MainWindow::onBookmarkActivated...

I think the same error exists in void MainWindow::setPassband(int bandwidth) - but I can not verify it because I have no remote...


shuyuan-liu:
@ulrichloose 

I think it looks like your recent commits didn't go to the right branch. The new keyboard shortcuts and mouse wheel page step, for example, seem unrelated to this pull request.

Anything you push to the branch you created the pull request with will continue to show up here. Maybe check which branch your are in?

ulrichloose:
Thanks for the hint,

As you can see, I'm a newbie with git and I already suspected that I would have problems ...
I still have to familiarize myself with this and won't "publish" any further changes for the time being.

I'm sorry to have done something wrong

shuyuan-liu:
@ulrichloose No don't worry, we all make mistakes ;)

ulrichloose:
I accidentally cited my master as the source for this request, so I close it and open a new one...

PR #924, opened 2021-01-01
Changing filter shape resets custom filter width to "normal" #908

ulrichloose:
for #908

This is the second pull request because I accidentally cited my master as the source for the first one ... I apologize for that - as a newcomer to github I still have some problems ... If I read that correctly, I can now set the first wrong pull request to done and the problem is eliminated.


ulrichloose:
I got lost somehow and that is not an improvement, but partly a worsening.

The first commit solves the real problem that should be solved correctly - but the changes in the mainwindow with the bookmarks are just garbage ...

I want to delete that and publish the really simple version that only does the reset again.

PR #927, opened 2021-01-01
A few little things with no particular connection ...

ulrichloose:


shuyuan-liu:
Three automatic audio FFT range looks like a useful feature to me. It should also apply to AM, USB, LSB and Raw IQ; would you consider implementing it for those modes?

kgarrels:
> Three automatic audio FFT range looks like a useful feature to me. It should also apply to AM, USB, LSB and Raw IQ; would you consider implementing it for those modes?

I made an experimental "auto range" feature, plus:
- limited horizontal scrolling when possible
- noise blanker for visualization (fft, waterfall), my neighbor runs an electric fence

Find it here: https://github.com/kgarrels/gqrx/tree/panadapter

- autostart DSP
- remember fullcreen in settings
- some changes for DEMOD OFF
- autorange for FFT and waterfall
- shift waterfall horizontally if possible (to be improved)
- add noise blanker for visualization

ulrichloose:
There is only a setting option for the audio FFT bandwidth with CWL and CWH and so that the "center frequency" is displayed in the middle of the audio FFT, you have to set this to the frequency set by the user. Personally, I would even prefer it if there were a zero in the middle and left minus right plus of the "center frequency".

All other types of modulation always start on the left at zero and, depending on the bandwidth, a narrower or wider spectrum is displayed. You can easily set the right one yourself with a click and scroll wheel on the horizontal scale.

The only thing that could be done would be not to use the fixed specifications that are stored in the program, but the set filter bandwidth as a basis.

PR #930, opened 2021-01-05
A cosmetic change in demodulator list...

ulrichloose:
The new demodulator AM-Sync appears somewhat illogically at the end of the selection list. I think the entry belongs to the next position after AM ...

### DIFF
--- a/src/qtgui/dockrxopt.h
+++ b/src/qtgui/dockrxopt.h
@@ -67,15 +67,15 @@ public:
         MODE_OFF        = 0, /*!< Demodulator completely off. */
         MODE_RAW        = 1, /*!< Raw I/Q passthrough. */
         MODE_AM         = 2, /*!< Amplitude modulation. */
-        MODE_NFM        = 3, /*!< Narrow band FM. */
-        MODE_WFM_MONO   = 4, /*!< Broadcast FM (mono). */
-        MODE_WFM_STEREO = 5, /*!< Broadcast FM (stereo). */
-        MODE_LSB        = 6, /*!< Lower side band. */
-        MODE_USB        = 7, /*!< Upper side band. */
-        MODE_CWL        = 8, /*!< CW using LSB filter. */
-        MODE_CWU        = 9, /*!< CW using USB filter. */
-        MODE_WFM_STEREO_OIRT = 10, /*!< Broadcast FM (stereo oirt). */
-        MODE_AM_SYNC    = 11, /*!< Amplitude modulation (synchronous demod). */
+        MODE_AM_SYNC    = 3, /*!< Amplitude modulation (synchronous demod). */
+        MODE_NFM        = 4, /*!< Narrow band FM. */
+        MODE_WFM_MONO   = 5, /*!< Broadcast FM (mono). */
+        MODE_WFM_STEREO = 6, /*!< Broadcast FM (stereo). */
+        MODE_LSB        = 7, /*!< Lower side band. */
+        MODE_USB        = 8, /*!< Upper side band. */
+        MODE_CWL        = 9, /*!< CW using LSB filter. */
+        MODE_CWU        = 10, /*!< CW using USB filter. */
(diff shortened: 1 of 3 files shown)


src/qtgui/dockrxopt.h
```diff
@@ -67,15 +67,15 @@ class DockRxOpt : public QDockWidget
...
-        MODE_NFM        = 3, /*!< Narrow band FM. */
-        MODE_WFM_MONO   = 4, /*!< Broadcast FM (mono). */
-        MODE_WFM_STEREO = 5, /*!< Broadcast FM (stereo). */
-        MODE_LSB        = 6, /*!< Lower side band. */
-        MODE_USB        = 7, /*!< Upper side band. */
-        MODE_CWL        = 8, /*!< CW using LSB filter. */
-        MODE_CWU        = 9, /*!< CW using USB filter. */
-        MODE_WFM_STEREO_OIRT = 10, /*!< Broadcast FM (stereo oirt). */
-        MODE_AM_SYNC    = 11, /*!< Amplitude modulation (synchronous demod). */
+        MODE_AM_SYNC    = 3, /*!< Amplitude modulation (synchronous demod). */
```
csete, line 70: I think these numbers are stored in the configuration files and changing the order will restore the wrong mode. Perhaps the mode ID should be split from the dropdown index and stored as user data instead.

ulrichloose: Thanks for the hint...

I have now tested it again very carefully and can not find any problems.

The last set demodulator is entered in the default.conf via

https://github.com/csete/gqrx/blob/be965db4ab484c20f93543ca551571c3684b4aa6/src/qtgui/dockrxopt.cpp#L454-L460

so that the new order is retained and is also correctly set again when reading in.

https://github.com/csete/gqrx/blob/be965db4ab484c20f93543ca551571c3684b4aa6/src/qtgui/dockrxopt.cpp#L445-L450

I also completely tested the new keyboard shortcuts and the tcp remote. The new order is also taken into account. 

Because I am an autodidact and beginner when it comes to programming, I have to admit that I have only changed something that already exists. Such an extensive change in the direction of separating dropdown index and storing as userdata is beyond my capabilities. (:-(

ulrichloose: I think I misunderstood you.... You want the old settings to be compatible with the new order, so that a user can use an older version side by side with the new one?

If so, I think I can do that....

ulrichloose: another question about cosmetics...

It seems to me that MODE_WFM_STEREO_OIRT belongs to another place - just after MODE_WFM_STEREO. Then the question arises whether MODE_LSB, MODE_USB, MODE_CWL and MODE_CWU, because it is also amplitude modulation, belongs to a Position prior the FM entries - in my view even before AM...

The most logical order?

MODE_OFF
MODE_RAW
MODE_LSB
MODE_USB
MODE_CWL
MODE_CWU
MODE_AM
MODE_AM_SYNC
MODE_NFM
MODE_WFM_MONO
MODE_WFM_STEREO
MODE_WFM_STEREO_OIRT

csete: > You want the old settings to be compatible with the new order, so that a user can use an older version side by side with the new one?

Indeed, my concern is that when the user upgrades to the new version of gqrx, the mode ID stored in the configuration file will restore a different mode than what the user expects. Of course, this will only happen once per configuration file and can be considered a minor inconvenience, however, since gqrx is a mature application, such scenarios should at least be considered and assessed whether it's okay.

I admit that it was a bad idea to store the index in the configuration file and perhaps it is time to fix it once and for all. Such a fix could be:

1. Change the mode order to whatever you feel is natural
2. Add common functions to convert between mode string and numerical ID.
3. Change the TCP interface to use the above functions.
4. Change the configuration to store a mode string instead of a numeric ID.
5. Increment the config version to 3 and add function to "upgrade" config version 2 mode ID to the new ID when reading the configuration.

I don't know how much work this is or whether it's worth it, but surely it would be an advantage to "decouple" the mode ID  stored in the configuration files from the internal array index.

PR #934, opened 2021-01-11
Extend WFM filter width to +-192kHz

jj1bdx:
This patch extends the filter width range for WFM from +-120kHz to +-192kHz. In a real-world example, the +-192kHz deviation may happen quite often.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1125,7 +1125,7 @@ void MainWindow::selectDemod(int mode_idx)
     case DockRxOpt::MODE_WFM_STEREO:
     case DockRxOpt::MODE_WFM_STEREO_OIRT:
         /* Broadcast FM */
-        ui->plotter->setDemodRanges(-120e3, -10000, 10000, 120e3, true);
+        ui->plotter->setDemodRanges(-192e3, -10000, 10000, 192e3, true);
         uiDockAudio->setFftRange(0,24000);  /** FIXME: get audio rate from rx **/
         click_res = 1000;
         if (mode_idx == DockRxOpt::MODE_WFM_MONO)


argilo:
Thank you for your contribution, and sorry for the delay in responding.

I agree that it would be desirable to increase the WFM filter width. However, it's not as simple as adjusting the limits in `setDemodRanges`. For one thing, the WFM receiver currently uses a quadrature rate of 240 ksps, so anything beyond ±120 kHz would result in aliasing:

https://github.com/csete/gqrx/blob/46c3199a5717d548636c2bb0f57ffd9ce58854d7/src/receivers/wfmrx.cpp#L29

In addition, the downconverter stage filters out signals outside ±120 kHz:

https://github.com/csete/gqrx/blob/46c3199a5717d548636c2bb0f57ffd9ce58854d7/src/dsp/downconverter.cpp#L29

Both of those values would need to be increased as well.

jj1bdx:
@argilo Yeah, I guess many related parts of the source code should be fixed, as you suggested.

pinkavaj:
It is really the case the deviation can go as width as +-192 kHz? Is this for normal FM analog broadcast eg. https://en.wikipedia.org/wiki/FM_broadcast_band#Deviation_and_bandpass or somethink else?

@jj1bdx any chance to finish this?

argilo:
I don't think 192 kHz deviation is common (75 kHz is customary for broadcast stations), but the bandwidth of an FM signal can be substantially wider than the deviation.

vladisslav2011:
Does anybody have an example of station with 192kHz deviation?

Even in my QTH, where FM broadcasters are not following any standard and are trying to make the station louder, than others, I have never seen more, than 220kHz bandwidth and more, than 100kHz deviation. Overcompressed and clipped by the modulator filter distorted signals are common.

argilo:
Another point to keep in mind is that the 240 ksps channel cannot actually pass 240 kHz of bandwidth without distortion. The bandwidth will necessarily be lower because filters are not perfect. It may be that some distortion of the baseband signal is occurring due to the channel filter.

jj1bdx:
Observation on NHK-FM in Tokyo and nationwide Japan has proven that sometimes the stations may cause +- >100kHz or up to +-150kHz deviation. OTOH many radio stations in Japan conform with limiting to +-75kHz deviation. See https://github.com/jj1bdx/airspy-fmradion for an implementation example.

rag2:
To receive an FM station with ±75kHz peak deviation with *low 
distortion* requires a wider bandwidth than Carson's rule suggests. 
Hi-Fi FM tuners have offered bandwidths up to 400kHz to provide 
distortion down to 0.1%. (It's the consequence of those crafty Bessel 
functions putting energy in the higher order sidebands.)

That is the justification for being able to select filter widths up to 
384kHz (or 400kHz.)

Robin, G8DQX

vladisslav2011:
The clipping is clearly visible on the main panadapter/waterfall. And the signal fits into selected channel bandwidth.
 This distortion is not Gqrx issue. Every hardware receiver, I've tried, outputs distorted sound: a car audio system, a walkie-talkie with FM radio, an mp3 player with FM radio... 
gqrx-wfm-bad-station
It looks like the station at 105.7 does not fit into the filter bandwidth and should sound bad, but it sounds perfectly and gives perfect RDS decode. The station at 106.1 fits into the filter bandwidth, but it sounds bad with strong distortion and gives no RDS decode. The abnormal spectral power distribution is clearly visible on the panadapter.

argilo:
I played a bit with gr-rds, and confirmed that widening the bandwidth (from 200 kHz to 270 kHz) reduces distortion and improves RDS signal-to-noise ratio: https://github.com/bastibl/gr-rds/pull/80

I may take a crack at making the necessary adjustments in Gqrx to allow for higher bandwidth.

Issue #933, opened 2021-01-09
PlutoSDR support (macOS / pre-compiled)


chilippso:
I'm curious whether the precompiled macOS build supports the PlutoSDR, since I can't get it to work, but gnuradio works fine on my machine (all runtime dependencies are installed).

Maybe someone can give me some advice.

Thanks in advance!

argilo:
The release builds don't currently have Pluto support. I'll see about adding that.

chillophil:
Alright, in the meantime I discovered why I was not able to connect my PlutoSDR with Gqrx but with "vanilla" GnuRadio: The app package contains all the runtime files on its own - so it is not using the machine wide GnuRadio installation and plugins but the provided/packages ones.

Maybe I can add PlutoSDR support back again on my own and raise a pull request for that - if I find some free (spare) time.

As far as I know, PlutoSDR support was previously realized via gr-osmosdr-gqrx on branch "plutosdr" - is this still valid or does it need a complete "rework"?

I am asking, because I would like to add PlutoSDR support via gr-iio instead, so we can keep "vanilla" qr-osmosdr for all the other devices.

argilo:
> As far as I know, PlutoSDR support was previously realized via gr-osmosdr-gqrx on branch "plutosdr" - is this still valid or does it need a complete "rework"?

I think a better approach would be to ship the [SoapyPlutoSDR](https://github.com/pothosware/SoapyPlutoSDR) module with the Gqrx disk image. I tried that yesterday (https://github.com/csete/gqrx/compare/master...argilo:add-pluto) but for some reason dylibbundler was not bundling its libiio and libad9361 dependencies.

chillophil:
I agree with you, that seems to be a good, if not the best, solution!

I will dig into the dylibbundler issue if you haven't already - so let me know ;)

argilo:
Yes, I'd appreciate help with dylibbundler.

ulrichloose:
libiio / libad9361 are not dylibs and can not be bundled with dylibbundler...

libiio is a framework installed in library/frameworks via sudo

ulrichloose:
Another small problem: You would have to change SOAPY_SDR_ROOT in the info.plist so that the driver that comes with the bundle of gqrx is found. However, this means that the user is then off other soapydrivers that are probably in /usr/local/lib/SoapySDR/modules0.X

Whether this is still a good idea to pack the three Pluto files into the gqrx bundle would have to be questioned.

[...]

PR #936, opened 2021-01-16
A possible solution for #933 

ulrichloose:
I don't have a PlutoSDR, but here is a solution that should probably work, in any case the workflow runs without errors but I can't check that because I'm not using a Mac with the latest operating system.

Maybe someone will check this with macOS 10.15 and a PlutoSDR?

I don't know whether the whole thing meets the requirements at all. But it should be enough to address the PlutoSDR from Gqrx. If you want more, you just have to install the necessary drivers and additional programs.

I have also tested this with another shellscript and a changed build.yml file like

name: Compile
        working-directory: build
        run: make
      - name: Build iio, ad9361 and SoapyPluto
        run: ./iio_ad9361.sh

it works also.

Furthermore I have tried to patch the brewed files but they are so old - not HEAD - and it takes so much time to get an error message because always all steps are started from the beginning.

`cp /Library/Frameworks/iio.framework/iio /usr/local/lib/libiio.dylib
install_name_tool -id "/usr/local/lib/libiio.dylib" /usr/local/lib/libiio.dylib
cp /Library/Frameworks/ad9361.framework/ad9361 /usr/local/lib/libad9361.dylib
install_name_tool -id "/usr/local/lib/libad9361.dylib" /usr/local/lib/libad9361.dylib
install_name_tool -delete_rpath /Library/Frameworks /usr/local/lib/libad9361.dylib
install_name_tool -change @rpath/iio.framework/Versions/0.21/iio /usr/local/lib/libiio.dylib /usr/local/lib/libad9361.dylib
install_name_tool -change @rpath/iio.framework/Versions/0.21/iio /usr/local/lib/libiio.dylib /usr/local/lib/SoapySDR/modules0.*/libPlutoSDRSupport.so
install_name_tool -change @rpath/ad9361.framework/Versions/0.2/ad9361 /usr/local/lib/libad9361.dylib /usr/local/lib/SoapySDR/modules0.*/libPlutoSDRSupport.so
`
is working locally on my mac, but not with the workflow. The cp gives an error file not found. I think there could be an error with the brew file - the destination /Library/Frameworks needs a sudo?!


ulrichloose:
I think I have found a better way...

PR #937, opened 2021-01-17
SoapyPlutoSDR support

ulrichloose:
In my opinion, Homebrew has a flaw because the two Analog Device Frameworks would have to be installed per sudo according to Library/Frameworks, but I can't find a sudo in the formula.

Here is the proposal to install the two frameworks per git and also SoapyPlutoSDR because it uses the two frameworks to compile.

As a second step, a patch script is executed before the DMG file is created, which converts the frameworks into dylibs and also adjusts SoapyPlutoSDR accordingly.

The workflow runs flawlessly - but I don't have the possibility to test the result with 10.15.


argilo:
The library fails to load with the following error:
```
[ERROR] SoapySDR::loadModule(/Volumes/Case-sensitive test/Gqrx.app/Contents/soapy-modules/libPlutoSDRSupport.so)
  dlopen() failed: dlopen(/Volumes/Case-sensitive test/Gqrx.app/Contents/soapy-modules/libPlutoSDRSupport.so, 1): Library not loaded: /usr/local/lib/libiio.dylib
  Referenced from: /Volumes/Case-sensitive test/Gqrx.app/Contents/soapy-modules/libPlutoSDRSupport.so
  Reason: image not found
```

argilo:
It's probably necessary to add one more `-x` option here?

https://github.com/csete/gqrx/blob/6fce5aae8a906b6b7e067b0e1515c5d23fe3b0ae/macos_bundle.sh#L44

argilo:
This works after the changes to macos_bundle.sh. Thanks!

I squashed the commits and manually merged in 6c4c812b792709e8a57f22c9ab2f9d76f0058624.

PR #943, opened 2021-02-22
Misc UI changes

shuyuan-liu:
I made some mainly cosmetic changes to the UI, including:

- Use icons from system icon theme
- Make Lock (renamed Sync) button span across the two sliders that it syncs
- Use icons on the record and play buttons in the Audio dock panel
- Rearranging buttons here and there
- Make all input fields `MinimumExpanding` horizontally
- Removed information text to the right of the "FFT size", "Rate", "Time span" etc. because I didn't find them very useful
- Removed empty space at the bottom of IQ recorder which looks like unused'

Tested with icon themes Adwaita, Breeze, Papirus, and Numix. All icons do show up with these themes.

See screenshots below:

Screenshot from 2021-02-23 16-03-55

The Mac CI tests are failing with "gr-osmosdr has been disabled because it does not build". Am I doing anything wrong?


ulrichloose:
Just a small interim remark: There are also operating systems other than "Linux"...

shuyuan-liu:
> Just a small interim remark: There are also operating systems other than "Linux"...

If you are referring to icons, good point, I should have considered that. Unfortunately I don't have a Mac to test retrieving icons on, so either someone with a Mac will need to help with that, or I probably will just restore the original icons. I changed them in the first place because (personally at least) they were not that pretty, and using an oscilloscope-like icon for the IQ recorder / player didn't seem the most straightforward to me.

I was rushing to open the PR honestly. I'll close it, figure out why the Mac builds aren't working, maybe redesign the icons, make an issue first and PR again.

PR #955, opened 2021-04-15
include gr-iqbalance

ulrichloose:
As the name suggests - modified build.yml and ci.yml to include the component gr-iqbalance in the image...


argilo:
Thanks! It does seem like a good idea to include this in the macOS disk image.

To keep CI fast, I've intentionally stripped it down to include only Gqrx's direct dependencies. So I think it would be best to add gr-iqbal only to `build.yml`.

PR #956, opened 2021-04-16
Rc rds

kb8u-gqrx:
Added remote control commands to turn on and off FM RDS decoder and to return the RDS program identification number.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -23,6 +23,8 @@ Supported commands:
     Get the value of the gain setting with the name <gain_name>
  L <gain_name>_GAIN <value>
     Set the value of the gain setting with the name <gain_name> to <value>
+ p RDS Program Identification
+    Get the RDS program identification
  u RECORD
     Get status of audio recorder
  U RECORD <status>
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -24,7 +24,7 @@ Supported commands:
  L <gain_name>_GAIN <value>
     Set the value of the gain setting with the name <gain_name> to <value>
  p RDS Program Identification
-    Get the RDS program identification
+    Get the RDS PI code (in hexadecimal). Returns 0000 if not applicable.
  u RECORD
     Get status of audio recorder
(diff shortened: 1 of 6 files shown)


argilo:
Out of curiosity, what is your use case for this? Are you building an automated FM scanner?

I expect you'll be interested in the RDS demodulator improvements that were added in version 2.14.5. (See #984 for details.)

resources/remote-control.txt
```diff
@@ -31,6 +33,8 @@ Supported commands:
     Get DSP status
  U DSP <status>
     Set DSP status to <status>
+ U RDS <status>
+    Set RDS decoder to <status>.  Only functions in WFM mode.
```
argilo, line 37: There should probably be a corresponding `u RDS` comand to get the RDS status.

resources/remote-control.txt
```diff
@@ -23,6 +23,8 @@ Supported commands:
     Get the value of the gain setting with the name <gain_name>
  L <gain_name>_GAIN <value>
     Set the value of the gain setting with the name <gain_name> to <value>
+ p
+    Get the RDS PI code (in hexadecimal). Returns 0000 if not applicable.
```
argilo, line 27: Gqrx's remote control protocol is meant to be compatible with [rigctld's](https://www.systutorials.com/docs/linux/man/8-rigctld/). I see it uses `p` to get a parameter (which probably makes sense here) but also has a parameter name. Perhaps this should be `p RDS_PI`? That would also leave room for future expansion. (For example, it might be useful to fetch other RDS data.)

kb8u:
> 
> 
> Out of curiosity, what is your use case for this? Are you building an automated FM scanner?
> 
> I expect you'll be interested in the RDS demodulator improvements that were added in version 2.14.5. (See #984 for details.)

I have built a site mapping FM reception for North America, see [https://rabbitears.info/tvdx/fm_all_tuners](url) I have scanners working for SDR Sharp and SDR Console and a prototype for gqrx.  I'll get to work on your requested changes soon.

PR #966, opened 2021-07-11
Implement band plan ranges per region

LouDou:
Hi all,

I was interested in GQRX Band Plan showing ranges applicable to my own region (UK). To that end, I've updated the code to support toggling on and off each region in the band plan, and updating the default CSV file with data from https://rsgb.org/main/operating/band-plans


francisuk1989:
+1 for this idea!

Even if we have multiples of CSV files like so

UK Band Plan.csv
USA Band Plan.csv
SE Band Plan.csv

In the program we could just select the CSV file for the band plan.

LouDou:
Hi @francisuk1989 

That is another, different approach to bandplan management. Just as valid, I might add.

However, do users of this software really want to be messing about with custom CSV files and loading them in? Or should the software ship with a much more complete list of band information, and allow the user to filter that information in the GUI?

I'd personally be more comfortable with the latter, in which case the simple checkboxes for regions I implemented is also a sub-optimal approach. It might be more appropriate to structure the bandplan data in such a way that each entry is annotated with various labels, and then allow the user to filter by labels+values? In that way, the user could filter by attributes such as:

- region
- band
- mode
- country
- signal type (voice/data/etc)
- and anything else useful that could be suggested

How does that sound to other gqrx users?

PR #974, opened 2021-08-11
Updated PMR446 for A/D Added AIrwave/FRS/Airband

francisuk1989:
Added TETRA AIrwave for UK
Updated PMR46 fpr Analog & Digital
Added FRS (Family Radio Service) for USA (Like PMR446 for EU)
Updated Air Band Voice to 8.33k not sure if EU or US is the same? https://www.caa.co.uk/General-aviation/Aircraft-ownership-and-maintenance/8-33-kHz-radios/


pinkavaj (approved):

resources/bandplan.csv
```diff
@@ -104,11 +104,23 @@
 614000000, 698000000, AM,  10, #902E98BB, Broadcast Television
 698000000, 763000000, AM,  10, #902E98BB, Broadcast Television
 
+# Europe PMR446
+446000000, 446094000, NFM,  6250, #9000FF00, PMR446 Analog 1-8
```
pinkavaj, line 108: The end frequency here should be 446100000 (446.093750 + 6250) according to  https://en.wikipedia.org/wiki/PMR446#Analog_(NFM)_and_digital_TDMA_(DMR)

pinkavaj (changes requested):
Please update the band plan.

resources/bandplan.csv
AsciiWolf: Actually, this is still incorrect. There is 16 PMR446 channels used for analog. and 8 different ones for dPMR446. See for example this: https://cs.wikipedia.org/wiki/PMR_446#Technick%C3%A9_parametry

AsciiWolf: Sorry, also 16 channels for dPMR446, not 8.

AsciiWolf: I see that your linked page mentions 32 dPMR446 channels: "Before January 2018, only the upper 16 channels were allowed for digital FDMA operation."

This may be correct, however I am not sure.

pinkavaj: I guess it will not be possible to add all those details here, can we try to create some generally useful summary so the MR have chance to get it into `master`?

@AsciiWolf So the spacing above should be 12.5 rather than 6.25, right?

AsciiWolf: https://www.ctu.cz/sites/default/files/obsah/ctu/vseobecne-opravneni-c.vo-r/3/6.2016-9/obrazky/vo-r03-062016-09.pdf
(Page 2)

This is probably more accurate than Wikipedia (and should be valid for the whole European Union).

pinkavaj: Nice one! Would you be so kind and try to write here some suggestion what would be best to use here by your opinion?

AsciiWolf: Probably the best will be:
```
446006250, 446193750, NFM,  12500, #9000FF00, PMR446 Analog 1-16
446003125, 446196875, NFM,  6250, #9000FF00, PMR446 Digital 1-32
```

pinkavaj: The 446006250 is center of a channel, should not be here 446000000 to not cut the half of the channel? The same for apply for the upper bound.

AsciiWolf: Good point. :-)

PR #978, opened 2021-09-11
Bandplan settings

LouDou:
Continues and replaces https://github.com/csete/gqrx/pull/966

I went ahead and added more filtering options for the band plan, which also prompted me to put those settings into its own dock widget as well.

Screenshot_20210911_143458

I still think the csv file could do with some improvements as well:
 - better colouring for the UK Amateur bands
 - data for more regions and non-Amateur bands.


LouDou:
Updated the UK band plan regions as per the colours used on the RSGB website: 

Screenshot_20210911_154919

LouDou:
I've also included the bandplan update from https://github.com/csete/gqrx/pull/974 in the interest of avoiding a merge conflict.
@francisuk1989

LouDou:
Final update I think here, adds a table of the visible band plan items - double click to set the rx parameters, just like bookmarks:

Screenshot_20210911_173210

LouDou:
I've noticed though that there are items in the bandplan.csv which are >9.98GHz - which gqrx doesn't seem to support? I'm not going to attempt to fix that, I assume they are there for a reason.

LouDou:
Here's a preview of a 2nd band plan, using the ofcom data:
Screenshot_20210914_211707

I'm not including in this PR because it's another level of substantial difference, and I'm interested in any feedback on this one first.
If you want to check it out though, go to my fork: https://github.com/LouDou/gqrx/tree/ofcom-spectrum-map

argilo:
A couple first impressions:

1. When upgrading from an older version of Gqrx, the bandplan doesn't work because `bandplan.csv` is not updated. That's because Gqrx avoids overwriting the bandplan, assuming the user might have edited it: https://github.com/gqrx-sdr/gqrx/blob/4a67620a941b99c64c4e42cd9c4d02b31a0dcad4/src/qtgui/bandplan.cpp#L56-L61 Maybe that was a bad choice. I doubt that many users would edit the file. Should we just remove the `if` statement here?
2. When the Band Plan Settings panel in enabled, the Gqrx window suddenly becomes extremely tall and the Audio panel is pushed off the screen completely. This is because the panel appears between the Receiver Options and Audio panels, taking of a lot of vertical space. Screenshot from 2021-10-09 11-30-22 I think it would be better if it appeared as a tab grouped together with the Input controls, Receiver Options & FFT Settings panels. But even still, the Band Plan Settings panel is much taller than the others, making the minimum size of the Gqrx window quite a bit larger. Can it be made shorter? The Region and Country selectors could certainly be made shorter, as they have a lot of empty space currently. And maybe the selectors could be turned into drop-downs?

LouDou:
I suppose the new file which is built in should be saved using a different name and definitely avoid clobbering any user file that might exist with the same name. It can also probably detect if the file being read is the older format; so I reckon gqrx should raise a messagebox if either of these happen:

1. a new built in bandplan file was saved and used -> prompt the user to update/merge their custom data
2. gqrx tries to read in a bandplan file of an older version; it could in this case create backups and use a new default -> continue as case #1

----

As for the panel size, yes I agree it's pretty huge. I'm not sure if Qt has mult-select dropdowns, I can take look at that, it'd help if it does. I wonder also if the table of ranges could also be inside a collapsing container.

markjfine:
You could create a test file using the old standard bandplan.csv, then diff it against the existing file to see if anything changed.

If it changed, rename it to bandplan.csv.orig before clobbering it and pop up a warning to the user that lets them know where to find the old one.

An alternative would be to just read in the old file and add anything new to it before clobbering it with a new file.

Whatever new file is created, might want to have something in it that is isn't in original, like a magic number, that signifies the new format and nothing needs to be done to it.

argilo:
Hi @LouDou. I'm just popping in to let you know I haven't forgotten about this. My spare time for Gqrx work comes and goes, so I can't promise I'll get to it soon, but I will review this eventually.

LouDou:
Hi @argilo  - that's no problem, I understand and I'm in much the same position. Honestly, I wanted to spend a bit of time on the gqrx code over the last couple of weeks but life and (mild) sickness have got in the way.

pinkavaj:
@LouDou and @argilo Any possible progress here, this looks very interesting.

PR #984, opened 2021-10-03
Add symbol synchronization to RDS receiver

### LINKED ISSUE #981: Poor RDS receiver performance
In #788, @es5nhc noticed that Gqrx's RDS receiver performs poorly. As noted in https://github.com/gqrx-sdr/gqrx/pull/788#issuecomment-930370806, I had a closer look at this and found two contributing factors:

* There is no symbol synchronization.
* Half of each Manchester-encoded symbol is discarded.

Fixing these problems should result in much better performance.

argilo:
Fixes #981.

As noted in #981, the RDS receiver in Gqrx is poor because it lacks symbol synchronization. After making some improvements to gr-rds's symbol synchronization in https://github.com/bastibl/gr-rds/pull/51, I'm copying the changes into Gqrx here. The receiver appears to work much better than before.

/cc @es5nhc

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -5,6 +5,7 @@
        NEW: Added SoapyPlutoSDR to DMG release.
        NEW: Preliminary support for GNU Radio 3.10.
      FIXED: Compatibility with case-sensitive filesystems on MacOS.
+  IMPROVED: Better reception of RDS signals.
 
 
     2.14.4: Released December 28, 2020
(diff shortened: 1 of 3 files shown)


es5nhc:
I gave the patch a test drive with local and Latvian stations and I can confirm improved performance. Receiver stays synced noticeably better.

argilo:
Thanks for testing!

Issue #995, opened 2021-10-26
feature request: left-right arrow keys change digit


pidloop:
Hi, loving 2.14.5 on macOS 11.6 with CloudIQ, many thanks.

I can click on a frequency digit and increase or decrease with the up-down arrow keys. Would be great if I could move to an adjacent digit with the left-rights keys.

Thanks and 73, Elwood, WB0OEW

LouDou:
This is a little strange, it seems to work under Linux. The frequency control widget uses the standard `Qt::Key_Left`/`Qt::Key_Right` codes in the keypress handler, which should work if `Qt::Key_Up`/`Qt::Key_Down` are already working.

pidloop:
Yes, strange, I would have expected the implementation to be something like that.

Here are some more clues: if I take care to use just up and down and change digits with the mouse then all ok. But as soon as I hit left or right, then up and down stop working too. If use the mouse the select a different digit then use up or down the app soon crashes with the message **get_center_freq failed**.

LouDou:
Part 1: Left/Right moves the mouse cursor. On macOS (tested on Monterey) this is only allowed if you enable in settings > Security & Privacy > Privacy > Accessibility > Allow the apps below to control your computer > check Gqrx?

Part 2: even when the cursor is moved, the digit focus hasn't changed, therefore the next up/down doesn't work. I'll look into this a little bit

LouDou:
Part 3: I'm not sure the app really needs to even move the mouse cursor at all. The widget has enough code in it to allow changing the active digit without doing this. What's the use of moving the cursor? Especially as it actually will obscure the active digit as wel?

pidloop:
Still running Big Sur here. Gqrx.app was already checked.

If I said cursor I misspoke, I'm only interested in moving the digit focus with the l/r arrow keys. Thanks.

markjfine:
Don't mean to parachute in on someone else's feature request, but wouldn't it be more feasible to use the left/right arrow keys as a means for up/down tuning of a user-selected (or band-standard) stepsize, like on other tuning packages? That seems more of a logical, consistent, and user-friendly way to easily slew through a set of frequencies.

LouDou:
> If I said cursor I misspoke, I'm only interested in moving the digit focus with the l/r arrow keys. 

[...]

PR #999, opened 2021-11-02
change digit in CFreqCtrl without moving the mouse cursor

LouDou:
Fix for #995


argilo:
Thanks for the fix, and sorry for the long delay in reviewing this.

I took this for a spin, and it seems to work well. It's great that changing digits doesn't magically move the mouse anymore! I'll merge once CI is happy.

argilo (approved):

Issue #332, opened 2016-01-26
/dev/random issue for windows build


guruofquality:
I was testing the GQRX for the new PothosSDR build. Not surprisingly it bombs on the /dev/random FIXME part. Otherwise, I don't think its that far away from working since all of the dependencies are working.

```
    if (input_device.empty())
    {
        // FIXME: other OS
        src = osmosdr::source::make("file=/dev/random,freq=428e6,rate=96000,repeat=true,throttle=true");
    }
```

I'm not sure how to work around this since there is not a random device. But perhaps a cheap substitute would be to load a file on the fly with random floats and use that instead? I'm willing to try anything out.

csete:
Technically, we could read any file even if it contains NaN values since this source is only used until the configuration is read and not while the flow graph is running - at least not as long as the user config is OK. And if the user config is bad the application will crash anyway at some point.

Alternatively, we could try using an rtl_tcp source instead. I believe that is also available without any further hardware requirements.

guruofquality:
It works!

![screenshot](http://i.imgur.com/2zsU4pz.png)

My little patch
====
Not the best way to handle it, but basically we sub in a temp null and temp random file on repeat. I used Qt API to make it cross platform. This could be generalized better or copied into their own file. Also the wav source doesnt have a repeat option so the behaviour wouldnt be identical in that case:

```
diff --git a/src/applications/gqrx/receiver.cpp b/src/applications/gqrx/receiver.cpp
index 55254e3..d84eb48 100644
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -47,6 +47,55 @@
 #include <gnuradio/audio/sink.h>
 #endif
 
+#include <QFileInfo>
+#include <QTemporaryFile>
+#include <QDataStream>
+
+static std::string get_random_file(void)
+{
+    static std::string path;
+    if (path.empty())
+    {
+        path = "/dev/random";
+        QFileInfo checkFile(QString::fromStdString(path));
+        if (!checkFile.exists())
+        {
+            //static temp file persists until process end
+            static QTemporaryFile temp_file;
+            temp_file.open();
+            path = temp_file.fileName().toStdString();
+            {
+                QDataStream stream(&temp_file);
+                for (size_t i = 0; i < 1024*8; i++) stream << qint8(rand());
+            }
+            std::cout << "Created random file " << path << std::endl;
+        }
+    }
+    return path;
+}
+
[...]

PR #1001, opened 2021-11-09
allow app to start on windows - correctly quote file source names - use zero data in default source

LouDou:
Something about the 'random' file on Windows doens't work and causes an exception because the device string ends up malformed.

NUL is a valid (virtual) file on Windows which returns \0's and allows the app to start with the default device options before the user has selected a real input device.

### DIFF
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -29,24 +29,28 @@
 
 std::string receiver::get_random_file(void)
 {
-    static std::string path;
-    if (path.empty())
+#ifdef WIN32
+  return "NUL";
+#else
+  static std::string path;
+  if (path.empty())
+  {
+    path = "/dev/urandom";
+    QFileInfo checkFile(QString::fromStdString(path));
+    if (!checkFile.exists())
     {
-        path = "/dev/urandom";
-        QFileInfo checkFile(QString::fromStdString(path));
-        if (!checkFile.exists())
-        {
-            //static temp file persists until process end
-            static QTemporaryFile temp_file;
-            temp_file.open();
-            path = temp_file.fileName().toStdString();
-            {
-                QDataStream stream(&temp_file);
-                for (size_t i = 0; i < 1024*8; i++) stream << qint8(rand());
-            }
-            temp_file.close();
-            std::cout << "Created random file " << path << std::endl;
-        }
+      //static temp file persists until process end
+      static QTemporaryFile temp_file;
+      temp_file.open();
+      path = temp_file.fileName().toStdString();
+      {
+        QDataStream stream(&temp_file);
+        for (size_t i = 0; i < 1024 * 8; i++) stream << qint8(rand());
+      }
+      temp_file.close();
+      std::cout << "Created random file " << path << std::endl;
     }
-    return path;
+  }
+  return path;
+#endif
 }
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -21,17 +21,18 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#include "applications/gqrx/receiver.h"
+#include <iostream>
+#include <sstream>
+
+#include <boost/io/quoted.hpp>
 #include <QFileInfo>
 #include <QTemporaryFile>
 #include <QDataStream>
-#include <iostream>
+
+#include "applications/gqrx/receiver.h"
 
 std::string receiver::get_zero_file(void)
 {
-#ifdef WIN32
-  return "NUL";
-#else
   static std::string path;
   if (path.empty())
   {
@@ -45,12 +46,28 @@ std::string receiver::get_zero_file(void)
       path = temp_file.fileName().toStdString();
       {
         QDataStream stream(&temp_file);
-        for (size_t i = 0; i < 1024 * 8; i++) stream << qint8(rand());
+        for (size_t i = 0; i < 1024 * 8; i++) stream << qint8(0);
       }
       temp_file.close();
-      std::cout << "Created random file " << path << std::endl;
+      std::cout << "Created zero file " << path << std::endl;
     }
   }
   return path;
-#endif
+}
+
+std::string receiver::double_quote_path(std::string path)
+{
+  if (path == "") {
+    return path;
+  }
+
+  std::stringstream q1;
+  q1 << boost::io::quoted(path, '\\', '\'');
+  std::stringstream q2;
+  q2 << boost::io::quoted(q1.str(), '\\', ',');
+  std::string qp = q2.str();
+  if (qp[0] == ',')
+    qp = qp.substr(1, qp.size() - 2);
+
+  return qp;
 }
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -27,7 +27,7 @@
 #include <QDataStream>
 #include <iostream>
 
-std::string receiver::get_random_file(void)
+std::string receiver::get_zero_file(void)
 {
 #ifdef WIN32
   return "NUL";
@@ -35,7 +35,7 @@ std::string receiver::get_random_file(void)
   static std::string path;
   if (path.empty())
   {
-    path = "/dev/urandom";
+    path = "/dev/zero";
     QFileInfo checkFile(QString::fromStdString(path));
     if (!checkFile.exists())
     {
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1599,8 +1599,9 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
     storeSession();
 
     auto sri = (int)samprate;
-    auto devstr = QString("file='%1',rate=%2,throttle=true,repeat=false")
-            .arg(filename).arg(sri);
+    QString quotedFilename = receiver::double_quote_path(filename.toStdString()).c_str();
+    auto devstr = QString("file=%1,rate=%2,throttle=true,repeat=false")
+            .arg(quotedFilename).arg(sri);
 
     qDebug() << __func__ << ":" << devstr;
 
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -76,7 +76,7 @@ receiver::receiver(const std::string input_device,
 
     if (input_device.empty())
     {
-        src = osmosdr::source::make("file="+get_random_file()+",freq=428e6,rate=96000,repeat=true,throttle=true");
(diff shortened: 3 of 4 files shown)


LouDou:
For reference as well, here's how I made the build on Windows:

```
REM Step 1: install vcpkg and cmake
REM Step 2: use vcpkg to install: boost, qt5, mpir
REM Step 3: install Pothos SDR distribution: https://downloads.myriadrf.org/builds/PothosSDR/
REM Step 4: install python 3.9 (or version matching GNU Radio in the Pothos SDR installation)

set CMAKE="C:\Program Files\CMake\bin\cmake.exe"
set VCPKG_DIR=C:\DATA\SDKs\vcpkg
set POTHOSSDR_DIR=C:\DATA\Apps\PothosSDR
set PYTHON_DIR=C:\Python39

REM Step 5: Run cmake

md build-gqrx
cd build-gqrx

%CMAKE% ^
 -DCMAKE_TOOLCHAIN_FILE:PATH=%VCPKG_DIR%\scripts\buildsystems\vcpkg.cmake ^
 -DVolk_DIR=%POTHOSSDR_DIR%\lib\cmake\volk ^
 -DPython_ROOT_DIR=%PYTHON_DIR% ^
 -DBUILD_STATIC=ON ^
 -DBUILD_EXAMPLES=OFF ^
 ^
 ../gqrx

REM Step 6: Open solution in VS

start grqx.sln

REM Step 7: Build the solution in VS!
REM Step 8: Find the binaries in src/[Debug|Release]/
```

vladisslav2011:
Using /dev/urandom as an input file on *NIX is a very bad idea. Interpreting /dev/urandom as floats would give +inf, -inf, NaN's and very large correct numbers, but all of these are not correct sample values. These incorrect samples may result in, at least, iir filters runaway or may crash some blocks into exception. May it be better to choose /dev/zero, that outputs safe zeroes?

LouDou:
I agree it seems like asking for trouble. I've changed it and the function name.

argilo:
> Something about the 'random' file on Windows doens't work

The existing code already has a fallback for non-Unix systems. I'd like to understand why it's not working before making changes.

LouDou:
> The existing code already has a fallback for non-Unix systems. 

I don't see it; it's been called in the receiver constructor unconditionally and using /dev/<something> even on windows.

argilo:
The `if (!checkFile.exists())` block is meant to handle non-Unix operating systems.

argilo:
Also, the `get_random_file` function was originally created to support Windows: (#332, #335), and I've used Gqrx on Windows (installed through Conda) and didn't have a problem. So the problem you're seeing doesn't affect all Windows installations, and I'd like to know why that is before making changes.

LouDou:
Right. So for whatever reason, that branch wasn't taken on the Windows build. It also intends to create a temp file with random data, which would potentially cause the same issues as reading from /dev/urandom.

I'll run my build through the debugger and find out what's happening.

I have seen the same error behaviour in the gqrx binary shipped in the Pothos SDR bundle.

argilo:
Another point to consider is that we can probably just get rid of the ugly hack of reading from a random (or null) file. I did some of the work in #750.

I think a good path forward would be to use a Null Source block in place of the Osmosdr Source block. This will be necessary if we replace gr-osmosdr with gr-soapy, which is something I'd like to do anyway.

argilo:
> a good path forward would be to use a Null Source block

Or perhaps we could just leave the flow graph disconnected when no device is selected.

In either case, adjustments would need to be made, since the current code assumes that an Osmocom Source block is always present.

LouDou:
I found the problem with the original implementation. The temp file is created but the final path has a space in it:

```
std::string teststr = "file=C:/Users/My Name/AppData/Local/Temp/gqrx.nERGgV,freq=428e6,rate=96000,repeat=true,throttle=true";
```

Which does not parse in the osmocom source node. Incidentally, escaping the space spaces works.

```
std::string teststr = "file=C:/Users/My\\ Name/AppData/Local/Temp/gqrx.nERGgV,freq=428e6,rate=96000,repeat=true,throttle=true";
```

----

But, given that gqrx doesn't even need to read this file at all (it has no meaningful data in it), accepting the patch for `NUL` might be a quick fix until someone can do the additional work of not requiring the default source at all.

argilo:
> Which does not parse in the osmocom source node. Incidentally, escaping the space works.

Aha! I'm glad you were able to track down the root cause.

> accepting the patch for NUL might be a quick fix

I'd hesitate to do this for two reasons:

* It adds another `#ifdef` to the codebase. I've been trying to get rid of those, since conditionally compiled code makes debugging and testing more difficult.
* It doesn't address the underlying bug.

I'd rather fix the escaping bug. In addition to the two places that use `get_random_file`, it looks like the same bug is present here:

https://github.com/gqrx-sdr/gqrx/blob/75d108f7e9b0bb69bb1847173d957b8e41a0cc8f/src/applications/gqrx/mainwindow.cpp#L1602-L1603

argilo:
Actually I see there are quotes around the filename in that third case, so spaces and commas probably work fine. (But I guess there would still be breakage if the filename contained `'` or `\`.)

We could put quotes around the filename on these two lines to make spaces (and commas) work properly:

https://github.com/gqrx-sdr/gqrx/blob/75d108f7e9b0bb69bb1847173d957b8e41a0cc8f/src/applications/gqrx/receiver.cpp#L79

https://github.com/gqrx-sdr/gqrx/blob/75d108f7e9b0bb69bb1847173d957b8e41a0cc8f/src/applications/gqrx/receiver.cpp#L218

argilo:
On the gr-osmosdr side, parsing is done with `boost::escaped_list_separator`:

https://github.com/osmocom/gr-osmosdr/blob/a100eb024c0210b95e4738b6efd836d48225bd03/lib/arg_helpers.h#L66

Its escaping rules are documented here: https://www.boost.org/doc/libs/1_31_0/libs/tokenizer/escaped_list_separator.htm

argilo:
Does this patch fix the Windows startup issue?
```diff
diff --git a/src/applications/gqrx/receiver.cpp b/src/applications/gqrx/receiver.cpp
index fd6e44a0ec..e140428875 100644
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -76,7 +76,7 @@ receiver::receiver(const std::string input_device,
 
     if (input_device.empty())
     {
-        src = osmosdr::source::make("file="+get_random_file()+",freq=428e6,rate=96000,repeat=true,throttle=true");
+        src = osmosdr::source::make("file='"+get_random_file()+"',freq=428e6,rate=96000,repeat=true,throttle=true");
     }
     else
     {
@@ -215,7 +215,7 @@ void receiver::set_input_device(const std::string device)
     catch (std::exception &x)
     {
         error = x.what();
-        src = osmosdr::source::make("file="+get_random_file()+",freq=428e6,rate=96000,repeat=true,throttle=true");
+        src = osmosdr::source::make("file='"+get_random_file()+"',freq=428e6,rate=96000,repeat=true,throttle=true");
     }
 
     if(src->get_sample_rate() != 0)
```

vladisslav2011:
I don't think that this patch would fix the startup problem in all cases.
I suggest using boost::io:quoted from boost/io/detail/quoted_manip.hpp to escape the file name.

LouDou:
Just tried a few test cases with boost quoted:
```
      std::stringstream ss;
      ss << boost::io::quoted(path, '\\', '\'');
      path =  ss.str();
```
Using a single quote `'` because double quote `"` does not work at all.


and consuming it like
```
        std::string srcstr = "file=" + get_zero_file() + ",freq=428e6,rate=96000,repeat=true,throttle=true";
        src = osmosdr::source::make(srcstr);
```


Forming these device strings:

1. If the file path has spaces:
```
file='/tmp/test space/\'foo.ENBmWW',freq=428e6,rate=96000,repeat=true,throttle=true
```
This works.


2. If the file path also contains the quote character:
```
file='/tmp/test space/\'foo.ENBmWW',freq=428e6,rate=96000,repeat=true,throttle=true
```
It doesn't work.

vladisslav2011:
It works. You have to escape the file name twice:
```
#include <iostream>
#include <sstream>
#include <boost/io/detail/quoted_manip.hpp>


int main()
{
    auto fn="gqrx',t 123#\\.raw";
    std::stringstream ss1;
    std::stringstream ss2;
    std::string res;
    ss1<<boost::io::quoted(fn,'\\','\'');
    ss2<<boost::io::quoted(ss1.str(),'\\',',');
    res=ss2.str();
    if(res[0]==',')
        res=res.substr(1,res.size()-2);
    std::cout<<"file="<<res<<std::endl;

    return 0;
}
```
result:
```
file='gqrx\\'\,t 123#\\\\.raw'
```

LouDou:
I've pushed a change which uses this quoting strategy wherever a "file=" device string is used.

It'll need some testing on other Windows boxes; I've got inconsistent build behaviour here:
- Debug build - cannot start, crashes trying to create the top-block, therefore doesn't even get to this code
- RelWithDebInfo - produces empty strings for the temp_file name, therefore the "file=" string becomes malformed
- Release - seems to work, but as I cannot attach the debugger nor see any console output, I cannot see the exact string it created and used

🤷

vladisslav2011:
>Release - seems to work, but as I cannot attach the debugger nor see any console output

I have no windows machine to test, but maybe this would help with console output:

src/CMakelists.txt
```
#build a win32 app, not a console app
if (WIN32)
    set(CMAKE_EXE_LINKER_FLAGS "/entry:mainCRTStartup ${CMAKE_EXE_LINKER_FLAGS}")
    set_target_properties(${PROJECT_NAME} PROPERTIES WIN32_EXECUTABLE ON)
endif (WIN32)
```
Remove this "if" block and gqrx should become windows console application (I guess).

argilo:
> You have to escape the file name twice

I had a look under the hood in gr-osmosdr and the outer escaping layer actually uses space as the separator instead of comma. And I guess technically that layer of escaping should be done on the whole argument list rather than just the filename.

Also, we can use `std::quoted` and avoid another dependence on Boost if we update Grqx to use C++14. (I don't see a reason not to do that.)

argilo:
Actually I don't think the difference in delimiter matters, since `std::quoted` always quotes the string.

The following escaping code should work properly:
```cpp
#include <iostream>
#include <iomanip>
#include <sstream>

int main()
{
    std::string fn = "a b\\c'd.raw";
    std::stringstream ss1;
    std::stringstream ss2;
    std::string res;

    ss1 << std::quoted(fn, '\'', '\\');
    ss2 << std::quoted(ss1.str(), '\'', '\\');
    res = ss2.str();

    std::cout << res << std::endl;
}
```

argilo:
I did a bit of cleanup and put these changes into #1012. Please let me know what you think.

argilo:
Closing, since this was fixed in #1012.

PR #1004, opened 2021-11-21
Dsp freeze fix

### LINKED ISSUE #1003: DSP freeze and device lost after sample rate change
When the sample rate (or filter bandwidth) is changed in running state, the DSP freezes with the following error log:
```
SaveConfig
New LNB LO: 0 Hz
saveSettings *** FIXME_ SQL on/off
New FFT rate: 25 Hz
Loading configuration from: "/home/vlad/.config/gqrx/default.conf"
Configuration file: "/home/vlad/.config/gqrx/default.conf"
Set input device:
  old: rtl=0
  new: rtl=0
gr-osmosdr 0.1.4 (0.1.4) gnuradio 3.7.11
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy airspyhf soapy redpitaya freesrp 
Using device #0 Realtek RTL2838UHIDIR SN: 00000013
usb_claim_interface error -6

FATAL: Failed to open rtlsdr device.

Trying to fill up 1 missing channel(s) with null source(s).
This is being done to prevent the application from crashing
due to gnuradio bug #528.

New antenna selected: ""
********************
********************
Set output device:
   old: 
   new: 

Failed to set RX input rate to 2.4e+06
Your device may not be working properly.
```

Reproduced with Hackrf one, RTL-SDR.
System: Ubuntu 18.04.6 LTS
Gqrx was built from source from current git with g++-7 (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0.
Git head at build time was at 2172ef1a96d33a81675678033dc5e81f2761fe1e.

Steps to reproduce:
1. Select a real device (RTL-SDR for example)
2. Start DSP, adjust gain, tune to some frequency, set up demodulator and waterfall, etc
3. Open the device dialog and change the input rate (for example. Changing buffers/bias without changing device triggers the bug too)
4. Confirm changes

Expected behavior:
Sample rate is changed, DSP restarts with new sample rate, waterfall/panadapter continues to update with new data, sound continues to output (if tuned frequency is still in new bandwidth)

Observed behavior:
Sample rate is changed, DSP stops, starting DSP does not show nor waterfall/panadapter with new data, nor sound.

vladisslav2011:
Fix  #1003.
Switching input devices works even on the fly without DSP freeze/open errors.
Changing device options like direct sampling/bias tee works on the fly too.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -208,7 +208,18 @@ void receiver::set_input_device(const std::string device)
         tb->disconnect(src, 0, iq_swap, 0);
     }
 
+#if GNURADIO_VERSION < 0x030802
+    //Work around GNU Radio bug #3184
+    //temporarily connect dummy source to ensure that previous device is closed
+    src = osmosdr::source::make("file="+escape_filename(get_zero_file())+",freq=428e6,rate=96000,repeat=true,throttle=true");
+    tb->connect(src, 0, iq_swap, 0);
+    tb->start();
+    tb->stop();
+    tb->wait();
+    tb->disconnect(src, 0, iq_swap, 0);
+#else
     src.reset();
+#endif
 
     try
     {
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -2,6 +2,7 @@
     2.15.1: In progress...
 
      FIXED: DSP stops when device is changed.
+     FIXED: USB error when device settings are changed.
 
 
       2.15: Released December 15, 2021


src/applications/gqrx/receiver.cpp
```diff
@@ -178,14 +178,25 @@ void receiver::stop()
...
         return;
+    if (input_devstr.compare(device) == 0)
+    {
+#ifndef QT_NO_DEBUG_OUTPUT
+        qDebug() << "No change in input device:";
+        qDebug() << "  old: " << input_devstr.c_str();
+        qDebug() << "  new: " << device.c_str();
+#endif
+        return;
+    }else{
```
boushley, line 193: Looks like the rest of the code puts parentheses on their own lines. Should we update this to match?

src/applications/gqrx/receiver.cpp
```diff
@@ -194,7 +205,7 @@ void receiver::set_input_device(const std::string device)
     {
         tb->stop();
         tb->wait();
-    }
+    };
```
boushley, line 208: Does the semicolon here have a purpose? C++ is not my strong suit but I don't think this semicolon is adding any value, right?

src/applications/gqrx/receiver.cpp
```diff
@@ -206,7 +217,13 @@ void receiver::set_input_device(const std::string device)
         tb->disconnect(src, 0, iq_swap, 0);
     }
 
-    src.reset();
+    //temporarily connect dummy source to ensure that previous device is closed
```
boushley, line 220: Thanks for the helpful comment explaining the purpose of the dummy device connect 👍

boushley (commented):
👋 I'm new here, but this looks like a clean PR after a couple formatting changes.

src/applications/gqrx/receiver.cpp
vladisslav2011: Fixed in 060e0af6750fc4c83f83041fd6b99fdef6420b2b.

src/applications/gqrx/receiver.cpp
vladisslav2011: Fixed in 060e0af6750fc4c83f83041fd6b99fdef6420b2b.

argilo:
The underlying bug leading to #1003 (and #757) was already fixed in GNU Radio 3.8+: https://github.com/gnuradio/gnuradio/pull/3184

As a result, I think any changes to `receiver::set_input_device` should be confined to an `#if GNURADIO_VERSION < 0x030800` block.

argilo:
Or perhaps `#if GNURADIO_VERSION < 0x030802`, since Ubuntu 20.04 shipped with 3.8.1.0-rc1, a release-candidate version which did not include the fix.

vladisslav2011:
Thanks for suggestion.
I've added GNU Radio version check and provided better fix for white waterfall/no sound after playing back bad IQ file.

argilo:
> and provided better fix for white waterfall/no sound after playing back bad IQ file

This change seems unrelated to the first one. Could you please split it off into its own pull request so that it can be reviewed separately?

src/applications/gqrx/receiver.cpp
```diff
@@ -180,14 +180,29 @@ void receiver::stop()
...
-    qDebug() << "  old:" << input_devstr.c_str();
-    qDebug() << "  new:" << device.c_str();
-
     std::string error = "";
 
     if (device.empty())
         return;
+    if (input_devstr.compare(device) == 0)
+    {
+#ifndef QT_NO_DEBUG_OUTPUT
```
argilo, line 189: Why is `#ifndef QT_NO_DEBUG_OUTPUT` needed? Doesn't setting that disable `qDebug()` automatically?

vladisslav2011:
>Could you please split it off into its own pull request so that it can be reviewed separately?

OK. I've opened 2 new issues #1018 and #1019, and corresponding pull requests https://github.com/gqrx-sdr/gqrx/pull/1020 and https://github.com/gqrx-sdr/gqrx/pull/1021

src/applications/gqrx/receiver.cpp
vladisslav2011: I don't know where did I took this from. No useful information in case, the input device did not change. So, reverted to previous state and added a check for changed device string as it was done before, but removed later for some unknown purpose.
Why should we always reopen a device even if it did not change? If we have to work around some buggy device, then it is better to check for it's specific device string and reopen if it matches.
Moved this to https://github.com/gqrx-sdr/gqrx/pull/1021.

argilo:
Thanks for splitting those out!

At a glance the change looks good. I'll merge once I've verified that things work well with an older GNU Radio version.

argilo (approved):
Tested on Ubuntu 18.04, and this works well. I'll merge once CI passes.

Issue #994, opened 2021-10-25
S-meter oscillation


LongnoseRob:
As mentioned in the matrix channel, I observe regularrly, that after 10h or more of continious operation the S-meter indication seems to start pulsing/oscilalting.
Switching the decoder to another mode and back solves this issue (NFM->AM-->NFM).
Audio-decoding does not seem to be affected.
Running gqrx 2.14.5 built from sources.

Will try to get some video or still-pictures of this happening in the next days.

LongnoseRob:
The problem also occurs when running a file-source and using big FFT numbers and high refresh-rates.

argilo:
Closing, since #1006 should fix this.

PR #1006, opened 2021-11-29
Fix S-meter oscillation

argilo:
@LongnoseRob noticed in #994 that the S-meter sometimes begins oscillating. This can easily be reproduced by using a large buffer length with an RTL-SDR by setting the device string to `rtl=0,buflen=1048576`.

To fix this I've changed the rx_meter block to operate the same way the rx_fft block does: incoming samples are stored in a circular buffer (which can accommodate bursts up incoming samples up to one second long), and the same number of samples is always averaged. Clock measurements determine which batch of samples is averaged.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -3,6 +3,7 @@
 
        NEW: Start/stop RDS and fetch Program Identification via remote control.
        NEW: Added IQ balancing to MacOS release.
+     FIXED: Erratic behaviour of S-meter.
      FIXED: Compilation error due to missing include.
   IMPROVED: Reordered the receiver mode drop-down list.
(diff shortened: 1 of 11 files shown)


LongnoseRob:
@argilo Thank you very much for this fix.
I already built it on my sdr-system and will give it a few days of test time.
I think by the weekend we should have some feedback.

argilo:
@LongnoseRob Great, thanks!

I forgot to mention that this pull request includes some cleanup of the rx_meter block as well:

1. I've switched the detector type from RMS to average. Since it's measuring the signal power (and not amplitude) I don't think RMS is physically meaningful.
2. Removed the other detector types, which were never used.
3. Removed the `get_level` output, which was never used.

PR #1008, opened 2021-12-01
Iq tool fixes

### LINKED ISSUE #881: IQ recording informats other than F32
IQ recorder in GQRX could get multiple formats in which data would be saved other than F32 , for example  int16 and int8.
F32 is quite challenging on bit older systems or regular laptops at higher samplerates like 3 or 6MSPS causing sampledrops etc.

### LINKED ISSUE #977: Loading a second raw file causes a crash
When I do not have an SDR attached (press RETURN three tames to blow past select radio dialogs) and load a raw file, I load a raw file, listen to it then try to load any other file (or same file) the program crashes.
The IQ recording and play tool seems to not allow you to select a different file.  
My guess is that the .raw file never gets closed.

vladisslav2011:
Some small IQ tool improvements.
Things done:
- Buffered file writer. Makes it possible to record 20Msps and more to modern laggy HDDs without increasing DSP delay and skipping.
- Fix #977
- Add more formats (std::complex<short>,std::complex<char>). May close #881.
- GUI improvements. Block device selection and IQ tool controls while playing back/recording IQ file. Helps to prevent accidental file corruption/crash.
To be done later:
- Buffered player. May help with skipping/lag while reading IQ file from slow device (under IO load or just heavy fragmented)
- Remote control commands to start/stop IQ recording. Looks easy. Should I do it before switching to another issue?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ math(EXPR GNURADIO_BCD_VERSION
 add_definitions(-DGNURADIO_VERSION=${GNURADIO_BCD_VERSION})
 
 if(Gnuradio_VERSION VERSION_LESS "3.8")
-    find_package(Boost COMPONENTS system REQUIRED)
+    find_package(Boost COMPONENTS system thread REQUIRED)
 endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,13 @@ add_definitions(-D${BUILDTYPE})
 
 # We have some custom .cmake scripts not in the official distribution.
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules)
+set(CMAKE_THREAD_LIBS_INIT "-lpthread")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+set(CMAKE_HAVE_THREADS_LIBRARY 1)
+set(CMAKE_USE_WIN32_THREADS_INIT 0)
+set(CMAKE_USE_PTHREADS_INIT 1)
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+
 
 # Add valgrind build options if necessary
 if(${CMAKE_BUILD_TYPE} MATCHES "Valgrind")
@@ -145,7 +152,7 @@ math(EXPR GNURADIO_BCD_VERSION
 add_definitions(-DGNURADIO_VERSION=${GNURADIO_BCD_VERSION})
 
 if(Gnuradio_VERSION VERSION_LESS "3.8")
-    find_package(Boost COMPONENTS system thread REQUIRED)
+    find_package(Boost COMPONENTS system REQUIRED)
 endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD")
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -32,7 +32,7 @@ std::string receiver::get_random_file(void)
     static std::string path;
     if (path.empty())
     {
-        path = "/dev/urandom";
+        path = "/dev/zero";
         QFileInfo checkFile(QString::fromStdString(path));
         if (!checkFile.exists())
         {
--- a/src/applications/gqrx/file_resources.cpp
+++ b/src/applications/gqrx/file_resources.cpp
@@ -32,7 +32,7 @@ std::string receiver::get_random_file(void)
     static std::string path;
     if (path.empty())
     {
-        path = "/dev/zero";
+        path = "/dev/urandom";
         QFileInfo checkFile(QString::fromStdString(path));
         if (!checkFile.exists())
         {
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1605,6 +1605,8 @@ void MainWindow::startIqRecording(const QString& recdir, int bytes_per_sample, i
             toString("%1/gqrx_yyyyMMdd_hhmmss_%2_%3_%4.'raw'")
             .arg(recdir).arg(freq).arg(sr/dec).arg(suffix);
 
+    ui->actionIoConfig->setDisabled(true);
+    ui->actionLoadSettings->setDisabled(true);
     // start recorder; fails if recording already in progress
     if (rx->start_iq_recording(lastRec.toStdString(), bytes_per_sample, buffers_max))
     {
@@ -1636,6 +1638,8 @@ void MainWindow::stopIqRecording()
         ui->statusBar->showMessage(tr("Error stopping I/Q recoder"));
     else
         ui->statusBar->showMessage(tr("I/Q data recoding stopped"), 5000);
+    ui->actionIoConfig->setDisabled(false);
+    ui->actionLoadSettings->setDisabled(false);
 }
 
 void MainWindow::startIqPlayback(const QString& filename, float samprate, double center_freq, int bytes_per_sample)
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -281,9 +281,9 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     connect(&DXCSpots::Get(), SIGNAL(dxcSpotsUpdated()), this, SLOT(updateClusterSpots()));
 
     // I/Q playback
-    connect(iq_tool, SIGNAL(startRecording(QString)), this, SLOT(startIqRecording(QString)));
+    connect(iq_tool, SIGNAL(startRecording(QString,int)), this, SLOT(startIqRecording(QString,int)));
     connect(iq_tool, SIGNAL(stopRecording()), this, SLOT(stopIqRecording()));
-    connect(iq_tool, SIGNAL(startPlayback(QString,float)), this, SLOT(startIqPlayback(QString,float)));
+    connect(iq_tool, SIGNAL(startPlayback(QString,float,double,int)), this, SLOT(startIqPlayback(QString,float,double,int)));
     connect(iq_tool, SIGNAL(stopPlayback()), this, SLOT(stopIqPlayback()));
     connect(iq_tool, SIGNAL(seek(qint64)), this,SLOT(seekIqFile(qint64)));
 
@@ -1544,7 +1544,7 @@ void MainWindow::stopAudioStreaming()
 }
 
 /** Start I/Q recording. */
-void MainWindow::startIqRecording(const QString& recdir)
+void MainWindow::startIqRecording(const QString& recdir, int bytes_per_sample)
 {
     qDebug() << __func__;
     // generate file name using date, time, rf freq in kHz and BW in Hz
@@ -1552,12 +1552,25 @@ void MainWindow::startIqRecording(const QString& recdir)
     auto freq = (qint64)(rx->get_rf_freq());
     auto sr = (qint64)(rx->get_input_rate());
     auto dec = (quint32)(rx->get_input_decim());
+    QString suffix = "fc";
+    switch(bytes_per_sample)
+    {
+    case 2:
+        suffix="8";
+    break;
+    case 4:
+        suffix="16";
+    break;
+    default:
+        bytes_per_sample=8;
+        suffix="fc";
+    }
     auto lastRec = QDateTime::currentDateTimeUtc().
-            toString("%1/gqrx_yyyyMMdd_hhmmss_%2_%3_fc.'raw'")
-            .arg(recdir).arg(freq).arg(sr/dec);
+            toString("%1/gqrx_yyyyMMdd_hhmmss_%2_%3_%4.'raw'")
+            .arg(recdir).arg(freq).arg(sr/dec).arg(suffix);
 
     // start recorder; fails if recording already in progress
-    if (rx->start_iq_recording(lastRec.toStdString()))
+    if (rx->start_iq_recording(lastRec.toStdString(), bytes_per_sample))
     {
         // reset action status
         ui->statusBar->showMessage(tr("Error starting I/Q recoder"));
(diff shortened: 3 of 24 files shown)


CMakeLists.txt
```diff
@@ -145,7 +145,7 @@ math(EXPR GNURADIO_BCD_VERSION
 add_definitions(-DGNURADIO_VERSION=${GNURADIO_BCD_VERSION})
 
 if(Gnuradio_VERSION VERSION_LESS "3.8")
-    find_package(Boost COMPONENTS system REQUIRED)
+    find_package(Boost COMPONENTS system thread REQUIRED)
```
argilo, line 148: Could you please use `std::thread` instead? I'm trying to remove all usage of Boost from Gqrx.

src/applications/gqrx/file_resources.cpp
```diff
@@ -32,7 +32,7 @@ std::string receiver::get_random_file(void)
     static std::string path;
     if (path.empty())
     {
-        path = "/dev/urandom";
+        path = "/dev/zero";
```
argilo, line 35: The name of this method should be changed to reflect its new purpose. Also, line 45 should be changed to fill the file with zeroes.

src/interfaces/file_sink.h
```diff
@@ -0,0 +1,120 @@
...
+      size_t get_written();
+      void set_buffers_max(int buffers_max);
+
+      int work(int noutput_items,
+               gr_vector_const_void_star &input_items,
+               gr_vector_void_star &output_items);
+};
+
+
+#endif
```
argilo, line 120: Please end files with a newline.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -197,9 +197,9 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent)
 
     /* connect signals and slots */
     connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), this, SLOT(setNewFrequency(qint64)));
-    connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), remote, SLOT(setNewFrequency(qint64)));
-    connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), uiDockAudio, SLOT(setRxFrequency(qint64)));
-    connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), uiDockRxOpt, SLOT(setRxFreq(qint64)));
+    //connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), remote, SLOT(setNewFrequency(qint64)));
+    //connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), uiDockAudio, SLOT(setRxFrequency(qint64)));
+    //connect(ui->freqCtrl, SIGNAL(newFrequency(qint64)), uiDockRxOpt, SLOT(setRxFreq(qint64)));
```
argilo, line 202: Please delete instead of commenting out code.

src/dsp/format_converter.h
```diff
@@ -0,0 +1,144 @@
...
+#include <volk/volk.h>
+#include <iostream>
+
+namespace dispatcher
+{
+    template <typename> struct tag{};
+}
+
+    /*!
+     * \brief Convert stream of one format to a stream of another format.
```
argilo, line 37: Can't you just use GNU Radio's built-in type conversion blocks for this?

src/interfaces/file_sink.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include <gnuradio/blocks/api.h>
+#include <gnuradio/sync_block.h>
+#include <boost/lockfree/queue.hpp>
+
+/*!
+    * \brief Write stream to file without blocking.
+    * \ingroup file_operators_blk
+    */
+class BLOCKS_API file_sink : virtual public gr::sync_block
```
argilo, line 35: I think it would be best to contribute this improvement upstream to GNU Radio rather than having Gqrx maintain its own copy of the File Sink block.

argilo:
@vladisslav2011 I'd suggest proposing changes in individual pull requests, instead of lumping many changes into one. This makes code review much easier, and avoids the situation where only some of the changes can be accepted.

src/interfaces/file_sink.h
argilo: Also, I'm not convinced that adding another layer of buffering on top of the standard library & operating system's buffering is a good idea, especially since it requires adding threads and incurring the risk of concurrency bugs.

CMakeLists.txt
vladisslav2011: Thank you for suggestion.
Done in 2db43829817932406037598e8a3dadf654c3346f.
boost::thread dependency removed in fce4a22876f7a7b6efee38053ad0bfb2d181fcc1.

src/applications/gqrx/file_resources.cpp
vladisslav2011: Thanks.
Reverted this change in a8e9e898695b64d852a9ef4a4537ffd7c6473bc4.

src/applications/gqrx/mainwindow.cpp
vladisslav2011: Sorry for forgetting to do it.
Removed in 0c09ce82fa59ba9e38773e97a687200c2ce6534b.

src/dsp/format_converter.h
vladisslav2011: I've did it before and switched to my own implementation later because:
1. GNU Radio's built-in type conversion blocks are not using libvolk (for some unknown reason), so performance is not great.
2. My GNU Radio installation (distribution supplied, yes, I'm lazy to compile it myself) lacks interleaved_char_to_complex and complex_to_interleaved_char blocks.
3. Using GNU Radio blocks requires to chain a multiply_const block for scaling. Libvolk functions do both conversion and scaling at one call.

src/interfaces/file_sink.h
vladisslav2011: I'll try to contribute this improvement upstream to GNU Radio (after some rework, adding and option switch buffering on/off, an option to allocate static buffer and maybe non-blocking file closing), but I don't think that it will be accepted fast enough... or accepted at all. But I want to use gqrx now and have perfect IQ recordings without skipping.
The standard library does not implement even double buffering and always end up with system write call, that may sometimes block for multiple seconds (on my modern SMR HDDs). Adding some application-controlled buffering always solves this problem at cost of (temporarily) increased memory usage. Writing data in large chunks improves performance and lowers CPU usage.
I have plans to implement buffered file_sourse too as playing back 10...20Msps IQ files, recorded from Hackrf, is problematic with current GNU Radio file_source.

vladisslav2011:
Thank you for review. (It looks, like I'll have to maintain my own fork like LouDou LoL).
I can split this PR into at least 6 (+ buffered file_source later) parts:
1. Writing different IQ file formats using stock GNU Radio file_sink and related GUI changes (QComboBox to select sample format)
2. Parsing center frequency from file name
3. Playing different IQ file formats and parsing sample format from file name
4. Fixing issues with center frequency changes while playing IQ file
5. GUI changes to improve IQ tool usability (disabling unusable/bad behaving controls)
6. Buffered file writer and related GUI changes. May be made optional with cmake defines or with runtime switch to choose between GNU Radio file_sink and buffered file_sink. May be replace with some better buffering implementation when somebody makes it.
7. TBD Buffered file reader that would not require GUI changes (take buffer count from recording buffering configuration and calculate buffer size from sample rate to allocate 1s buffers)
It may be, I forgot something, that may be split out without breaking things and introducing more bugs. Suggest your variants, please.

src/dsp/format_converter.h
argilo: As with the File Sink, I don't think Grqx is the right place to do this. I don't want the added maintenance cost of maintaining a parallel implementation.

> GNU Radio's built-in type conversion blocks are not using libvolk

I'm sure GNU Radio would appreciate a patch to fix this.

> My GNU Radio installation lacks interleaved_char_to_complex and complex_to_interleaved_char blocks.

What version are you using, and what version were these blocks added in? If needed, we could use a cascade of blocks in Gqrx to support older versions of GNU Radio.

> Using GNU Radio blocks requires to chain a multiply_const block for scaling.

Some of the type conversion blocks have built-in scaling. If others lack it, then I'd suggest contributing a patch to GNU Radio.

src/interfaces/file_sink.h
argilo: > system write call, that may sometimes block for multiple seconds

This sounds like a problem with your operating system. I'd expect write-back caching to solve this problem at the OS layer. Perhaps you need to adjust the cache size?

argilo:
> I can split this PR into at least 6 (+ buffered file_source later) parts

I think that's a sensible breakdown. But as mentioned, I'm not yet convinced that 6 and 7 are necessary; even if they are they should go into GNU Radio instead.

> Parsing center frequency from file name
> Playing different IQ file formats and parsing sample format from file name

In the longer term, I think the path forward is SigMF (https://github.com/gnuradio/SigMF). It would be great if I/Q recording and playback in Gqrx supported SigMF. Any improvements we make now should leave room for SigMF to be added.

src/dsp/format_converter.h
vladisslav2011: > I'm sure GNU Radio would appreciate a patch to fix this.
This is already done in GNU Radio 3.9.1 but setting minimum required GNU Radio version to 3.9.1 would break compilation on many major distributions.

> If needed, we could use a cascade of blocks in Gqrx to support older versions of GNU Radio.
I've written before, that I've did it this way first. But then I have found, that complex_to_interleaved_char block is not shipped by my base distribution (Ubuntu 18.04LTS). So I have to either build recent GNU Radio from source on all machines, where I use gqrx (different distributions, different architectures), or pull the block from GNU Radio source code, or replace all conversion blocks with small template class, that may be easily extended to support more formats later (it is possible as we do not have to support python binding). 
Cascade blocks are bad from performance point of view: more unneeded buffers, more threads.
Sometimes all major distributions will adopt GNU Radio 3.9.1 and we'll be able to get rid of this small header.
As an option, I can convert added blocks to implement modern GNU Radio interface (inherit or just typedef would be enough?) and put the implementation into "compat" folder, that would not get compiled/included if we have found GNU Radio 3.9.1. This would reduce maintenance time later. But this would add more #ifdefs and grow CMakelists.txt...

src/interfaces/file_sink.h
vladisslav2011: All machines, all operating systems (different versions of Linux) suffer from it more or less. And windows, of course, suffer from it too in a much worse way. I do not own a mac, so maybe mac lives in some different world where there is more RAM, than all running processes (chrome e.g.) could consume and where is always enough free RAM space for disk buffers/cache.

Some fine tuning may help, but, as write system call is blocking by design and can not be made non blocking (like send system call), there is absolutely no guarantee, that it will finish in predictable time. The only way to get predictable low blocking time is application level buffering with large enough buffers. Buffers should be preallocated, as sbrk system call may block or even fail and maybe marked with madvice as being used in near future. 

After turning off swaps (took about 5 minutes) things became a little bit better. Much less skipping in normal conditions.

My test results (hackrf, 20Msps) are the following:
- Current git master: gqrx starts dropping buffers after maximum 10 seconds of recording. The resulting file is unplayable. Constant crackling after 10 seconds.
- iq_tool_fixes: gqrx finishes recording gracefully with error message after 1 minute. The resulting file does not have lost samples. Recording with 16 bits per component (32 bits per sample) proceeds till the end of disk. The resulting file is playable.

Is it better to have a short, but correct file and error message showing, that the target device is too slow or a large file full of garbage?

vladisslav2011:
SigMF is fine. I have looked at it's specs.
Implementing metadata readout in a player is easy.
Implementing writing metadata file in recorder should be easy too.
Reading/writing a container is a bit harder, but easy enough in sequental way. Seeking tar (maybe compressed tar) is harder, but still possible.

src/dsp/format_converter.h
argilo: My suggestion would be to have two different versions of the flow graph, depending on the GNU Radio version. Gqrx already does that in a few places, for example:

https://github.com/gqrx-sdr/gqrx/blob/75d108f7e9b0bb69bb1847173d957b8e41a0cc8f/src/dsp/rx_rds.cpp#L85-L103

I would suggest to use a cascade of blocks for type conversion in older versions, and the new IChar blocks (starting in 3.8?).

As a side note, supporting GNU Radio 3.7 is already becoming a burden, so I'd like to drop support for it soon. Users on old distributions can always use the AppImage releases.

src/interfaces/file_sink.h
argilo: > Is it better to have a short, but correct file and error message showing, that the target device is too slow or a large file full of garbage?

I can believe that some users might prefer the recording to continue, but with dropped samples.

In any case, I still believe that Gqrx is not the right place for this. If you believe this improvement to the File Sink is valuable, please contribute it to GNU Radio. Then Gqrx and all other applications can benefit from it. I do not want to maintain copies of GNU Radio's blocks inside Gqrx.

argilo: An alternate suggestion: try calling `set_min_output_buffer` on the block preceding the File Sink.

src/applications/gqrx/file_resources.cpp
LouDou: This has been done in #1001

src/interfaces/file_sink.h
vladisslav2011: src->set_min_output_buffer was the first thing, I have tried. If it is set to ~2 seconds, it still gives choppy sound, interrupted every second. If it is set to more, than 2 seconds ... gqrx crashes with std::bad_alloc.
I've implement simple circular buffer based on std::vector (to prevent memory leaks) and got perfect results. No audio interruptions when the buffer is set to 6-7 seconds and acceptable memory usage. Seek performance is sometimes bad, as well as startup times.
+ I've added current file position readout to keep the slider in sync with actual file position. This prevents the slider runaway when the DSP is stopped, but file playback is not. This closes #1017. I do not see a way to make it work correctly using stock GNU Radio file_source...
As this PR is converted to draft (maybe just close it?), we can continue conversation here: https://github.com/gqrx-sdr/gqrx/pull/1016

vladisslav2011:
Outdated. Closing.

PR #1009, opened 2021-12-02
Iq tool center freq fix

vladisslav2011:
1. Parse the center frequency from IQ file name
2. Save center frequency and offset before starting IQ file playback
3. Restore center frequency and offset after finishing IQ file playback

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -283,7 +283,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     // I/Q playback
     connect(iq_tool, SIGNAL(startRecording(QString)), this, SLOT(startIqRecording(QString)));
     connect(iq_tool, SIGNAL(stopRecording()), this, SLOT(stopIqRecording()));
-    connect(iq_tool, SIGNAL(startPlayback(QString,float)), this, SLOT(startIqPlayback(QString,float)));
+    connect(iq_tool, SIGNAL(startPlayback(QString,float,qint64)), this, SLOT(startIqPlayback(QString,float,qint64)));
     connect(iq_tool, SIGNAL(stopPlayback()), this, SLOT(stopIqPlayback()));
     connect(iq_tool, SIGNAL(seek(qint64)), this,SLOT(seekIqFile(qint64)));
 
@@ -1588,7 +1588,7 @@ void MainWindow::stopIqRecording()
         ui->statusBar->showMessage(tr("I/Q data recoding stopped"), 5000);
 }
 
-void MainWindow::startIqPlayback(const QString& filename, float samprate)
+void MainWindow::startIqPlayback(const QString& filename, float samprate, qint64 center_freq)
 {
     if (ui->actionDSP->isChecked())
     {
@@ -1599,13 +1599,16 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
     storeSession();
 
     auto sri = (int)samprate;
+    auto cf  = center_freq;
+    double current_offset = rx->get_filter_offset();
     QString escapedFilename = receiver::escape_filename(filename.toStdString()).c_str();
-    auto devstr = QString("file=%1,rate=%2,throttle=true,repeat=false")
-            .arg(escapedFilename).arg(sri);
+    auto devstr = QString("file=%1,rate=%2,freq=%3,throttle=true,repeat=false")
+            .arg(escapedFilename).arg(sri).arg(cf);
 
     qDebug() << __func__ << ":" << devstr;
 
     rx->set_input_device(devstr.toStdString());
+    updateHWFrequencyRange(false);
 
     // sample rate
     auto actual_rate = rx->set_input_rate(samprate);
@@ -1616,6 +1619,11 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
     uiDockRxOpt->setFilterOffsetRange((qint64)(actual_rate));
     ui->plotter->setSampleRate(actual_rate);
     ui->plotter->setSpanFreq((quint32)actual_rate);
+    if (std::abs(current_offset) > actual_rate / 2)
+        on_plotter_newDemodFreq(center_freq, 0);
+    else
+        on_plotter_newDemodFreq(center_freq + current_offset, current_offset);
+
     remote->setBandwidth(actual_rate);
 
     // FIXME: would be nice with good/bad status
@@ -1659,6 +1667,14 @@ void MainWindow::stopIqPlayback()
 
     // restore frequency, gain, etc...
     uiDockInputCtl->readSettings(m_settings);
+    bool centerOK = false;
+    bool offsetOK = false;
+    qint64 oldCenter = m_settings->value("input/frequency", 0).toLongLong(&centerOK);
+    qint64 oldOffset = m_settings->value("receiver/offset", 0).toLongLong(&offsetOK);
+    if (centerOK && offsetOK)
+    {
+        on_plotter_newDemodFreq(oldCenter, oldOffset);
+    }
 
     if (ui->actionDSP->isChecked())
     {
--- a/src/qtgui/iq_tool.cpp
+++ b/src/qtgui/iq_tool.cpp
@@ -47,6 +47,7 @@ CIqTool::CIqTool(QWidget *parent) :
     bytes_per_sample = 8;
     sample_rate = 192000;
     rec_len = 0;
+    center_freq = 1e8;
 
     //ui->recDirEdit->setText(QDir::currentPath());
 
@@ -83,6 +84,7 @@ void CIqTool::setSampleRate(qint64 sr)
     }
 }
 
+
 /*! \brief Slot activated when the user selects a file. */
 void CIqTool::on_listWidget_currentTextChanged(const QString &currentText)
 {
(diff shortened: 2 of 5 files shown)


argilo:
Thanks for splitting this off into its own pull request. It's much easier to review this way. I'll have a look at this tonight.

src/qtgui/iq_tool.cpp
```diff
@@ -91,7 +91,8 @@ void CIqTool::on_listWidget_currentTextChanged(const QString &currentText)
     QFileInfo info(*recdir, current_file);
 
     // Get duration of selected recording and update label
-    sample_rate = sampleRateFromFileName(currentText);
+    //sample_rate = sampleRateFromFileName(currentText);
```
argilo, line 94: This can be removed.

src/qtgui/iq_tool.cpp
```diff
@@ -91,7 +91,8 @@ void CIqTool::on_listWidget_currentTextChanged(const QString &currentText)
     QFileInfo info(*recdir, current_file);
 
     // Get duration of selected recording and update label
```
argilo, line 93: This comment looks like it should be above line 96 instead.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1658,6 +1663,7 @@ void MainWindow::stopIqPlayback()
 
     // restore frequency, gain, etc...
     uiDockInputCtl->readSettings(m_settings);
+    on_plotter_newDemodFreq(backupFreq,backupOffset);
```
argilo, line 1666:
```suggestion
    on_plotter_newDemodFreq(backupFreq, backupOffset);
```

src/qtgui/iq_tool.cpp
```diff
@@ -355,22 +356,25 @@ void CIqTool::refreshTimeWidgets(void)
...
+    bool sr_ok;
+    qint64 sr=1e6;
+    bool ofs_ok;
+    double ofs=1e8;
 
     QStringList list = filename.split('_');
 
     if (list.size() < 5)
-        return sample_rate;
+        return ;
```
argilo, line 370:
```suggestion
        return;
```

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1597,10 +1597,13 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
     }
 
     storeSession();
+    backupFreq = ui->freqCtrl->getFrequency();
+    backupOffset = (qint64) rx->get_filter_offset();
 
     auto sri = (int)samprate;
-    auto devstr = QString("file='%1',rate=%2,throttle=true,repeat=false")
-            .arg(filename).arg(sri);
+    auto cf  = (long long) center_freq;
```
argilo, line 1604:
```suggestion
    auto cf = (long long) center_freq;
```

argilo:
So far the code looks good, I just suggested a few formatting improvements.

I'll test this out tonight.

vladisslav2011:
Thank you for a review.
Sorry for sending a dirty PR. I forgot to check code formatting and remove unused code. Again...
Should I fix code formatting by editing commits followed by push --force, or by adding a new commit?

argilo:
> Should I fix code formatting by editing commits followed by push --force, or by adding a new commit?

I'd suggest to force-push a new commit. That will keep the commit history clean once this is merged.

vladisslav2011:
Done.

argilo:
I noticed a bug. When switching from receive (in this example, with an RTL-SDR) to I/Q playback, the plotter frequency axis updates, but the frequency controls do not. Here's what I see when I start playback of an I/Q file with a center frequency around 390 MHz:

gqrx-iq-bug-1

The frequency control still shows 929 MHz, and the indicated offset is 0 Hz. If I click to retune, then the correct frequencies are displayed.

vladisslav2011:
Thank you for reporting this bug.
Fixed in d7b7c94e258b92567353a61936ac9207404bc1f6.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1597,11 +1597,14 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
     }
 
     storeSession();
+    backupFreq = ui->freqCtrl->getFrequency();
+    backupOffset = (qint64) rx->get_filter_offset();
```
argilo, line 1601: Why is it necessary to add these variables? Doesn't `storeSession()` already save them in the `input/frequency` and `receiver/offset` settings?

argilo:
Another minor issue I noticed is that when the I/Q file's center frequency falls outside the hardware limits of the currently selected device, then the frequency is not set correctly. (It's constrained to be within the hardware device's limits.) It looks like this happens because the hardware frequency range is not updated. This change appears to fix the problem:
```diff
diff --git a/src/applications/gqrx/mainwindow.cpp b/src/applications/gqrx/mainwindow.cpp
index 34a7d5eb6b..a78237d736 100644
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1609,6 +1609,7 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate, double
     qDebug() << __func__ << ":" << devstr;
 
     rx->set_input_device(devstr.toStdString());
+    updateHWFrequencyRange(false);
 
     // sample rate
     auto actual_rate = rx->set_input_rate(samprate);
```

argilo:
Thanks again for the pull request, and sorry for the long delay in reviewing this. I've suggested a couple further improvements. When you make those changes, could you also squash your commits into a single one? That will help keep the commit history clean.

src/applications/gqrx/mainwindow.cpp
vladisslav2011: Using the "Device dialog" or the "Save settings" button may overwrite settings during IQ playback as both are calling `storeSession()`. But using any of these buttons breaks playback too...
OK. Ill switch to reading center/offset from stored settings and add disabling "Save settings" button during playback to https://github.com/gqrx-sdr/gqrx/pull/1011.

vladisslav2011:
>This change appears to fix the problem

Thanks. Fix confirmed. Included into 489af53.

>When you make those changes, could you also squash your commits into a single one?

No problem. Done.
I'll rebase subsequent PRs on top of this and squash commits into one too.

src/applications/gqrx/mainwindow.cpp
argilo: The suggestion to use `storeSession()` was just for consistency with the current design. I agree that the current design is not ideal, so maybe we should think about changing it (in a future PR, perhaps).

argilo:
Another bug:

If `CIqTool::parseFileName` fails to parse the filename for any reason (for instance, if the filename is `foo.raw`), it leaves `center_freq` uninitialized, which leads to undefined behaviour. Its value needs to be set, similarly to how `sample_rate` is set:

https://github.com/gqrx-sdr/gqrx/blob/49de5f35ec6bd5b835b953db030ff5acb845dfeb/src/qtgui/iq_tool.cpp#L48

https://github.com/gqrx-sdr/gqrx/blob/49de5f35ec6bd5b835b953db030ff5acb845dfeb/src/applications/gqrx/mainwindow.cpp#L617

vladisslav2011:
Thanks for more testing.
Hmmm. If we want to support arbitrary-named files in the IQ Player, then it's up to user to supply correct sample rate, center frequency (and sample format). So, we have to provide some GUI elements.
Another issue was spotted while testing this PR: Incorrect frequency calculation. `on_plotter_newDemodFreq` wants current tuned frequency and offset, not center frequency and offset.
I've fixed both and force pushed a new version.

vladisslav2011:
I forgot to remove debug prints.
Pushed a clean version.

argilo:
> If we want to support arbitrary-named files in the IQ Player, then it's up to user to supply correct sample rate, center frequency (and sample format). So, we have to provide some GUI elements.

That might be a useful feature, but for now I think it's fine to default to the current sample rate and center frequency, and assume complex floating point samples.

vladisslav2011:
OK. I'll split this feature into separate PR. Wait some minutes.

vladisslav2011:
Ne GUI controls are moved to https://github.com/gqrx-sdr/gqrx/pull/1026

argilo:
Another bug:

1. Open the I/Q tool
2. Click on a file, but don't play it yet
3. Close the I/Q tool
4. Open the I/Q tool again
5. Press the Play button

In this case, the center frequency will not be adjusted.

argilo:
Also, in that same case the wrong file duration will be displayed if the sample rate of the file differs from the sample rate of the currently selected device. I guess both bugs are triggered by the additional code that was added to `MainWindow::on_actionIqTool_triggered`.

vladisslav2011:
> In this case, the center frequency will not be adjusted.

Confirmed. It looks, like getting default sample rate/center frequency from current selected device does not work as expected.

+ I have found another bug. Wrong center frequency is set when the current offset does not fit into bandwidth of selected IQ file:
1. Set sampling rate to ~2 Msps and record IQ file.
2. Select wideband device (LimeSDR/HackRF), set sampling rate to some large value. 16 Msps for example.
3. Set offset to +7 MHz for example
4. Select IQ file, recorded at step 1 and hit play.

I'll fix both bugs and reopen this PR.
I think, we should not declare support of playing back arbitrary named IQ files until https://github.com/gqrx-sdr/gqrx/pull/1026 gets merged.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1588,7 +1588,7 @@ void MainWindow::stopIqRecording()
         ui->statusBar->showMessage(tr("I/Q data recoding stopped"), 5000);
 }
 
-void MainWindow::startIqPlayback(const QString& filename, float samprate)
+void MainWindow::startIqPlayback(const QString& filename, float samprate, double center_freq)
```
argilo, line 1591: `qint64` is used elsewhere in Gqrx for frequencies, so I think we should use that instead of `double` or `long long` (below).

vladisslav2011: Done.

argilo:
The code looks good now, and the change appears to be working well. I'll merge once CI passes.

Issue #951, opened 2021-03-25
set_rf_freq() does not check hw tuning result


willcode:
When the hardware can not tune exactly, that fact is silently ignored.

Possibly solution: read back the actual hardware freq (this is a FIXME to this effect in the code). Use a second frequency translation to account for any (reasonable) difference. This will give the user the requested tuning, and will also preserve any offset entered in the receiver dialog.

vladisslav2011:
Do you mean something like this?
```
diff --git a/src/applications/gqrx/mainwindow.cpp b/src/applications/gqrx/mainwindow.cpp
index adcb9b6..ee3adbe 100644
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -863,10 +863,14 @@ void MainWindow::setNewFrequency(qint64 rx_freq)
     auto hw_freq = (double)(rx_freq - d_lnb_lo) - rx->get_filter_offset();
     auto center_freq = rx_freq - (qint64)rx->get_filter_offset();
 
-    d_hw_freq = (qint64)hw_freq;
-
     // set receiver frequency
     rx->set_rf_freq(hw_freq);
+    // read it back
+    hw_freq = rx->get_rf_freq();
+    rx_freq = (qint64) hw_freq + d_lnb_lo + rx->get_filter_offset();
+    center_freq = rx_freq - (qint64)rx->get_filter_offset();
+
+    d_hw_freq = (qint64)hw_freq;
 
     // update widgets
     ui->plotter->setCenterFreq(center_freq);
```

PR #1010, opened 2021-12-02
Iq tool center freq lock

vladisslav2011:
Prevent accidental change of hw freq while playing back IQ file.
Includes https://github.com/gqrx-sdr/gqrx/pull/1009


vladisslav2011:
Rebased onto updated iq_tool_center_freq_fix + some code style fixes.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1597,12 +1614,15 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate)
...
 
     auto sri = (int)samprate;
+    auto cf  = (long long) center_freq;
     QString escapedFilename = receiver::escape_filename(filename.toStdString()).c_str();
-    auto devstr = QString("file=%1,rate=%2,throttle=true,repeat=false")
-            .arg(escapedFilename).arg(sri);
-
+    auto devstr = QString("file=%1,rate=%2,freq=%3,throttle=true,repeat=false")
+            .arg(escapedFilename).arg(sri).arg(cf);
+optimize_c
```
argilo, line 1625: This line is causing a build failure.

vladisslav2011: I've fixed it.
Editing commits at github without verification is bad thing. I've done a wheel misclick while scrolling up to submit changes.

argilo:
There are quite a few code changes here, so it would be very helpful to have some explanation of what's going on.

Some questions:
* In what ways is it possible to accidentally change the hardware frequency?
* What are the consequences of changing the hardware frequency during I/Q playback?
* What changes were necessary to solve this?

vladisslav2011:
> * In what ways is it possible to accidentally change the hardware frequency?

Wheel over freqCtrl, middle button dragging of plotter horizontal axis, dockrxopt frequency input.
I do it all the time while using the IQ Tool.
It is possible to change hardware frequency with remote command 'F', but it is highly unlikely that anybody would use remote control while playing IQ file.

> * What are the consequences of changing the hardware frequency during I/Q playback?

Bookmarks become not valid, freqCtrl shows wrong frequency, plotter shows wrong frequency and so on.

> * What changes were necessary to solve this?

The main change is here: https://github.com/gqrx-sdr/gqrx/pull/1010/commits/ceaa48c2c421e33e9e18233ab52364bb19fd2c8a#diff-2fc3518522047c9d0c5d1f75d1de006499a71e9812311c7228c892a5250bad48R866-R885
Other changes were made to remove event flow loops and make all frequency changes and widgets updates go through `MainWindow::setNewFrequency`.

Thanks for a review.
I think, we can avoid calling `rx->set_rf_freq()` twice and close https://github.com/gqrx-sdr/gqrx/issues/951 with this PR too.
I'll do required changes, make some tests and reopen the PR soon.

argilo:
What if the filename does not have a frequency, or the frequency is incorrect? Should the user not be allowed to adjust the frequency in those cases?

argilo:
I guess the core of the problem is the behaviour of the main frequency control, which currently changes the hardware frequency rather than the channel filter offset. I sometimes find this annoying even when using hardware inputs, because changing the hardware frequency is disruptive. (It is not instantaneous, especially when using remote devices, it shifts the waterfall, and it invalidates FFT averages.)

I agree that it makes sense for the main frequency control to adjust the channel filter offset during I/Q playback. It might even make sense for that to be the default for hardware inputs. (SDR# works this way, for instance.)

I'll have to play around with this a bit more and think about what, if anything, should be done to allow the hardware frequency to be adjusted during I/Q playback.

argilo:
One minor issue: if an I/Q file is opened using the I/O Devices panel instead of the I/Q tool, then the current frequency continues to be used, even if it is well outside the viewable range:

Screenshot from 2021-12-31 17-31-58

Getting it back in range using the main frequency control is a bit painful. Clicking on the FFT gets it back in range more easily.

The frequency should probably be clamped to be within the bounds of the FFT.

vladisslav2011:
> What if the filename does not have a frequency, or the frequency is incorrect?

This case is addressed by this PR: https://github.com/gqrx-sdr/gqrx/pull/1026

> what, if anything, should be done to allow the hardware frequency to be adjusted during I/Q playback.

I think, the hardware frequency should be adjusted during IQ playback only in the same way, it had changed during recording of the corresponding file. This may be implemented by processing tags, generated by SigMF source in future.
Another case, when hardware frequency is changing during IQ recording is satellite signal recording with remote control to compensate the doppler shift. In that case the hardware frequency should not be adjusted during playback due to obvious reasons...

> The frequency should probably be clamped to be within the bounds of the FFT.

This can be done in `MainWindow::setNewFrequency`. Then it may be possible to remove bounds checking in the plotter. I'll try to implement this and update this PR.

vladisslav2011:
> The frequency should probably be clamped to be within the bounds of the FFT.

Done.
My tests show no regressions nor UX degradation.
Bounds checking in the plotter is done only to prevent moving the waterfall/spectrogram far away from visible area while zooming. Nothing to do here.

PR #1012, opened 2021-12-08
Escape filename in device specification string

argilo:
Replaces #1001.

Based on the work of @LouDou an @vladisslav2011 I've put together a cleaner version of a fix for both startup crashes on Windows and failure to play back I/Q files where the filename contains special characters.

Would you mind taking this branch for a spin @LouDou @vladisslav2011?


vladisslav2011:
I can confirm, that this fix works well even when a path contains backslash, comma and spaces.

argilo:
Great! Thanks for testing.

Issue #946, opened 2021-03-02
Feature: Multiple demodulators simultaneously


kord1e:
The ability to set up multiple demodulators simultaneously might be somewhat useful, for example, when one wants to listen to several air traffic channels at once.

KevWal:
A plus one from me, the ability to run 2 copies of a decoder like wsjtx, linked to two different audio streams out of Gqrx, to decode FT8 and WSPR for example for my balloon tracking.

LouDou:
I've started looking into this on a branch here: https://github.com/LouDou/gqrx/tree/multirx

It took about 4hrs work to de-tangle and re-work the "receiver" controller class, and I've proven that the DSP graph is capable of doing this - with each receiver outputting to its own audio stream (when using pulseaudio or portaudio).

None of the GUI work has been done yet, and I've had to disable all recording, RDS, AFSK features too - in fact, I don't even know the best way to represent the multiples of all the receiver settings... it's fundamentally quite intrusive changes with regards to using the application. Also, there's quite a bit of work to do around adding/removing receivers without completely breaking everything.

... but it's a start.

LouDou:
![Screenshot_20210922_224304](https://user-images.githubusercontent.com/289750/134426305-2745b803-ab40-4e20-8a13-bce9781ea393.png)

Well, lets demodulate 2 broadcast FM stations at once then :)

Still lots to do, but some basic control over two Receivers is possible here.

Geremia:
@kord1e This would be a great addition. The (now defunct, unmaintained?) [CubicSDR](https://github.com/cjcliffe/CubicSDR) can do this. From [its docs](https://cubicsdr.readthedocs.io/en/latest/application-window.html#adding-a-modem:~:text=To%20add%20additional%20modems%20hold%20the%20shift%20key%20while%20placing%3B%20the%20color%20will%20change%20to%20green%20to%20indicate%20a%20new%20modem%20will%20be%20created%20%28as%20opposed%20to%20moving%20an%20existing%20one%20which%20will%20appear%20in%20red%29%2E):

> To add additional modems hold the shift key while placing [i.e., "hovering and clicking on the waterfall"]; the color will change to green to indicate a new modem will be created (as opposed to moving an existing one which will appear in red).

PR #1015, opened 2021-12-14
Implement quick and dirty squelch-triggered wav recording

### LINKED ISSUE #109: SQL triggered auto-recording option
It would be nice to have an auto-record feature when SQL is triggered, so it would start recording on SQL trigger, stop on SQL, and save to a datetime tagged file. Useful for monitoring low-usage channels and recording just when someone is talking.

vladisslav2011:
Close #109.
And switch to pwr_squelch to get tags added for later proper
implementation.


vladisslav2011:
This should be done inside the receiver block after removing rx_gain# blocks and moving rx_agc to correct place (after the demodulator, not before).

vladisslav2011:
New, better implementation, made keeping in mind, that https://github.com/gqrx-sdr/gqrx/issues/946 would be next.
CI is failing as I have not updated dependencies... Still draft so far.

vladisslav2011:
New implementation is ready.
New rx agc implementation is required, so review after "New rx agc..." sequence, please.

Bad: requires libsndfile
Good: It may be possible to implement recording of different audio formats (ogg to reduce file sizes, RAW audio to be played through IQ tool, etc.) independent of GNU Radio version.

Bug reports and suggestions are welcome.

PR #1020, opened 2021-12-15
Quick fix for #1019

### LINKED ISSUE #1019: BUG: Changing IO devices stops the DSP
Regression.

Steps to reproduce:
1. Start gqrx, select some hardware device, start the DSP
2. Go back to I/O device dialog without stopping the DSP and choose some other device

Expected behavior:
The DSP restarts after switching devices. Waterfall continues to update with data from selected device, the sound should continue to output too.

Observer behavior:
The DSP does not restart.

Some old gqrx version did not suffer from this bug, so this is a regression...

vladisslav2011:
Fixes #1019.

Store previous DSP state in a variable to use it later as on_actionDSP_triggered changes ui->actionDSP state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,7 @@ The following people and organisations have contributed to gqrx:
 * Valentin Ochs
 * Vesa Solonen
 * Vincent Pelletier
+* Vladisslav P
 * Will Scales
 * Wolfgang Fritz, DK7OB
 * Youssef Touil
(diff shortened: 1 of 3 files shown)


argilo:
I see that `MainWindow::stopIqPlayback` has the same bug. Perhaps we should fix that here too.

argilo:
Actually, it looks like applying the same fix to `MainWindow::stopIqPlayback` would cause the DSP to always start when I/Q playback is stopped, which we don't want. Best to leave any fix to a separate pull request, I guess.

argilo (approved):
Looks good. I'll merge this once CI passes.

PR #1021, opened 2021-12-15
Fix white waterfall and no sound after playing back bad IQ file.

### LINKED ISSUE #1018: BUG: White waterfall and no sound after playing back certain IQ files
Attempting to play back some IQ files (rtl_sdr console tool recording in uint8_t format for example) leaves the DSP in incorrect state.

Steps to reproduce:
1. Prepare "bad" IQ file. dd if=/dev/urandom of=gqrx_20211213_164230_103000000_1800000_fc.raw bs=128M count=1
2. Start gqrx, choose hardware device, start the DSP, enable DC Remove in input control dock, check, that there is correct waterfall, spectrogram and audio output.
3. Open "bad" file in IQ player, hit play. There should be no sound and white waterfall.
3. Stop IQ playback.
4. Close IQ player and start DSP with hardware source.

Expected behavior:
After switching back to hardware device, the gqrx should show correct waterfall and output sound

Observed behavior:
After switching back to hardware device, the gqrx still shows white waterfall and no sound.
Switching from nbfm to wbfm and back may help with sound. Turning the DC remove off may help with white waterfall.

vladisslav2011:
Fix #1018.
Workaround gnuradio blocks state corruption after device change by recreating dc_corr
(fixes white waterfall and no sound) and resetting the demod (fixes corrupted sound).
It looks like sometimes demodulators are not happy with bad samles too.
We should recreate the graph, temporarily switching out current
demodulator to prevent bad behavior.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -233,6 +233,12 @@ void receiver::set_input_device(const std::string device)
         tb->connect(src, 0, iq_swap, 0);
     }
 
+    //reset DSP IIR filters to safe initial state from possible runaway after
+    // playing bad IQ file
+    auto last_demod = d_demod;
+    set_demod(RX_DEMOD_OFF, true);
+    set_demod(last_demod);
+
     if (d_running)
         tb->start();
 
@@ -1315,6 +1321,7 @@ void receiver::connect_all(rx_chain type)
 
     if (d_dc_cancel)
     {
+        dc_corr = make_dc_corr_cc(d_decim_rate, 1.0);
         tb->connect(b, 0, dc_corr, 0);
         b = dc_corr;
     }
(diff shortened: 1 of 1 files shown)


argilo:
> Do not reopen a device when it's string did not change as it speeds up switching filter bandwidths/sample rate.

That change seems unrelated. Could you put it in its own pull request so that it can be reviewed separately?

vladisslav2011:
Unrelated changes are split into https://github.com/gqrx-sdr/gqrx/pull/1024

argilo:
I agree that it's a good idea to reset the DSP blocks when the input device is changed. Triggering that by calling `set_demod` twice seems like a hack though, and I could believe it might lead to future bugs. Let me think about whether there's a better way to accomplish this.

vladisslav2011:
Closed in favor of https://github.com/gqrx-sdr/gqrx/pull/1036.

PR #1024, opened 2021-12-17
Do not reopen a device when it's string did not change as it speeds u…

vladisslav2011:
…p switching filter bandwidths/sample rate.
That was done before 8b4c2efd394734f120245739e72b6f61e8370e32 and removed in, I think, attempt to work around GNU Radio bug gnuradio/gnuradio#3184. We can revert it back at least for input device as we already have a workaround for that bug merged.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -189,6 +189,9 @@ void receiver::set_input_device(const std::string device)
     if (device.empty())
         return;
 
+    if (input_devstr.compare(device) == 0)
+        return;
+
     input_devstr = device;
 
     // tb->lock() can hang occasionally


argilo:
@kolbma Do you remember why you removed this check in #774? I think it would be useful to put it back, assuming it doesn't break anything.

argilo:
I found one possible reason: if one device is unplugged, and another is plugged in that happens to have the same device string, then the new device will not be initialized if `set_input_device` exits early.

argilo:
If the user opens the "Configure I/O devices" dialog, it's probably reasonable to assume they might also unplug or plug in a device. (This was the motivation for adding the "Device scan" button. So perhaps it's best not to skip `osmosdr::source::make` even if the device string happens to be the same.

vladisslav2011:
I've returned this check in attempt to at least partially work around #1019: as device reopen attempt always fails, skipping device reopen makes it possible to change sample rate/filter settings without a freeze. As #1019 is closed this partial work around is not required anymore. Closing.
P. S. While testing this branch, I've found a new interesting issue #1025.

kolbma:
Hey, yes the reason was when device got unplugged and plugged in again, there was the need to restart the application. With the forced device set in the dialog the scan succeeds.  
Maybe a bool parameter to force the reset during configuration would be an approvement?

argilo:
> Maybe a bool parameter to force the reset during configuration would be an approvement?

I don't think it's worth it. Let's just leave this as it is.

PR #1026, opened 2021-12-18
Iq tool add sampling rate and frequency GUI inputs

vladisslav2011:
This should help to play arbitrary named iq files at correct sampling rate and align waterfall/panadapter to correct center frequency.

### DIFF
--- a/src/qtgui/iq_tool.cpp
+++ b/src/qtgui/iq_tool.cpp
@@ -45,11 +45,14 @@ CIqTool::CIqTool(QWidget *parent) :
     is_recording = false;
     is_playing = false;
     bytes_per_sample = 8;
-    sample_rate = 192000;
+    recording_sample_rate = 192000;
+    playback_sample_rate = 192000;
     rec_len = 0;
     center_freq = 1e8;
 
     //ui->recDirEdit->setText(QDir::currentPath());
+    ui->sampleRateSpinBox->setValue(playback_sample_rate);
+    ui->centerFreqSpinBox->setValue(center_freq);
 
     recdir = new QDir(QDir::homePath(), "*.raw");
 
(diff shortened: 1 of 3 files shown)


pinkavaj:
First, thx for the MR. I have taken a look and found few thinks which I hope will help to get it into `mater`.

20221110_22h43m36s_grim

The numbers in spinboxes have different formating compared to what is already used, eg. LNB LO setting seen on the picture, which is aligned to the right and have the units set as a suffix right in the spin-box it, self.

vladisslav2011:
Thanks.
The 'sample rate' spin box should probably be in sps, like in the 'I/O devices' dialog:
gqrx-input-rate
The 'frequency' spin box format was picked from the 'Receiver options':
gqrx-frequency

Convert the 'Sample rate' to sps or convert both to Msps/MHz?

pinkavaj:
Good question. My guess is the sps is usually in range xx kHz (sound card input eg 48kHz) to XX MHz (eg. 20 MHz hackRf), 20000 kHz still looks IMHO readable so I would go for `ksps`.

For frequency I would probably vote for MHz.

The other option to consider might be to replace double spin boxes with integer and enable group separator, then even 1 GHz would make the number still readable (eg `1,000,000,000` in EN `1 000 000 000` in my country) and it would avoid using floats, which might get tricky at some point. So `Hz` and `sps` might really work.

src/qtgui/iq_tool.cpp
```diff
@@ -46,10 +46,13 @@ CIqTool::CIqTool(QWidget *parent) :
     is_playing = false;
     bytes_per_sample = 8;
     sample_rate = 192000;
+    playback_sample_rate = 192000;
```
pinkavaj, line 49: NIT: Consider rename the `sample_rate` to match the `playback_sample_rate` (eg. `record_sample_rate`). My hope is it will improve the code readability.

vladisslav2011:
I have not merged this into my primary dev branch. Usually I just rename/symlink files from other SW to Gqrx-accepted filename format.

vladisslav2011:
> The other option to consider might be to replace double spin boxes with integer and enable group separator

I tried this and failed. Integer spinboxes does not allow values larger, than 1000000000...

src/qtgui/iq_tool.cpp
vladisslav2011: Updated.

pinkavaj:
> > The other option to consider might be to replace double spin boxes with integer and enable group separator
> 
> I tried this and failed. Integer spinboxes does not allow values larger, than 1000000000...

Yep, having the 'sample rate' spin box the same as in 'I/O devices' dialog makes sense to me too.

The unit still should be moved from label into the spinbox suffix, at least this is main blocker IMHO.

PR #1029, opened 2021-12-22
Make some SoapySDR devices work when selected from drop-down list

vladisslav2011:
Correctly escape device strings to make SoapySDR devices work when selected from drop-down list.
HackRF and LimeSDR should work.
RTL-SDR does not work as it gets incorrectly enumerated: invalid rtl=# parameter is added.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.15.2: In progress...
+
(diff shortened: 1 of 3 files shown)


argilo (approved):
Thanks! This makes sense, since gr-osmosdr's `to_string()` only takes care of one layer of escaping, while its constructor decodes two layers. I'll merge once CI passes.

PR #1031, opened 2021-12-27
improved frequency accuracy at high zoom levels

dl2kcd:
At high zoom levels, like 100 Hz span or less at 145 MHz, there were several artifacts in the spectrum / waterfall plots that are addressed by this pull request:

- The resolution and accuracy of frequency labels is now 1 Hz (spectrum and tooltips).
- The alignment of spectrum plot and frequency grid is more exact.
- When zooming in/out by mouse wheel, the frequency at the mouse pointer now really remains fixed within 1 Hz.
- At extreme zoom levels, frequency panning remains possible.

A small python script was added that allows to create a reference I/Q data file to test the above features with the play I/Q data tool in GQRX.

Some remarks concerning the use of float or double in frequency calculations:

For the frequency labels, float variables have been changed to double, as float does not provide the required precision.
Type double is now also used in some conversions between plot x coordinate and frequency, while the original code preferred to use float. As long as only baseband frequencies are concerned, float may have sufficient accuracy, but double seems more safe. If performance on embedded systems is a concern and the original motive to use float, then calculations in CPlotter::getScreenIntegerFFTData(...) could be changed back from double to float, since they are performed plotWidth many times per spectrum update. Other occurances of double should only be rare events. On most systems, the performance advantage of float versus double is questionable.

Best regards, Joachim

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -146,7 +146,7 @@ CPlotter::CPlotter(QWidget *parent) : QFrame(parent)
     m_HdivDelta = 70;
     m_BandPlanHeight = 22;
 
-    m_FreqDigits = 3;
+    m_FreqDigits = 6;
 
     m_Peaks = QMap<int,int>();
     setPeakDetection(false, 2);
@@ -230,7 +230,7 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
                 if (m_TooltipsEnabled)
                     QToolTip::showText(event->globalPos(),
                                        QString("Demod: %1 kHz")
-                                               .arg(m_DemodCenterFreq/1.e3f, 0, 'f', 3),
+                                               .arg(m_DemodCenterFreq/1.e3, 0, 'f', 3),
                                        this);
             }
             else if (isPointCloseTo(pt.x(), m_DemodHiCutFreqX, m_CursorCaptureDelta))
@@ -267,7 +267,7 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
                 if (m_TooltipsEnabled)
                 {
                     qint64 hoverFrequency = freqFromX(pt.x());
-                    QString toolTipText = QString("F: %1 kHz").arg(hoverFrequency/1.e3f, 0, 'f', 3);
+                    QString toolTipText = QString("F: %1 kHz").arg(hoverFrequency/1.e3, 0, 'f', 3);
                     QFontMetrics metrics(m_Font);
                     int bandTopY = (m_OverlayPixmap.height() / m_DPR) - metrics.height() - 2 * VER_MARGIN - m_BandPlanHeight;
                     QList<BandInfo> hoverBands = BandPlan::Get().getBandsEncompassing(hoverFrequency);
@@ -302,7 +302,7 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
             QToolTip::showText(event->globalPos(),
                                QString("%1\n%2 kHz")
                                        .arg(tt.toString("yyyy.MM.dd hh:mm:ss.zzz"))
-                                       .arg(freqFromX(pt.x())/1.e3f, 0, 'f', 3),
+                                       .arg(freqFromX(pt.x())/1.e3, 0, 'f', 3),
                                this);
         }
     }
@@ -346,21 +346,24 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
             // pan viewable range or move center frequency
             int delta_px = m_Xzero - pt.x();
             qint64 delta_hz = delta_px * m_Span / (m_OverlayPixmap.width() / m_DPR);
-            if (event->buttons() & Qt::MidButton)
+            if (delta_hz != 0) // update m_Xzero only on real change
             {
-                m_CenterFreq += delta_hz;
-                m_DemodCenterFreq += delta_hz;
-                emit newDemodFreq(m_DemodCenterFreq, m_DemodCenterFreq - m_CenterFreq);
-            }
-            else
-            {
-                setFftCenterFreq(m_FftCenter + delta_hz);
-            }
-            updateOverlay();
+                if (event->buttons() & Qt::MidButton)
+                {
+                    m_CenterFreq += delta_hz;
+                    m_DemodCenterFreq += delta_hz;
+                    emit newDemodFreq(m_DemodCenterFreq, m_DemodCenterFreq - m_CenterFreq);
+                }
+                else
+                {
+                    setFftCenterFreq(m_FftCenter + delta_hz);
+                }
+                updateOverlay();
 
-            m_PeakHoldValid = false;
+                m_PeakHoldValid = false;
 
-            m_Xzero = pt.x();
+                m_Xzero = pt.x();
+            }
         }
     }
     else if (LEFT == m_CursorCaptured)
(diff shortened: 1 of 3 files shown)


argilo:
Thanks for the pull request! I agree that `float` should not be used for frequency calculations, because it doesn't have sufficient precision. A similar bug was fixed in #838.

A couple initial comments on this pull request:

1. I'm not sure `testdata/generate_testdata.py` should be included in Gqrx itself. You could just post the code in the description of this pull request.
2. At high zoom levels, the numbers on the frequency axis are partially truncated: Screenshot from 2021-12-27 12-21-47

vladisslav2011:
> 2\. At high zoom levels, the numbers on the frequency axis are partially truncated:

Suggested change:
```
diff --git a/src/qtgui/plotter.cpp b/src/qtgui/plotter.cpp
index 9950165..cc218f2 100644
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1435,7 +1435,7 @@ void CPlotter::drawOverlay()
     painter.setPen(QColor(PLOTTER_TEXT_COLOR));
     for (int i = 0; i <= m_HorDivs; i++)
     {
-        int tw = metrics.boundingRect(m_HDivText[i]).width();
+        int tw = metrics.boundingRect(QRect(0, 0, w, h), Qt::AlignHCenter | Qt::AlignVCenter, m_HDivText[i]).width();
         x = (int)((float)i*pixperdiv + adjoffset);
         if (x > m_YAxisWidth)
         {
```
This fixes rendering to me. Tested with QT_FONT_DPI=60, QT_FONT_DPI=120 and default.

argilo:
> Suggested change:

Even simpler would be to set `tw` to a large value (e.g. 256). I don't think there's any reason the rectangle can't be larger than the text.

dl2kcd:
Hi, I opt for Vladisskav's solution to calculate the label width correctly, just in case someone uses a jumbo screen where the labels are more than 256 pixels wide. I actually could not see the effect on my screen, but when I print tw, it now gives 1 or 2 more pixels than with the original code.
I see that the test data script should not be added to the code and removed it.
With these changes I performed a new commit and hope that is technically the right thing to do, as this is my very first pull request.

argilo:
> just in case someone uses a jumbo screen where the labels are more than 256 pixels wide

This is not a concern, because Qt uses a virtual coordinate system that is independent of the monitor's pixel density: https://doc.qt.io/qt-5/highdpi.html#high-dpi-support-on-a-system-level

> when I print tw, it now gives 1 or 2 more pixels than with the original code

That discrepancy is explained in the [documentation](https://doc.qt.io/qt-5/qfontmetrics.html#boundingRect-2):

> The bounding rectangle returned by this function is somewhat larger than that calculated by the simpler boundingRect() function. This function uses the [maximum left](https://doc.qt.io/qt-5/qfontmetrics.html#minLeftBearing) and [right](https://doc.qt.io/qt-5/qfontmetrics.html#minRightBearing) font bearings as is necessary for multi-line text to align correctly.

However, the documentation for minLeftBearing() and minRightBearing() also warn:

> Note that this function can be very slow if the font is large.

That suggests to me that `boundingRect()` expensive, and we'd be better off using a fixed-size rectangle, as suggested here: https://stackoverflow.com/questions/24831484/how-to-align-qpainter-drawtext-around-a-point-not-a-rectangle

> I performed a new commit and hope that is technically the right thing to do

In this case it would be better to squash your commits and force-push the updated commit. But don't worry if you have trouble doing that, since I can squash the commits for you before merging.

vladisslav2011:
256 pixels...

I think, the plotter width is a correct "safe value":
```
diff --git a/src/qtgui/plotter.cpp b/src/qtgui/plotter.cpp
index cc218f2..70046ca 100644
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1435,7 +1435,7 @@ void CPlotter::drawOverlay()
     painter.setPen(QColor(PLOTTER_TEXT_COLOR));
     for (int i = 0; i <= m_HorDivs; i++)
     {
-        int tw = metrics.boundingRect(m_HDivText[i]).width();
+        int tw = w;
         x = (int)((float)i*pixperdiv + adjoffset);
         if (x > m_YAxisWidth)
         {
```

dl2kcd:
Hi, so is `w = m_OverlayPixmap.width() / m_DPR` actually in the correct frame of reference, or should `OverlayPixmap.width()` be used?
Saving calls to `metrics.boundingRect(...)` for efficiency makes sense. Is it guaranteed that `painter.drawText(...)` does not become inefficient when called with an unnecessarily large rectangle? I would assume it uses the coordinates merely for calculations and not as loop boundaries, but I know little about the internals of Qt.
As for squashing the commits, I can try to do it just for the exercise, except that the Git documentations warns about rebasing commits that were already pushed to a public repo.

dl2kcd:
After a look at the Qt sources it seems that `drawText()` also handles overly large rectangles efficiently, so just using the window width for `tw` should have no performance impact. Also, after some experimenting with QT_SCREEN_SCALE_FACTORS and some trial text drawn on the screen it seems that `w` is already scaled correctly.

I pushed a new commit that simply sets `tw = w`.

BTW, I do not see how label clipping was related to including more digits, as I did not change the way `tw` was calculated from the text. I went back to the commit where I had branched off and inserted code that prints the original `tw` and the `tw` calculated using Vladisslav's first proposal, and the original values were already about 2 pixels smaller. So my expectation would be that labels are already clipped on your screen with fewer digits, unless this does not manifest itself by chance due to some rounding effects in pixel numbers.

Should I rally go ahead and squash commits despite the warning in the Git documentation?

argilo:
> I do not see how label clipping was related to including more digits

I expect this bug was always present, but became more obvious as the strings got longer.

> Should I rally go ahead and squash commits despite the warning in the Git documentation?

Yes please. The warning against rebasing applies to situations where other people may have done further work on top of your commits, but that shouldn't be the case here.

src/qtgui/plotter.cpp
```diff
@@ -1554,9 +1585,8 @@ int CPlotter::xFromFreq(qint64 freq)
 // Convert from screen coordinate to frequency
 qint64 CPlotter::freqFromX(int x)
 {
-    qint64 w = width();
-    qint64 StartFreq = m_CenterFreq + m_FftCenter - m_Span / 2;
-    qint64 f = StartFreq + m_Span * (qint64) x / w;
+    double ratio = (double)x / (double)width();
+    qint64 f = (m_CenterFreq + m_FftCenter - m_Span / 2) + ratio * m_Span;
```
argilo, line 1589: What is the rationale behind this change?

argilo:
Since squashing commits will remove this script, I'm posting it here:
```python
#!/usr/bin/env python3
#----------------------------------------------------------------------------
# Script to generate test data for GQRX.
# The test data allow to verify S-meter calibration and accuracy of
# displayed spectrum frequencies.
#----------------------------------------------------------------------------
import numpy as np
#----------------------------------------------------------------------------
# parameters
duration = 60         # duration in seconds
span     = 500e3      # sample rate and I/Q bandwidth
fcenter  = 100e6      # center frequency in Hz (used for filename)
foffsets = [-218.75e3, 125e3, 218.75e3] # offsets of test signals in Hz
sig_dBFS = -30        # signal amplitude in dBFS
noise_dBFS_1kHz = -50 # noise amplitude per 1 kHz bandwidth in dBFS
#----------------------------------------------------------------------------
# calculate data
sigamp   = pow(10, sig_dBFS / 20)
noiseamp = pow(10, noise_dBFS_1kHz / 20) * np.sqrt(span / 1e3)

t = np.arange(0, duration, 1/span, dtype = 'float64')
n = t.shape[0]

noise  = (np.random.normal(0,1,n) + 1j * np.random.normal(0,1,n)) / np.sqrt(2)

u = noiseamp * noise
for foffset in foffsets:
	signal = np.exp(2j * np.pi * foffset * t)
	u = u + sigamp * signal
#----------------------------------------------------------------------------
# save data
fname = 'gqrx_test_%d_%d_fc.raw' % (int(fcenter), int(span))
u.astype('complex64').tofile(fname)
print("output: %s" % fname)
```

src/qtgui/plotter.cpp
argilo: I guess it's meant to match the other `ratio` calculations, which makes sense. :+1:

argilo:
> Should I rally go ahead and squash commits despite the warning in the Git documentation?

Actually there's no need. I already rebased and squashed the commits before adding a commit of my own (to update the news file and contributors list).

I've tested this and it works very well, so I'll merge once CI passes.

dl2kcd:
Thanks a lot!  :-)

PR #1033, opened 2021-12-28
Set UHD package path in AppImage

### LINKED ISSUE #1032: Segmentation fault (Core dumped) when launch AppImage
Hi all,
I use the version 2.13.4 of the AppImage and I don't have problem, but when I want launch the version 2.14.X or 2.15.X I have a core dumped without error message.
I use Ubuntu 20.04
Do you have an idea ?

argilo:
Fixes #1032.

By pointing the `UHD_PKG_PATH` environment variable to the AppImage's mount point, UHD modules will be loaded from there instead of from the system (where they may be linked to an incompatible version of libuhd).

### DIFF
--- a/appimage.sh
+++ b/appimage.sh
@@ -69,6 +69,9 @@ chmod a+x *.AppImage
 mkdir -p ./AppDir/usr/lib
 cp -R /usr/lib/x86_64-linux-gnu/SoapySDR/modules* ./AppDir/usr/soapy-modules
 
+mkdir -p ./AppDir/apprun-hooks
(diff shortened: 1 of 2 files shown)


argilo:
@ebecamel @vladisslav2011 This pull request should fix the segfault when starting the AppImage on a system that has uhd-soapysdr installed. A pre-release build can be tested here: https://github.com/gqrx-sdr/gqrx/suites/4774227554/artifacts/133019782

argilo:
This is working well on my system, so I'll merge it.

PR #1034, opened 2021-12-28
Implement audio sample rate selection

### LINKED ISSUE #615: Lack of an option to change the audio output sample rate
Welcome,
Although there is a drop-down list dedicated for changing an audio output sample rate the only available option is "48 kHz".
Is it possible to add another options, such as 192 kHz?
According to [this message](https://groups.google.com/forum/#!msg/gqrx/hWX4qebbIwM/De1S_nxrCQAJ]) it is possible to achieve that by modifying directly the source code and recompiling the application.
Is there any reason a way to change this setting has not been included in the application by default beside the one described in the linked message? Is it (from the code perspective) a lack of a standarized/general method to deduce the maximum supported sample rate for a particular soundcard?
Or is there any way to change this setting without source code modifications, for example by manual editing default.conf configuration file?

### LINKED ISSUE #1018: BUG: White waterfall and no sound after playing back certain IQ files

### LINKED ISSUE #1030: BUG: Freeze after using WAV recorder while playing IQ file
Steps to reproduce:

1. Start gqrx
2. Open IQ Tool
3. Select some file
4. Click Play
5. Click Rec in audio dock
6. Click Rec in audio dock once more to stop recording
7. Click Play to stop playback
8. Try to close IQ Tool...

Expected behavior:
gqrx switches to hardware device, IQ Tool gets closed, etc.

Observed behavior:
Freeze

Current git master and appimage are affected.
Fixed in https://github.com/gqrx-sdr/gqrx/pull/1013 (just workaround GNU Radio bug by resetting reciever->rx before calling set_input_rate)
Primary cause: incorrect GNU Radio hier_block2_detail implementation...

vladisslav2011:
Close #615
Includes https://github.com/gqrx-sdr/gqrx/pull/1036, that closes #1030 and #1018.


argilo:
Could you please limit this pull request to just the audio rate selection? As I've mentioned previously, combining multiple changes into a single pull request makes it difficult to review.

vladisslav2011:
> Could you please limit this pull request to just the audio rate selection?

Making audio rate selection work properly automatically closes #1030 and #1018.
I can split bug fixes into separate PR and make this PR depend on new one.

src/applications/gqrx/receiver.cpp
```diff
@@ -124,14 +124,7 @@ receiver::receiver(const std::string input_device,
...
 
-#ifdef WITH_PULSEAUDIO
-    audio_snk = make_pa_sink(audio_device, d_audio_rate, "GQRX", "Audio output");
-#elif WITH_PORTAUDIO
-    audio_snk = make_portaudio_sink(audio_device, d_audio_rate, "GQRX", "Audio output");
-#else
-    audio_snk = gr::audio::sink::make(d_audio_rate, audio_device, true);
-#endif
-
+    make_audio_device(audio_device);
```
argilo, line 127: I think this would be clearer and more consistent with the rest of the code if `make_audio_device` returned the device instead of internally assigning it to `audio_snk`.

src/applications/gqrx/receiver.cpp
```diff
@@ -276,13 +260,8 @@ void receiver::set_output_device(const std::string device)
...
     try {
-#ifdef WITH_PULSEAUDIO
-        audio_snk = make_pa_sink(device, d_audio_rate, "GQRX", "Audio output");
-#elif WITH_PORTAUDIO
-        audio_snk = make_portaudio_sink(device, d_audio_rate, "GQRX", "Audio output");
-#else
-        audio_snk = gr::audio::sink::make(d_audio_rate, device, true);
-#endif
+        make_audio_device(device);
+        std::cerr<<"d_audio_rate="<<d_audio_rate<<std::endl;
```
argilo, line 264: Was this just for debugging?

argilo:
I noticed a bug: the frequency axis on the audio FFT seems to be incorrect when lower audio sample rates are used.

src/applications/gqrx/receiver.cpp
vladisslav2011: > `make_audio_device`

Just a bad name selection.
Renamed to `create_audio_device`. I hope, this should not interfere with GNU Radio naming conventions.
This function was created to eliminate some common code and get rid of some `#ifdefs`. Making it return the device would add more `#ifdefs`.

src/applications/gqrx/receiver.cpp
vladisslav2011: Yes. I forgot to remove it.
Removed.

vladisslav2011:
Thanks for a review.

> the frequency axis on the audio FFT seems to be incorrect

Confirmed. It may be fixed by setting correct sampling rate on `receiver::audio_fft`. Wait some minutes, I'll check it and reopen the PR.

argilo:
Another problem: filter widths are not restricted, and so aliasing can now occur when the audio rate is decreased. For instance, when I widen the USB filter, a tone that is outside the displayed filter skirts can be heard:

Screenshot from 2021-12-30 17-27-14

argilo:
This is because the demod ranges defined in `selectDemod` are built under the assumption that the audio rate will be 48000. For instance:

https://github.com/gqrx-sdr/gqrx/blob/7b736c41f150a2741f6a30d81d190c076e5e5003/src/applications/gqrx/mainwindow.cpp#L1157

argilo:
It looks like this will need some careful consideration to account for all the places in the UI and DSP that are making assumptions about the audio rate.

vladisslav2011:
> For instance, when I widen the USB filter, a tone that is outside the displayed filter skirts can be heard:

That's just filter leakage. It can be heard at 48000Hz rate too. Setting the filter to 'sharp' reduces the peak and tone, setting the filter to 'soft' increases it.
I see no aliasing even when the peak is in filter range:

Audio is resampled always correctly with filtering applied before decimation. https://github.com/gqrx-sdr/gqrx/blob/master/src/dsp/resampler_xx.cpp#L52-L60

On one hand, limiting the maximum filter width (AM, SSB, CW, RAW IQ modes) to a value, that will always fit into output bandwidth is a good idea, but on the other hand, altering user settings in unexpected way may be bad.
In case of FM the bandwidth includes the deviation, so in case of WFM it can be very wide even when modulating signal bandwidth is narrow. So, we should not limit the filter width in FM modes.
I've fixed the bug with incorrect frequency axis on the audio FFT. Reopening.

argilo:
> That's just filter leakage. 

Ah, I think you're right. After looking closer, I see that nbrx and wfmrx have audio resamplers late in their receive chains, which should prevent aliasing. My suggestion to limit the demod ranges was based on that incorrect assumption, so I think it's fine to leave them as they are.

I'll still need to do a bit more testing to make sure that there aren't any other bugs.

argilo:
I see now why I was confused by the behaviour when I expanded the filter bandwidth. The filter's transition width depends on the low & high filter cutoffs, so it becomes very large when the filter is expanded:

https://github.com/gqrx-sdr/gqrx/blob/7b736c41f150a2741f6a30d81d190c076e5e5003/src/applications/gqrx/receiver.cpp#L678-L705

This is surprising. I would have expected the transition width to depend only on the mode and the filter shape.

src/applications/gqrx/receiver.cpp
argilo: > Making it return the device would add more `#ifdefs`.

I suppose it would. I guess this is fine as it is.

argilo:
Another issue. While in 8 kHz mode, switching from USB to FM (stereo) leaves the audio FFT in a weird configuration where negative frequencies are displayed:

Screenshot from 2021-12-31 00-47-29

argilo:
Maybe that happens because the requested range exceeds what is possible given the current audio sample rate. There's even a FIXME about that:

https://github.com/gqrx-sdr/gqrx/blob/4eaeb163ac470d187d928a71622fb5dbdaedaa9e/src/applications/gqrx/mainwindow.cpp#L1132

Many other modes similarly request FFT ranges that may not be possible.

argilo:
I guess all the ranges should be capped to a maximum of half the audio rate.

vladisslav2011:
> I guess all the ranges should be capped to a maximum of half the audio rate.

Thanks for suggestion. This works very well.

vladisslav2011:
Rebased on top of current master.
It was quite hard. The previous "recreate the demod from scratch" approach was much, much easier.
I was hit by https://github.com/gnuradio/gnuradio/issues/5436 several times during rebase, so this PR needs some testing to ensure, that it is free from regressions...

I have added 88200 Hz and 96000 Hz audio rates (mostly for "RAW IQ" mode).

vladisslav2011:
Rebased on top of current master.
And closed by accident.
How can I reopen this PR? Pushing new commits, as github suggests, does not help...

vladisslav2011:
OK. Pushing a new commit solved the issue.

argilo:
Sorry, it looks like I caused a merge conflict when I merged #1066.

I just had a quick look over this code and it's looking decent. I think I'll try to including this in the next feature release (2.16) since it's a new feature.

vladisslav2011:
> I caused a merge conflict

Resolved.

src/dsp/fm_deemph.h
```diff
@@ -58,13 +58,15 @@ class fm_deemph : public gr::hier_block2
     ~fm_deemph();
 
     void set_tau(double tau);
+    void set_rate(double rate);
```
sultanqasim, line 61: `rate` argument should be a float, given that `d_quad_rate` is a float. Otherwise, double promotion warnings will result.

sultanqasim:
Aside from my nitpick regarding the double precision argument to the `set_rate` function in fm_deemph.h, this patch is working very well for me and I have not noticed any regressions in my testing. I would support this being merged once this the warning causing double precision nitpick is fixed.

src/dsp/fm_deemph.h
vladisslav2011: Fixed.
Thanks.

Issue #1039, opened 2022-01-01
Regression: some receiver settings does not get restored when switching from IQ Tool to hardware device


vladisslav2011:
...or when switching input rates/toggling DC remove.
Introduced in e5e16ccff5e7b50f2f9c50b222819e198a867e5d.
I think, we should store rx filter parameters, cw offset and SQL level to private class members on change and restore when the demodulator gets recreated.
And we should track and restore state of RDS decoder too to prevent a freeze.
PR in progress...

argilo:
Couldn't we just use something like loadConfig to load all the settings?

vladisslav2011:
No.
`loadConfig` does not restore required settings in this case. `loadConfig` is called when switching input rates/changing input devices, but SQL level and filter low/high cut frequencies still get reset to default. The GUI shows everything to be correct, but demodulator state goes out of sync.

vladisslav2011:
PR is ready https://github.com/gqrx-sdr/gqrx/pull/1040

argilo:
It would probably be worth adding those settings into the settings file. That would prevent them from being lost when Gqrx is closed, and would remove the need to add yet more complexity to the I/Q tool.

argilo:
In any case, that can go in a separate pull request. I'll have a look at this one when I get a chance.

vladisslav2011:
These settings are already added to the settings file and restored correctly when gqrx is closed and then restarted. This is different issue.

vladisslav2011:
That's not enough.
We should copy NFM max deviation, deemphasis, agc settings, AM dc reduction too.
Adding settings-copying constructor/factory to the `receiver_base` class and tracking setting values in `receiver_base` private fields  may be better solution.

vladisslav2011:
PR redone. All settings are restored correctly now.

vladisslav2011:
It looks, like this is not the regression.
It's just separate bug, that was made to occur more often after https://github.com/gqrx-sdr/gqrx/commit/e5e16ccff5e7b50f2f9c50b222819e198a867e5d.
It can be reproduced by following steps:
1. Set AM DCR to disabled. The audio spectrum should show DC spike at center.
2. Switch to WFM
3. Switch back to AM

Expected result:
DC spike on the audio spectrum and AM DCR setting disabled.

Observed result:
No DC spike on the audio spectrum, but AM DCR setting disabled.

vladisslav2011:
> It would probably be worth adding those settings into the settings file.

[...]

PR #1036, opened 2021-12-30
Fix #1018

vladisslav2011:
And automatically fix #1030

PR #1040, opened 2022-01-01
Restore demodulator settings after recreating it

vladisslav2011:
Fix regression, introduced in e5e16ccff5e7b50f2f9c50b222819e198a867e5d.
Store filter limits and shape to private class members.
Restore filter bandwidth/offset, squelch level, cw offset after the
demodulator is recreated in connect_all to match actual settings.
Track state of RDS decoder and restore it too to prevent a freeze.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -24,6 +24,7 @@
 #include <iomanip>
 #include <iostream>
 #include <sstream>
+#include <typeinfo>
 #include <QDebug>
 
 #include <gnuradio/prefs.h>
@@ -1325,7 +1326,9 @@ void receiver::connect_all(rx_chain type)
     // Visualization
     tb->connect(b, 0, iq_fft, 0);
 
+    bool rds_decoder_active = rx && rx->is_rds_decoder_active();
     // RX demod chain
+    receiver_base_cf_sptr old_rx = rx;
     rx.reset();
     switch (type)
     {
@@ -1354,6 +1357,10 @@ void receiver::connect_all(rx_chain type)
         tb->connect(rx, 1, audio_gain1, 0);
         tb->connect(audio_gain0, 0, audio_snk, 0);
         tb->connect(audio_gain1, 0, audio_snk, 1);
+        if(old_rx)
+            rx->restore_settings(old_rx);
+        if(rds_decoder_active)
+            rx->start_rds_decoder();
     }
 
     // Recorders and sniffers
--- a/src/dsp/rx_demod_am.cpp
+++ b/src/dsp/rx_demod_am.cpp
@@ -86,20 +86,16 @@ void rx_demod_am::set_dcr(bool dcr)
     if (d_dcr_enabled)
     {
         // Switching from ON to OFF
-        lock();
         disconnect(d_demod, 0, d_dcr, 0);
         disconnect(d_dcr, 0, self(), 0);
         connect(d_demod, 0, self(), 0);
-        unlock();
     }
     else
     {
         // Switching from OFF to ON
-        lock();
         disconnect(d_demod, 0, self(), 0);
         connect(d_demod, 0, d_dcr, 0);
         connect(d_dcr, 0, self(), 0);
-        unlock();
     }
 
     d_dcr_enabled = dcr;
@@ -164,20 +160,16 @@ void rx_demod_amsync::set_dcr(bool dcr)
     if (d_dcr_enabled)
     {
         // Switching from ON to OFF
-        lock();
         disconnect(d_demod2, 0, d_dcr, 0);
         disconnect(d_dcr, 0, self(), 0);
         connect(d_demod2, 0, self(), 0);
-        unlock();
     }
     else
     {
         // Switching from OFF to ON
-        lock();
         disconnect(d_demod2, 0, self(), 0);
         connect(d_demod2, 0, d_dcr, 0);
         connect(d_dcr, 0, self(), 0);
-        unlock();
     }
 
     d_dcr_enabled = dcr;
--- a/src/receivers/nbrx.cpp
+++ b/src/receivers/nbrx.cpp
@@ -120,11 +120,13 @@ void nbrx::set_audio_rate(float audio_rate)
 
 void nbrx::set_filter(double low, double high, double tw)
 {
+    receiver_base_cf::set_filter(low, high, tw);
     filter->set_param(low, high, tw);
 }
 
 void nbrx::set_cw_offset(double offset)
 {
+    receiver_base_cf::set_cw_offset(offset);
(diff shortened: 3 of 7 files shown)


argilo:
Destroying the whole receiver in #1036 seems like a heavy-handed approach that's now resulting in the need to add more complexity to the receivers here. Do you think it would be simpler to just add a reset method to the receivers that would re-initialize things like IIR blocks?

vladisslav2011:
> Do you think it would be simpler to just add a reset method to the receivers that would re-initialize things like IIR blocks?

I don't think so. Resetting IIR blocks by calling set_taps may be enough to fix the white waterfall and no sound issue, but this would leave the second bug (triggering https://github.com/gnuradio/gnuradio/issues/5436) open. Recreating the whole receiver is quick and efficient fix. If the receiver will be changed: more modulations added, digital modes and more, then we'll have to track IIR block usage and not forget to add them to reset method. More work, more code, more possible bugs.
I can implement a different solution to https://github.com/gqrx-sdr/gqrx/issues/1018 (IIR filters resetting `with set_taps`) in separate PR just to check, how would it work and to see, how much changes would it contain.
And it looks, like https://github.com/gqrx-sdr/gqrx/issues/1030 should be reopened with slightly different description: "Random GUI freeze after stopping WAV recorder and changing receiver settings".
I did some tests and was able to reproduce it:
1. Select AM demodulator
2. Start WAV recorder
3. Stop WAV recorder
4. Change AM demodulator DCR setting

I have updated the linked issue: https://github.com/gqrx-sdr/gqrx/issues/1039. It looks, like this is a separate bug, not a regression after https://github.com/gqrx-sdr/gqrx/commit/e5e16ccff5e7b50f2f9c50b222819e198a867e5d...

vladisslav2011:
> I can implement a different solution to #1018

Done. See my branch https://github.com/vladisslav2011/gqrx/tree/white_waterfall_reset_iir.
Long story short: it does not work at all...

argilo:
What I had in mind was to destroy & recreate the IIR blocks, not just set the taps.

argilo:
> `set_taps` is expected to do exactly the thing, we need

Oh, nice!

> But it does not work. [...] It looks, like simple_squelch is blocking the signal. I'll try to recreate it too.

Sounds good. Whether or not we go with this approach, it will be good to understand which blocks have an infinite impulse response and therefore cannot tolerate invalid input.

argilo:
The Simple Squelch block does use an IIR filter under the covers, so that would explain it:

https://github.com/gnuradio/gnuradio/blob/b8d853a1a04b16cf9261e216b5330a4e7523b6ad/gr-analog/lib/simple_squelch_cc_impl.h#L25

argilo:
Also, its `set_alpha` method resets the IIR filter's taps, so I expect that would get it back into a good state.

vladisslav2011:
The second attempt: https://github.com/vladisslav2011/gqrx/tree/white_waterfall_reset_iir2
Setting taps does not work, recreating the block works.

argilo:
> Setting taps does not work

Why not? I'd want to understand why that's the case before rejecting the approach.

vladisslav2011:
It does not work, because there are still invalid samples in the buffer...
This may be solved by emitting a tag from src, marking the first valid buffer start and setting taps from the tag handler at exact point of device switch...

vladisslav2011:
There are 2 reasons, why resetting iir filters does not work:
1. `single_pole_iir_filter_cc` block, used by `dc_corr` block does not have a `reset` method. The underlying `single_pole_iir` template class does have a `reset` method, but it is not exposed by the `single_pole_iir_filter_cc` class. So, there is now way to reset a `single_pole_iir_filter_cc` block without recreating it.
2. `simple_squelch` block have exactly the same issue: It uses the `single_pole_iir` template class, but does not use it's `reset` method.

vladisslav2011:
The https://github.com/vladisslav2011/gqrx/tree/white_waterfall_reset_iir2 branch does not work well on my 'weak' test system - Intel Atom N270 (single core, 32-bit only) laptop. The waterfall gets restored, but the sound does not.
The previous solution, merged into master, works almost well. The spectrogram gets overflowed and needs to be reset by switching to different fft width - that's different issue, related to GNU Radio 3.7.9, or Ubuntu 16.04 or 32-bit CPU. My 'test' branch https://github.com/vladisslav2011/gqrx/tree/combined works very well on such  an ancient hardware: WBFM mono @2.88Msps. Better, than ever. WBFM stereo it still limited to 300ksps max.

argilo:
I'm starting to think that it might not be worth the effort to fix #1018 (and potentially introduce new bugs along the way), and that a better course of action might be just to revert #1036. If a user plays an invalid I/Q file and messes up the IIR filters, they can always restart Gqrx.

argilo:
In that case, an alternate solution for #1030 would be needed, but a temporary workaround might be more appropriate there since https://github.com/gnuradio/gnuradio/issues/5436 should eventually be fixed.

vladisslav2011:
> If a user plays an invalid I/Q file and messes up the IIR filters, they can always restart Gqrx.

The restart is not required. Disabling/enabling "DC Cancel" is enough to fix the 'White waterfall' for one time. Switching from narrow to WFM and back or vice versa is enough to restore the sound for one time. The #1018 is not so critical, as some crashes are. It's more a usability issue. I've faced it multiple times while trying to play different IQ files, downloaded from internet to test AFSK1200 tool for regressions (have not found correct AX.25 recording yet). https://github.com/gqrx-sdr/gqrx/pull/1036 is a quick attempt to fix it in easy way. It exposed the #1039 and introduced a small regression (triggering https://github.com/gnuradio/gnuradio/issues/5436 after switching input devices with RDS decoder enabled). Both issues are fixed by this PR. From my point of view, always resetting the receiver after demodulator/input device change is good thing as it leaves much less points, where more regressions, related to triggering https://github.com/gnuradio/gnuradio/issues/5436, may be introduced.

> In that case, an alternate solution for #1030 would be needed

There is an alternate fix, included here: https://github.com/vladisslav2011/gqrx/commits/white_waterfall_reset_iir2
Should I split it into separate commit and open a PR?

argilo:
Closing this, as it's not needed now that #1036 is reverted.

argilo:
> always resetting the receiver after demodulator/input device change is good thing as it leaves much less points, where more regressions, related to triggering gnuradio/gnuradio#5436, may be introduced.

There were still plenty of places that gnuradio/gnuradio#5436 could be triggered even after #1036. I'd rather make a targeted workaround (as in #1049) instead of redesigning Gqrx to work around an upstream bug that will eventually be gone.

argilo:
> This may be solved by emitting a tag from src, marking the first valid buffer start and setting taps from the tag handler at exact point of device switch...

Agreed, stream tags might be a good solution.

PR #1041, opened 2022-01-03
Reconnection fixes

vladisslav2011:
Disconnect iq_swap and udp_sink when not in use to reduce cpu load.
Connect audio consumers to audio_gain# to prevent clipping in recorded files/prevent sending clipped samples to AFSK1200 decoder and external programs and to make rx_fft reflect gain changes.


vladisslav2011:
This may be splint into several small PR's.

vladisslav2011:
Done with splitting.
audio_gain part is skipped as there is better solution in https://github.com/gqrx-sdr/gqrx/pull/1061
Closing.

PR #1042, opened 2022-01-03
Get rid of boost::circular_buffer to improve performance

vladisslav2011:
Replace boost::circular_buffer with stdlib/GNU Radio alternatives
rx_fft_c,rx_fft_f: replace boost::circular_buffer with GNU Radio buffer/reader
rx_meter_c: replace boost::circular_buffer with GNU Radio buffer/reader
rx_rds: replace boost::circular_buffer with std:queue as it fits better here
sniffer_f: replace boost::circular_buffer with GNU Radio buffer/reader

### DIFF
--- a/src/dsp/rx_fft.h
+++ b/src/dsp/rx_fft.h
@@ -29,6 +29,9 @@
 #include <gnuradio/filter/firdes.h>       /* contains enum win_type */
 #include <gnuradio/gr_complex.h>
 #include <gnuradio/buffer.h>
+#if GNURADIO_VERSION >= 0x031000
+#include <gnuradio/buffer_reader.h>
+#endif
 #include <chrono>
 
 
@@ -88,12 +91,12 @@ public:
     void set_window_type(int wintype);
     int  get_window_type() const;
 
-    void set_fft_size(int fftsize);
+    void set_fft_size(unsigned int fftsize);
     void set_quad_rate(double quad_rate);
     unsigned int get_fft_size() const;
 
 private:
-    int d_fftsize;   /*! Current FFT size. */
+    unsigned int d_fftsize;   /*! Current FFT size. */
     double       d_quadrate;
     int          d_wintype;   /*! Current window type. */
 
@@ -159,11 +162,11 @@ public:
     void set_window_type(int wintype);
     int  get_window_type() const;
 
-    void set_fft_size(int fftsize);
+    void set_fft_size(unsigned int fftsize);
     unsigned int get_fft_size() const;
 
 private:
-    int d_fftsize;   /*! Current FFT size. */
+    unsigned int d_fftsize;   /*! Current FFT size. */
     double       d_audiorate;
     int          d_wintype;   /*! Current window type. */
 
--- a/src/dsp/rx_fft.h
+++ b/src/dsp/rx_fft.h
@@ -28,7 +28,7 @@
 #include <gnuradio/fft/fft.h>
 #include <gnuradio/filter/firdes.h>       /* contains enum win_type */
 #include <gnuradio/gr_complex.h>
-#include <boost/circular_buffer.hpp>
+#include <gnuradio/buffer.h>
 #include <chrono>
 
 
@@ -88,12 +88,12 @@ public:
     void set_window_type(int wintype);
     int  get_window_type() const;
 
-    void set_fft_size(unsigned int fftsize);
+    void set_fft_size(int fftsize);
     void set_quad_rate(double quad_rate);
     unsigned int get_fft_size() const;
 
 private:
-    unsigned int d_fftsize;   /*! Current FFT size. */
+    int d_fftsize;   /*! Current FFT size. */
     double       d_quadrate;
     int          d_wintype;   /*! Current window type. */
 
@@ -106,7 +106,8 @@ private:
 #endif
     std::vector<float>  d_window; /*! FFT window taps. */
 
-    boost::circular_buffer<gr_complex> d_cbuf; /*! buffer to accumulate samples. */
+    gr::buffer_sptr d_writer;
+    gr::buffer_reader_sptr d_reader;
     std::chrono::time_point<std::chrono::steady_clock> d_lasttime;
 
     void do_fft(unsigned int size);
@@ -158,11 +159,11 @@ public:
     void set_window_type(int wintype);
     int  get_window_type() const;
 
-    void set_fft_size(unsigned int fftsize);
+    void set_fft_size(int fftsize);
     unsigned int get_fft_size() const;
 
 private:
-    unsigned int d_fftsize;   /*! Current FFT size. */
+    int d_fftsize;   /*! Current FFT size. */
     double       d_audiorate;
     int          d_wintype;   /*! Current window type. */
 
@@ -175,7 +176,8 @@ private:
 #endif
     std::vector<float>  d_window; /*! FFT window taps. */
 
-    boost::circular_buffer<float> d_cbuf; /*! buffer to accumulate samples. */
+    gr::buffer_sptr d_writer;
+    gr::buffer_reader_sptr d_reader;
     std::chrono::time_point<std::chrono::steady_clock> d_lasttime;
 
     void do_fft(unsigned int size);
--- a/src/dsp/rx_meter.cpp
+++ b/src/dsp/rx_meter.cpp
@@ -39,7 +39,11 @@ rx_meter_c::rx_meter_c(double quad_rate)
       d_avgsize(quad_rate * 0.100)
 {
     /* allocate circular buffer */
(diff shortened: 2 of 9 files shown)


vladisslav2011:
Is it fine to unlock the mutex before applying the window?
https://github.com/gqrx-sdr/gqrx/pull/1042/commits/8abfc65cf52f6a04c5484e52526acb1ff032c11a#diff-fcf4675691a139e7f1a9f0884be68dd18e1daa93b3da5517d039e0c6c935d852R125-R126
I did some tests and see no artifacts both on AMD Ryzen 5950X system and Intel Atom N270 system. So, it should be fine, but may still require more testing.

argilo:
Could you please write proper descriptions for your pull requests?

I would suggest reading over this blog post, gives many tips on how to write good pull requests:

https://mtlynch.io/code-review-love/

vladisslav2011:
Sorry for being a bit lazy after holidays.
Most of changes here are made to improve performance, but switching to libvolk functions may be a bit unrelated to the PR title.
Is it better to split libvolk stuff into separate PR?

vladisslav2011:
I think, the answer would be "yes", so...
libvolk stuff is moved to a separate PR: https://github.com/gqrx-sdr/gqrx/pull/1044

src/dsp/rx_fft.h
```diff
@@ -94,6 +94,7 @@ class rx_fft_c : public gr::sync_block
 
 private:
     unsigned int d_fftsize;   /*! Current FFT size. */
+    unsigned int n_fftsize;   /*! New FFT size. */
```
argilo, line 97: Gqrx follows [GNU Radio's coding conventions](https://github.com/gnuradio/greps/blob/master/grep-0001-coding-guidelines.md#classes), including:

> All class data members shall begin with `d_foo`
>
> The big benefit is when you're staring at a block of code it's obvious which of the things being assigned to persist outside of the block.

argilo: Perhaps `d_new_fftsize` would be a better name.

But could you also explain why this new member variable is needed?

argilo:
What is the motivation behind removing `unsigned` in a bunch of places?

src/dsp/rx_fft.h
argilo: Also, why is it needed for `rx_fft_c` but not `rx_fft_f`?

I'd like to keep these classes as similar as possible, but the changes here are making them diverge.

src/dsp/rx_meter.cpp
```diff
@@ -55,11 +56,12 @@ int rx_meter_c::work(int noutput_items,
 {
     std::lock_guard<std::mutex> lock(d_mutex);
 
-    const gr_complex *in = (const gr_complex *) input_items[0];
-    (void) output_items; // unused
```
argilo, line 59: This should probably be kept, since it's meant to suppress a compiler warning.

src/dsp/rx_meter.cpp
```diff
@@ -55,11 +56,12 @@ int rx_meter_c::work(int noutput_items,
...
-
-    for (int i = 0; i < noutput_items; i++)
-        d_cbuf.push_back(in[i]);
+    if (noutput_items > d_writer->bufsize())
+        noutput_items = d_writer->bufsize();
```
argilo, line 60: I don't think this is right. This block is meant to skip over samples that cannot fit in the circular buffer. Reducing `noutput_items` will tell the GNU Radio scheduler that they haven't been consumed.

src/dsp/rx_rds.h
```diff
@@ -82,7 +82,7 @@ class rx_rds_store : public gr::block
     void store(pmt::pmt_t msg);
 
     std::mutex d_mutex;
-    boost::circular_buffer<pmt::pmt_t> d_messages;
+    std::queue<pmt::pmt_t> d_messages;
```
argilo, line 85: I presume `circular_buffer` was used (in #226) to ensure that `d_messages` didn't get too large. Is that concern unwarranted? (I think it might be, but I want to be sure before we remove the size limit.)

argilo:
I'm looking forward to getting rid of circular_buffer, since it is Gqrx's last dependency on Boost!

src/dsp/rx_fft.h
vladisslav2011: > why this new member variable is needed?

I want to unlock the mutex as soon, as possible to not block the GNU Radio thread for the time of FFT calculation. This solves audio dropouts and reduces tuning lag when the FFT size is set to a large number (512k..1M) or when the CPU is not capable of calculating the FFT in reasonable time (Atom N270 for example).
The variable is introduced to avoid race between `get_fft_data` and `set_fft_size` when `get_fft_data` is executing in different thread. This commit was picked from my different branch, where I tried to offload fft calculation to a dedicated thread. This worked very well, but it turns out that the rendering of waterfall/spectrum is next heavy task too be offloaded to the dedicated thread, so no PR yet.
I think, this change is unrelated and should be moved to that separate branch.
And mutex unlocking point change should be moved to a separate commit/PR as it may need more testing (see my first comment). Maybe we should apply the window under locked mutex and unlock it before actual FFT calculation...

I'll remove this stuff and push a clean version. Wait some minutes, please.

src/dsp/rx_meter.cpp
vladisslav2011: I have not removed it after attempting to reduce memory usage by shrinking the buffer to 2 * d_fftsize. This was done to avoid writing behind the end of the buffer.
If we tell GNU Radio scheduler that some items are not consumed, then GNU Radio scheduler will call `work` again until all items get consumed, so no problem here.
Shrinking the buffer did not work as expected - artifacts at low FFT size and low input rate, so I reverted the change, but forgot to remove this safety measure.
I still think, that we can safely limit the buffer to 2 * max FFT size and not waste ~160Mb and more of shared memory in case of HackRF/ LimeSDR at high sampling rate. But this should go into separate PR.

argilo: > If we tell GNU Radio scheduler that some items are not consumed, then GNU Radio scheduler will call `work` again until all items get consumed, so no problem here.

True, but there will be extra overhead, won't there? We could instead just skip over the excess input samples.

src/dsp/rx_rds.h
vladisslav2011: I can't imagine a situation where `MainWindow::rdsTimeout` would be unable to read out all messages fast enough. In that case the main thread would lock up reading RDS messages.
The maximum number of unread messages, I have ever seen is 1: `rx->get_rds_data(buffer, num)` sets `num` to 0 or -1 only.

vladisslav2011:
Thanks for a review.
I have updated the branch including suggested changes.
`n_fftsize` and delaying `set_params` are not required until introduction of dedicated FFT thread, so I removed it.

src/dsp/rx_rds.h
argilo: Right, as long as the RDS timer is running while the RDS decoder is (which looks to be the case) this won't be a problem. Thanks for checking!

argilo: While looking into this, I noticed that `MainWindow::rdsTimeout` has a bug. Fixed in #1050.

argilo:
I rebased, did some cleanup, fixed some bugs, and added support for GNU Radio 3.10 (which has made changes to the buffer interface).

So far it builds against GNU Radio 3.10, but I haven't done much testing yet. Let's see how CI goes. I'd appreciate help with testing.

vladisslav2011:
I do not like the idea of throwing samples away https://github.com/gqrx-sdr/gqrx/pull/1042/commits/33746f340b7a46b88f43ec88824779db82212fc3#diff-fcf4675691a139e7f1a9f0884be68dd18e1daa93b3da5517d039e0c6c935d852R96-R98
But it should never occur in the real world, so no problem.
Builds and works with GNU Radio git master. (No sound due to "Pulseaudio error: Connection refused")
Thanks.

argilo:
> I do not like the idea of throwing samples away

The only alternative would be to make the buffer larger. It is already long enough to store one full second of extra data, which should be enough unless the hardware input buffer is set to some excessively large size.

argilo:
Also, the previous code threw away samples in the same way. It wrote them all to the circular buffer, but of course when there were more than the length of the buffer the earliest samples would be overwritten, throwing them away.

argilo:
When using a USRP B200 at 32 Msps, this gives a noticeable improvement in CPU utilization (about 7-8% less on my laptop).

I haven't noticed any bugs yet, so I'll merge this.

argilo:
While reviewing this, I noticed that the sniffer / AFSK1200 receiver probably doesn't need a circular buffer. I think it should just do its signal processing directly in the `work` function.

PR #1043, opened 2022-01-03
Store missing demod settings to the config file and restore at startup

vladisslav2011:
Store AM DCR setting to receiver/am_dcr_off
    Store AM-Sync DCR setting to receiver/amsync_dcr_off
    Store AM-Sync PLL setting to receiver/amsync_pllbw
    Restore everything at startup.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1103,6 +1103,7 @@ void MainWindow::selectDemod(int mode_idx)
 
     case DockRxOpt::MODE_AM:
         rx->set_demod(receiver::RX_DEMOD_AM);
+        rx->set_am_dcr(uiDockRxOpt->currentAmDcr());
         ui->plotter->setDemodRanges(-40000, -200, 200, 40000, true);
         uiDockAudio->setFftRange(0,6000);
         click_res = 100;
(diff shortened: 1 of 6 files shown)


vladisslav2011:
Retested. Works correctly after rebase.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -1185,16 +1187,20 @@ void MainWindow::selectDemod(int mode_idx)
...
     ui->plotter->setFilterClickResolution(click_res);
-    rx->set_filter((double)flo, (double)fhi, d_filter_shape);
-    rx->set_cw_offset(cwofs);
-    rx->set_sql_level(uiDockRxOpt->currentSquelchLevel());
+    if (mode_idx != DockRxOpt::MODE_OFF)
+    {
+        rx->set_filter((double)flo, (double)fhi, d_filter_shape);
+        rx->set_cw_offset(cwofs);
+        rx->set_sql_level(uiDockRxOpt->currentSquelchLevel());
+     }
```
argilo, line 1195: Is this change related to the DCR/PLL changes, or is it fixing something else?

vladisslav2011: This may be unrelated. It just saves 3 calls. When the demodulator is disabled there is no filter to set hicut/locut/transition/offset and no squelch too. So it's pointless to parameter of missing blocks.

I see another problem: AMPLL PLL bandwidth is not set on demod selection, but it should be set.

vladisslav2011: Updated.

argilo:
I made a few changes in bb783ba5dd72a78a45cd6883630080679d91b01f. Let me know if you're happy with them.

* Replace `setCheckState` / `checkState` with `setChecked` / `isChecked`.
* Replace `am_dcr_off` / `amsync_dcr_off` with `am_dcr` / `amsync_dcr`.
* Multiply PLL value by 1e6, in case the PLL range is extended in the future.
* Don't store `amsync_pllbw` if it's the default value (Medium).

vladisslav2011:
Everything is fine.

argilo:
Sounds good.

I've tested and it works well. I'll merge once CI is happy.

PR #1044, opened 2022-01-04
Rx_fft rx_meter switch to volk

vladisslav2011:
Switch to using appropriate libvolk functions instead of trivial
for-loops. This should improve performance and reduce cpu load in some cases.

Requires https://github.com/gqrx-sdr/gqrx/pull/1042

### DIFF
--- a/src/dsp/rx_fft.cpp
+++ b/src/dsp/rx_fft.cpp
@@ -21,6 +21,7 @@
  * Boston, MA 02110-1301, USA.
  */
 #include <math.h>
+#include <volk/volk.h>
 #include <gnuradio/io_signature.h>
 #include <gnuradio/filter/firdes.h>
 #include <gnuradio/gr_complex.h>
@@ -148,8 +149,7 @@ void rx_fft_c::do_fft(unsigned int size)
     if (d_window.size())
     {
         gr_complex *dst = d_fft->get_inbuf();
-        for (unsigned int i = 0; i < size; i++)
-            dst[i] = p[i] * d_window[i];
+        volk_32fc_32f_multiply_32fc(dst, p, &d_window[0], size);
     }
     else
     {
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -3,7 +3,7 @@
 
      FIXED: RDS receiver skips some messages.
      FIXED: Buffer overruns in AFSK1200 decoder.
-  IMPROVED: FFT performance.
+  IMPROVED: FFT and S-meter performance.
 
 
     2.15.2: Released January 8, 2022
--- a/src/dsp/rx_meter.cpp
+++ b/src/dsp/rx_meter.cpp
@@ -21,6 +21,7 @@
  * Boston, MA 02110-1301, USA.
  */
 #include <math.h>
+#include <volk/volk.h>
 #include <gnuradio/io_signature.h>
 #include <dsp/rx_meter.h>
 #include <iostream>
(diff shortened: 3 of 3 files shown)


argilo:
Rebased to resolve a merge conflict.

src/dsp/rx_fft.cpp
```diff
@@ -359,16 +359,9 @@ void rx_fft_f::do_fft(unsigned int size)
     gr_complex *dst = d_fft->get_inbuf();
     float * p = (float *)d_reader->read_pointer();
     /* apply window, and convert to complex */
+    volk_32f_x2_interleave_32fc(dst, p, p, size);
```
argilo, line 362: Shouldn't the `q` buffer be zero here? Setting it to `p` increases the power of the signal by 3 dB, which makes the audio FFT inaccurate.

vladisslav2011: Thanks.
Yes. This in inaccurate.
Keeping a buffer full of zeroes does not sound as a good idea.
I think, we can compensate for the error by dividing the window by 2, maybe with `volk_32f_s32f_multiply_32f` here https://github.com/gqrx-sdr/gqrx/blob/801aa9e43e83e23f16346a9d5cabee4b772e317a/src/dsp/rx_fft.cpp#L426
It would be called once as there is no way to change audio FFT parameters.
And when we'll switch to complex input to this FFT to show RAW IQ demodulator output correctly, we'll have to disable this scaling for this specific demodulator type.

argilo: That sounds complicated, and I think someone reading the code might find it confusing that `rx_fft_f` computes the FFT of the signal multiplied by 1+i.

Maybe for now it would be simplest to remove the VOLK optimization from `rx_fft_f`. The audio rate is much lower than the input sample rate, so it should not be a performance bottleneck.

vladisslav2011: OK.
I've reverted 'empty window' branch to trivial for-loop and rebased to current master.

argilo: Thanks. The other branch still looks broken though?

vladisslav2011: Sorry. I'm wrong. Wait a minute. I'll push correct commit.

argilo: I think it would be best to limit this to `rx_fft_c` and `rx_meter_c` for now, since the calculations there are unchanged and those are the high-sample rate blocks.

vladisslav2011: Ready.
This should be correct. I'll test on my Intel Atom N270 system to see if there would be any improvement from VOLK optimization of `rx_fft_f`.

vladisslav2011: No noticeable difference even on Intel Atom N270. Let's keep it untouched and add a special case for handling complex output from 'RAW IQ' demodulator later.

argilo:
:thinking: So far I have not been able to observe a performance improvement on this branch. In fact, it looks like master might perform a bit better. I suppose it could be that the compiler is already applying SIMD optimizations to the loops.

argilo:
Have you been able to measure any performance improvement on your system? We probably shouldn't merge this until it's clear that there's a benefit.

vladisslav2011:
> Have you been able to measure any performance improvement on your system?

I see small improvement on Intel Atom N270 system.
128k FFT, 10 fps update rate, maximum zoom, no demodulator
master ~54.5% CPU consumed by main thread
this branch ~53.7% CPU consumed by main thread
Less than 1% difference. I think GCC autovectorization works very well on x86.
It may be better to test on ARM, but I do not own a GUI-capable ARM system.
There shouldn't be much improvement as application of the window takes much less time, than actual fft calculation and even less time, than rendering waterfall/spectrum.

argilo:
Since the performance difference was so small, I made more precise measurements by placing `std::chrono::high_resolution_clock` around just the lines that were changed, and averaged over many executions.

On my laptop, which has a i7-8550U CPU, I got the following results:

rx_meter_c, used with wideband FM (240 ksps):
* before: 0.232 ms
* after: 0.048 ms

This gives a savings of 0.184 ms per execution. At 10 executions per second, the total reduction in CPU utilization should be about 0.184%.

rx_fft_c, used with 1048576-point FFT:
* before: 1.965 ms
* after: 1.972 ms

Presumably the compiler already optimized this simple loop to use SIMD instructions.

These results would explain why the performance difference was not visible using top.

argilo:
Since there does appear to be a (very small) performance improvement, I'll merge this once GitHub Actions is working again and CI passes.

In the future, it would be very helpful if you would benchmark potential performance improvements so that it is easier to decide whether they are worthwhile.

argilo:
For reference, I used more-or-less the code from this answer: https://stackoverflow.com/questions/22387586/measuring-execution-time-of-a-function-in-c

PR #1050, opened 2022-01-08
Don't skip the first RDS message in the queue

argilo:
`MainWindow::rdsTimeout` always throws away the first message waiting in the RDS queue. Here I've fixed that be re-ordering its loop.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.15.3: In progress...
+
+     FIXED: RDS receiver skips some messages.
+
+
     2.15.2: Released January 8, 2022
 
        NEW: Save & restore center frequency when playing I/Q files.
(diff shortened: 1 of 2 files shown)


argilo:
Before this change, `num = -1` was passed into `DockRDS::updateRDS` on the last iteration of the loop. It doesn't matter that we're not doing that anymore, because it treats `-1` as an invalid value and ignores it.

vladisslav2011:
I have tested this branch and found no issues.
I don't see any significant difference in RDS update rate. That's because `rx_rds_store::get_message` does not clear the buffer if there are no messages in the queue and a message always gets passed to the `uiDockRDS->updateRDS`.
I'll test on my slow test system to see, if there would be any difference in case the main thread is busy calculating the FFT and rendering the waterfall/spectrum all the time.

vladisslav2011:
I do not see any significant difference even after adding commits from https://github.com/gqrx-sdr/gqrx/pull/1051 and making gqrx able to process 1M FFT on Intel Atom N270 (and adding debug message to the `MainWindow::rdsTimeout()` to monitor `num`).
I've seen several bursts of 7 messages (`num`=6) while switching FFT sizes and constant stream of 1 message in the queue (`num`=0) when the UI is not touched.
In any case merging this should not harm anything.

PR #1052, opened 2022-01-10
Fix buffer overruns in AFSK1200 decoder

argilo:
While testing #1042 with valgrind, I noticed buffer overruns in the AFSK1200 decoder. They are present on the master branch as well.

The problem occurs because `CAfsk12::demod` expects its input buffer to have `length + CORRLEN` samples. `Afsk1200Win::process_samples` fails to ensure that the extra `CORRLEN` samples are present on the first invocation, leading to a buffer overrun in `CAfsk12::demod`. Additionally, `Afsk1200Win::process_samples` underruns `buffer` whenever `length` is less than `CORRLEN`, which can happen if the AFSK1200 window is opened before the DSP is turned on.

I've corrected these problems by initializing `tmpbuf` with `CORRLEN` zeroes, and removing `length` samples from the beginning of `tmpbuf` after each invocation.

I also replaced `QVarLengthArray` with `std::vector`, since the Qt documentation says:
> QVarLengthArray is a low-level optimization class that only makes sense in very specific cases. It is used a few places inside Qt and was added to Qt's public API for the convenience of advanced users.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -2,6 +2,7 @@
     2.15.3: In progress...
 
      FIXED: RDS receiver skips some messages.
+     FIXED: Buffer overruns in AFSK1200 decoder.
 
 
     2.15.2: Released January 8, 2022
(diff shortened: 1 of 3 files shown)


argilo:
> smples?

Ah yes, there is a typo there. Fixed in e7557149c53d188d4f6ad077e83e9d08aac9152a.

PR #1054, opened 2022-01-10
CMeter: increase resolution to 0.1dB

vladisslav2011:
To match the resolution of the squelch setting.


argilo:
Thanks. This makes sense and works well, so I'll merge it.

I suppose it may also make sense to change `m_Siglevel` to a float, since the extra resolution could be useful on high-DPI displays.

Issue #1055, opened 2022-01-10
rx_agc_cc sometimes eats the whole CPU core


vladisslav2011:
When the squelch is closed, rx_agc_cc tries to consume the whole CPU core trying to normalize zeroes to full-scale...

Conditions:
gqrx built from git master, GNU Radio 3.7.11, Ubuntu 18.04
Tested on debian testing GNU Radio built from git master - same thing.

Steps to reproduce:
1. Start gqrx
2. Start the DSP
3. Tune to some unused frequency
4. Select AM mode
5. Set squelch to 'auto'

Expected behavior:
The rx_agc_cc thread should consume ~2..7% of CPU core time

Observed behavior:
The rx_agc_cc thread consumes up to 80-90% of CPU core time

Squelch closed:
![old_squelch_closed](https://user-images.githubusercontent.com/17007837/148844667-6caee221-466d-477f-9cd5-b280ae14e195.png)

Squelch open:
![old_squelch_open](https://user-images.githubusercontent.com/17007837/148844668-8cd3d6f1-0e8c-4f19-8a50-747303fc2eb7.png)

PR is ready.

PR #1056, opened 2022-01-10
New rx_agc implementation 1/2

vladisslav2011:
New AGC implementation with less complexity.
CPU load does not vary depending on input.
New controls, that should be easy to understand:
Maximum gain - maximum AGC gain in auto mode.
Target level - a level, to normalize the signal to in dB.
Attack - attack time in ms. Time to go from max gain to 0 dB (worst
    case). Defines the size of the delay buffer and, as
    a result, sets audio delay, introduced by the block.
Decay - decay time in ms. Time to go from 0 dB gain to maximum gain
    (worst case).
Hang - hang time in ms. Time to wait before increasing the gain after
    the peak.

GUI and setting are updated.

The next step is to put the AGC block after the demodulator to make it
usable in FM mode and to remove useless audio_gainX blocks.


argilo:
There are a lot of changes here. Unfortunately I don't think I'll have time to review such a large changeset soon. Is there any way to reduce the number of changes or split this into smaller pieces?

argilo:
Also, did you find out what the root cause of #1055 was? Is there a smaller fix we can apply in the interim, until there's time to review a major change to the AGC?

vladisslav2011:
Testing results.
Squelch open:
new_squelch_open
Squelch closed:
new_squelch_closed
Quick test with airband IQ recording have shown good performance, no noticeable bugs (needs more testing) and predictable reaction to controls.

OK. CI failed. Apple clang does not like exp10f. Moving to draft.

vladisslav2011:
> Also, did you find out what the root cause of #1055 was?

This https://github.com/gqrx-sdr/gqrx/blob/master/src/dsp/agc_impl.cpp#L238-L243

But I do not like the overall implementation.
Complex magnitude calculation for example: https://github.com/gqrx-sdr/gqrx/blob/master/src/dsp/agc_impl.cpp#L217-L220 - what the...
I do not like it's placement too. It's pointless to put the AGC before FM demodulator. It is much better to put the AGC after the demodulator. It would adjust for deviation differences and make listening to NFM signals more comfortable too.
+ It would be possible to remove audio_gain# blocks as the AGC would prevent clipping if enabled and emulate gain blocks when disabled. Less blocks - less CPU usage, less CPU usage - higher input rate on RPi and similar weak devices.

argilo:
Is there a reason not to just use one of the AGC blocks built into GNU Radio?

vladisslav2011:
I've replaced the `exp10f` with `powf(10, `. This should make clang happy. I'll reopen when CI ends.

vladisslav2011:
> Is there a reason not to just use one of the AGC blocks built into GNU Radio?

It may be possible. I have plans to test them too. Even make it possible to switch between different AGC implementations (not for upstream though).
But there is no GNU Radio stereo AGC block... So, putting a GNU Radio block after the demodultor would either require interleaving stereo to sort of "IQ" and deinterleaving later and separate gain blocks for disabled mode. Or using 2 AGC block and synchronizing gains... Not possible. Interleaving/deinterleaving would increase CPU usage and introduce inaccuracies. Using 2 separate block would almost kill the stereo.

This PR is the 1st step to proper AGC. The second step would include making this block stereo 2in/2out and moving it after the demodulator. The audio_gain# blocks would be useless in this case and may be removed, switching the GUI control to manual_gain of the AGC block.

All my PRs are small steps on the path to the attempt of #946 as I don't like LouDou's solution and want to try it myself. But I have to fix bugs first (iq_tool to make reproducible tests at high sample rates, reduce CPU usage to make it possible to use more demodulators at one in the future, etc.)

argilo:
> But there is no GNU Radio stereo AGC block...

I'm a bit confused. Why is stereo AGC needed?

vladisslav2011:
Different FM stations have different deviation. Where if WFM deviation setting?
The AGC would scale all stations to full scale and prevent clipping. If somebody would be uncomfortable with this, then there is 'AGC Off' setting and manual gain to emulate current behavior.

argilo:
Hmm, I'm not convinced this is a good idea. There are FM broadcast standards that define deviation. If we need to increase deviation in a bit in Gqrx to prevent clipping, then we should probably to do that instead. AGC will result in audio gain going up during silent periods, which seems undesirable.

argilo:
I do agree that AGC prior to FM demodulation is a waste of CPU cycles, so we could probably remove that.

vladisslav2011:
> AGC will result in audio gain going up during silent periods, which seems undesirable.

There is `hang` setting to prevent this from happening.
I know, that there are standards, but I see different stations broadcasting with different settings. Some use ~75kHz deviation, others use ~200kHz deviation... That's real world.

argilo:
> Where if WFM deviation setting?

If there are different standards in different parts of the world, we can add a setting for that. There is a setting for NBFM already.

vladisslav2011:
> There is a setting for NBFM already.

This does not help when one correspondent on 2M uses motorola @25kHz deviation and other uses baofeng with 7kHz real deviation. It's hard to listen to the conversation...

argilo:
OK, maybe there is some value in post-demod AGC for FM in an environment where nobody follows the rules. I haven't experienced that problem here.

vladisslav2011:
In my location some hams are always whispering far away from the microphone, some are always screaming and some are constantly switching between screaming and whispering... Funny to hear first time and not funny if you want to copy everything.

I'll do the second part... maybe tomorrow. Give it a try or leave it here until some free time would be available...

argilo:
:+1:

Sounds good. I'll look at this when I get a chance. Gqrx is a spare-time project, so I don't always have a lot of time for review. That's why smaller changes are more likely to get reviewed, tested and merged quickly.

vladisslav2011:
I've made some bugfixes and adjustments.
It is ready for testing.
Suggestions and bug reports are welcome.

src/applications/gqrx/receiver.cpp
```diff
@@ -853,13 +873,40 @@ receiver::status receiver::set_agc_decay(int decay_ms)
...
+}
 
-    return STATUS_OK; // FIXME
+/** Set audio mute. */
+receiver::status receiver::set_mute(bool mute)
+{
+    if (d_mute == mute)
+        return STATUS_OK;
+    tb->lock();
+    if (mute)
```
sultanqasim, line 891: This tb connect/disconnect logic needs to be wrapped with a check making sure `d_demod != RX_DEMOD_OFF`. Otherwise, the application will crash if you try to mute it while the demodulator is off.

sultanqasim:
Some issues/bugs I've encountered with this change:

- After unmuting, it takes around 2.5 seconds for audio to come back
- Pressing mute while recording causes the recording to end, and the UI thinks the recording is ongoing when it actually ended
- After pressing record, there's a 2-2.5 second delay before the recording starts, and audio playback also stops while waiting for recording to start. Audio playback also stops for 2.5 seconds when ending recording.

Tested on Fedora 38 and Mac OS 13.

vladisslav2011:
Thanks for testing.
Some samples are lost due to GNU Radio regression https://github.com/gnuradio/gnuradio/issues/6105
So it should be fixed on GNU Radio side.
It may be fixed by applying https://github.com/gnuradio/gnuradio/pull/6108/commits/3053725b8297e851f1c4b6e9b3c64f63e5a6c61a if GNU Radio is built from source...

src/applications/gqrx/receiver.cpp
vladisslav2011: Thanks.
It looks, like I have lost some changes during rebase. I'll fix this and update the PR.

vladisslav2011: Updated.

willcode:
I wish I understood the effects of making this change in GR better, wrt to tag propagation. I wonder if an optional param in lock/unlock would be useful in some way, to preserve the current behavior by default.

PR #1058, opened 2022-01-11
Check for demodulator before starting AFSK1200

### LINKED ISSUE #1057: Crash when turning on AFSK1200 decoder without demodulator
Gqrx crashes if AFSK1200 decoder is started without demodulator which produces audio. Some checks needs to be added. I will submit a PR shortly.

rgerganov:
AFSK1200 decoder doesn't work without demodulator. Make sure there is
selected demodulator before starting AFSK1200. Also make sure we stop
AFSK1200 when demodulation is turned off.

Closes #1057

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ The following people and organisations have contributed to gqrx:
 * Pavel Milanes, CO7WT
 * Pavel Stano
 * Phil Vachon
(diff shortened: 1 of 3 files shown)


argilo:
Thanks! I tested this, and it works well. I'll merge once CI passes.

PR #1060, opened 2022-01-14
New rx_agc implementation (2/4)

vladisslav2011:
Switch to float and move rx_agc to the output of demodulator.
Works for FM modes.
No more need to adjust gain while tuning between WFM stations.

Default presets may need to be adjusted.
New 'WFM' preset may be added.
Suggestions welcome.


vladisslav2011:
CI failed. Converting to draft.

vladisslav2011:
CI successful. Reopening.
Removal of audio_gain# is moved to separate PR https://github.com/gqrx-sdr/gqrx/pull/1061 to keep one major change per PR.
Recommended settings:

AM/SSB
Max gain: 80 dB or more
Preset: Fast or Medium

NFM
Max gain: 10-20 dB depending on input signals frequency deviation variations
Preset: Medium or Slow

WFM:
Max gain: 10-12 dB should be enough to generate soft clickless transitions 
Preset: User
Attack: 200 ms
Decay: 5000 ms
Hang: 500 ms or more

Notes:
The rx_agc at any settings should be able to prevent clipping even at 0 dB target level, but, for best results, Max gain should be set to minimum required value.
Introduced audio delay is equal to the Attack setting (200 ms should be enough even in WFM mode)
As the AGC is operating with fixed slope, actual attack/decay times are less than corresponding settings. Attack/decay settings define maximum slope times. 

S-curved slope generation may be added later (this would require more CPU time either in the reconfuguration function (and some memory to store precalculted curves) or in the work function (better to avoid).

Suggestions and bug reports are welcome.

PR #1062, opened 2022-01-14
Correctly name a sniffer_f block

vladisslav2011:
There are two rx_fft_f threads when the sniffer is enabled. One of these
threads is really sniffer_f thread.
Fix this by naming the sniffer_f block correctly.


argilo:
Good catch. I had a look at the other blocks, and they seem to be named correctly.

PR #1063, opened 2022-01-14
Disconnect iq_swap when not in use...

vladisslav2011:
...to reduce cpu load.


stefanino-ch:
@vladisslav2011 

On top of the main branch I integrated the following pull requests: 
1142
1147
1026
1010
1011
1013
1016
1063

To get 1063 to compile and run I had to change a method signature in src/applications/gqrx/receiver.h & .cpp

There are several method calls like
`set_demod(d_demod, true);`

These will not fit together with the default parameters of this method. 

```
/* old
    status      set_demod(rx_demod demod,

                          enum file_formats fmt = FILE_FORMAT_LAST,

                          bool force = false);
    */
    status      set_demod(rx_demod demod,
                          bool force = false,
                          enum file_formats fmt = FILE_FORMAT_LAST);
```

After changing the signature and updating in receiver.cpp in some of the method calls where all three parameters are passed to switch the parameters accordingly all seems to work. 

Not sure if this is a merge issue or from where this comes.

PR #1064, opened 2022-01-14
Restore input decimation after IQ playback

vladisslav2011:
While working on https://github.com/gqrx-sdr/gqrx/pull/1063, I have noticed, that input decimation does not get reset to 1 when IQ playback is started, nor gets reset back to configured value, when IQ playback stops.
Fix this by setting input decimation to 1 when starting IQ playback and resetting it back to configured value, when IQ playback stops.


stefanino-ch:
@vladisslav2011 

First of all, thank you for all your improvements!!!

I made a build with the follwing PRs:
1026	Iq tool add sampling rate and frequency GUI inputs 
1010	Iq tool center freq lock 
1011	Iq tool gui fixes 
1013	Iq tool formats 
1016	Iq tool buffered 
1063	Disconnect iq_swap when not in use... 
1064	Restore input decimation after IQ playback 

That was quiet challenging for me. I did the first time an integration of several PRs. Therefore some of the issues below could very well be based on an integration mistake done by myself. 

Now during testing of all my integration work I found some potential improvements related to all the PRs. I gonna add all in here, even if it might be related to another PR of the list above. 

**Waterfall does not stop after playback**
If the playback is at the end of the file the progress slider stops, but the waterfall don't. 

**HW Frequency is added into frequency display**
In the I/Q Recorder window the starting frequency is displayed. The value shown here is Frequency + Hardware freq. Maybe here should be just the bare Frequency without the HW part be displayed?

**Update frequency display in panadapter during playback**
I will use gqrx for I/Q recording in a wide frequency range. Therefore I do have a script which changes during recording the frequency. During playback the frequency shown in the panadapter is the one on which the recording has been started. I don't know if this is possible, but if the I/Q recording frequency is changed during recording, it would be a really nice feature if the panadapter would reflect this also during replay. 

**Pause button for playback**
Would it be possible to add a Pause button in the I/Q window? This button should pause the replay including panadapter and waterfall display. If during a replay only panadapter and waterfall is used for signal analysis, a pause button would be a helpful feature.

**Sizing/ layout issue**
I resized the I/Q window to be as small as possible. During replay, the Buffer indication jumps left and right depending of the size of the numbers shown for the Pos: indication. Moving the Buffer indication to its own line, or on the same line as the Format: indication would most probably avoid this flickering. 

**Jumping buttons**
In I/Q window, depending on record or playback mode the height needed to display the GUI elements below the filelist differs. This leads to vertically jumping buttons if between recording and playback is switched. This makes the windown somehow nervous. I think it would be a nice improvement if this jumping can be avoided. Something like fixed height for the filelist, a panel below the filelist holding the buttons, sliders, ... and at the final end a vertical spacer.  

**No Format indication during replay**
During replay only Format: is displayed without any following value indication.

vladisslav2011:
Many thanks for your feedback.
It is very pleasant to see that there is somebody, interested in improvements to this project.

>Waterfall does not stop after playback

GNU Radio does not have a method to test if the flowgraph is still running... Or I just have not found how, to detect the flowgraph stopping on it's own.

>HW Frequency is added into frequency display

I don't really understand, what do you want. There is no frequency display in the IQ Tool. The center frequency input should define the center frequency during playback if the file name is not parsed correctly. I'll retest and update #1026 to block both inputs if the filename is parsed and unblock if the filename does not look like Gqrx IQ recording.

>Update frequency display in panadapter during playback

That's not possible. The only purpose of keeping the ability to change a hardware frequency during IQ recording is satellite doppler compensation. In any other case the user is expected to stop IQ recording, tune to another center frequency, start IQ recording.

>Pause button for playback

Not required. You may just stop the DSP using main start-stop button on a toolbar. I've even made use of this state to implement "IQ exploration mode". See demonstration here https://www.youtube.com/watch?v=dwvVLcLEIqw . The branch with this feature (and many other unreleased features): https://github.com/vladisslav2011/gqrx/tree/fft_channelizer . I have not opened a PR as this feature depends on a modified buffered `file_source` block and @argilo is strongly against this change.

>Sizing/ layout issue
>Jumping buttons

I have encountered this too. Buttons/inputs are covering the slider when the window height becomes too small. I think, we should limit minimum height of IQ Tool window. Or just hide file list/buttons/inputs and keep the slider starting with some height.

>No Format indication during replay

Is it really required? The format is expected to be guessed from a filename during playback. Just look at the selected file and see the format suffix. Hmm. May be a visual bug as I have not tried to merge #1026 on top of other IQ Tool-related PRs.

stefanino-ch:
I will split by topic, this schould make the discussion easier....

**Waterfall does not stop after playback**
At if the replay arrives at the end of a file
- the panadapter stops
- the progress slider stops
- just the waterfall runs on

Would it not be possible to stop the DSP at file end, e.g. by imitating a DSP Start/Stop button press?

stefanino-ch:
**Sizing/ layout issue
Jumping buttons**

From my point of view increasing min height will not solve the problem. 

In a PyQt GUI I would try the following (not shure if this is working in Qt as well): 
- add a widget below the filelist
- make this widget a fixed height 
- into this widget place all the elements which are needed for record and replay
- the very  last item of all those single elements would be a vertical spacer

Like this there would be more or less free space at the bottom of the window, which is from my point of view acceptable.

stefanino-ch:
**HW Frequency is added into frequency display**
The frequency indicated in the bottom part of the I/Q window upon selecting a file is parsed from the filename. Now I understand the functionality. 

Then my issue turns more into the creation of the filename. If I set a receiving freq of 105.8 MHz, the Hardware freq is changed automatically. What will be used for the filename is a combination of my freq input and the hw freq. Therefore the freq indicated in the filename (and later displayed in the I/Q window) is not what I originally entered. 

From my point of view a minor issue and not related to the work you did. Might even be based on some missing know how on my side in regards of freq and hw freq relations.

stefanino-ch:
**Pause button for playback**
Cool, never thought about the DSP Button to have an influence on this. 

But the functionality as it is now is somehow asymetrical. 
- Play in I/Q window starts a replay from begin
- Again Play in I/Q window stops a replay and resets immediately panadapter

- Play in main window starts/ continues a replay
- Play again in main window will just pause but not reset panadapter 

Would it be possible to make behaving the play button in I/Q the same as in the main window?
- If a replay is running and the user clicks Play again, replay is just stopped but no jumping to the start of the file and no reset of the panadapter

OR alternative solution
- change the functinality of the Play button in I/Q to something like "Replay mode"
- file selection remains the same
- replay will the only be controlled by the play button in the main window
Potentially this will avoid a lot of status and button syncing between windows?

stefanino-ch:
**Progress slider in I/Q**
While plying today I've found two issues with the progress slider in I/Q. 

**Slider is editable**
As it is now the slider allows the user to move it. This somehow brought me to the idea it would be possible to move within the replay with the slider. But this does not work. From my point of view the slider should be changed to read only without any edit possibility for the user

**Jumping slider at replay start**
How to reproduce
- Start a replay (Play button in I/Q)
- Stop the replay (Play button in I/Q)
- Start the replay again (Play button in I/Q) 
- The slider will first show a too hight value, and jumps then back to the start indicating where the replay really is

For me it looks like there is a value in the progress calculation not updated before the first calculation is done?

vladisslav2011:
> Would it not be possible to stop the DSP at file end, e.g. by imitating a DSP Start/Stop button press?

No. (At least I don't know, how to make sure, that all sample are consumed consumed by the sink when the file ends.)
Another reason to not stop the DSP when the file ends: When the file ends, it should be possible to click the slider to restart playback from some point.

> there would be more or less free space at the bottom of the window

I prefer the file list to grow when a window is resized. Vertical spacer is a waste of precious screen space here.

> Therefore the freq indicated in the filename (and later displayed in the I/Q window) is not what I originally entered.

You may reset the offset frequency to zero first (with mouse right click), then tune to your desired center frequency.
Another option is to convert a center frequency indicator to an input widget and update tuning logic...

> Would it be possible to make behaving the play button in I/Q the same as in the main window?
> If a replay is running and the user clicks Play again, replay is just stopped but no jumping to the start of the file and no reset of the panadapter

No jumping to the start of the file is possible. Not resetting the panadapter is not as center frequency/offset frequency/bandwidth are restored and the panadapter should be updated to reflect these changes.

> OR alternative solution
> change the functionality of the Play button in I/Q to something like "Replay mode"

@argilo prefers current logic (and I agree with him). It is better to ask him, why current logic (IQ Tool Play button to start/stop playback, DSP button to pause/start playback) is better.

> **Slider is editable**
> As it is now the slider allows the user to move it. This somehow brought me to the idea it would be possible to move within the replay with the slider.

It should be possible to move within the replay with the slider. I'm doing it every time, I'm using the IQ Tool to play a file. If it does, not work, then you have broken something during rebase...

> **Jumping slider at replay start**

The slider is updated by the timer once per second. The timer is started when playback starts, so the first update is delayed by 1 second.

PR #1068, opened 2022-01-15
Prevent audio delay on lifftw plan generation

vladisslav2011:
When ~/.gr_fftw_wisdom is missing, and gqrx is run for the first time,
there is noticeable audio delay, that grows while switching the FFT to
larger sizes. The audio is skipping too.
This patch prevents this from happening by unlocking data mutex and
locking separate FFT mutex for the time of libfftw plan generation.

### DIFF
--- a/src/dsp/rx_fft.cpp
+++ b/src/dsp/rx_fft.cpp
@@ -126,12 +126,15 @@ void rx_fft_c::get_fft_data(std::complex<float>* fftPoints, unsigned int &fftSiz
     apply_window(d_fftsize);
     lock.unlock();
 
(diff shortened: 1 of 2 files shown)


argilo:
I'm not sure the second mutex is needed, because all methods other than `work` should be called by the UI thread.

I think we can simply remove the lock from `rx_fft_c::set_params()` and `rx_fft_f::set_fft_size()` since they no longer access the circular buffer.

argilo:
I've got an alternate fix ready in #1069.

argilo:
Closing, since #1069 was already merged.

PR #1069, opened 2022-01-15
Don't lock the buffer mutex while changing FFT size

argilo:
Replaces #1068.

Fixes the issue described in https://github.com/gqrx-sdr/gqrx/pull/1051#issuecomment-1013752360.

There's no need to lock the mutex in `rx_fft_c::set_params` or `rx_fft_f::set_fft_size` anymore, since they no longer access the circular buffer after #1051 and #1066 were merged.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.15.4: In progress...
+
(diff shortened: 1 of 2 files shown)


argilo:
Tested, and this appears to work well.

PR #1071, opened 2022-01-19
Clean up CMake build

argilo:
This pull request makes two improvements:

1. Unconditionally run `find_package(Volk)`, since VOLK is a direct dependency of Gqrx.
2. Move legacy code that is only needed for GNU Radio 3.7 into conditional blocks. This will make cleanup easier when support for GNU Radio 3.7 is eventually dropped.


argilo:
CI passes, and the builds work well. Merging.

Issue #1073, opened 2022-01-21
macOS Monterey 12.1 UI Bugs


maxbundscherer:
Hi,

I am using qqrx 2.15.5 on macOS Monterey 12.1 UI. Unfortunately, there are some very annoying UI bugs - after updating to monterey. I have tried different gqrx version on different macs, but the issue remains.

The Pand.dB and Wf.dB slider in fft-settings tabs don't work correctly. Mostly the sliders set lower and upper levels to the same level and I cant move the upper and lower independent of each other. Only restarting gqrx helps and I can set the levels 2-3 times after having the same issue again.

Maybe some of u know what the problem is or can tell me where I can fix it.

Thank you!

<img width="634" alt="Bildschirmfoto 2022-01-21 um 15 29 44" src="https://user-images.githubusercontent.com/24506308/150545258-13bed661-0801-4827-ae0d-6282ad193a6a.png">

maxbundscherer:
@LongnoseRob  - thank you for your answer. I have tried to click on the lock button, the bug remains (see attached screenshot). Yeah, it is a macOS only bug - on Linux, I have no problem.

<img width="745" alt="Bildschirmfoto 2022-01-22 um 11 38 23" src="https://user-images.githubusercontent.com/24506308/150635248-c843c159-1bdc-4e3e-ac19-0ab0d2efb0f7.png">

argilo:
Thanks for the bug report. I was able to reproduce this in Gqrx 2.11.5 on macOS 12.1, so at least I know that it's not due to recent changes in Gqrx.

I also once on Linux observed the behaviour where the range extends all the way to the left even though no slider handle is there. I haven't been able to reproduce it since.

The ctkRangeSlider widget was copied into Gqrx back in 2016, and I suppose it could have bugs that came up since then.

maxbundscherer:
@argilo thank you for the information!

argilo:
It seems like there are bugs with the QSlider widget in Monterey: https://bugreports.qt.io/browse/QTBUG-98093

The ctkRangeSlider widget is based on QSlider, so it would probably inherit those bugs.

Gqrx is built with [Qt from Homebrew](https://formulae.brew.sh/formula/qt@5), which is presently at version 5.15.2 (released November 2020, well before Monterey). QSlider fixes may have been backported to later 5.15 releases, but I wouldn't be surprised if Homebrew never picks them up.

[...]

PR #1079, opened 2022-01-24
Fix deprecation warnings reported by Qt 5.15

argilo:
I'd like to update Gqrx to Qt 6, because development on Qt 5 has now stopped. Updating will also make it possible to solve the UI bug reported in #1073.

According to the [Qt 6 porting guide](https://doc.qt.io/qt-6/portingguide.html), the first step is to remove use of deprecated API, as reported by Qt 5.15. I've done that here, with each commit addressing one source of deprecation warnings.

I'll merge this once CI is happy and I've had a chance to test the parts that have changed.


argilo:
CI passes, and the affected areas of Gqrx still seem to work well.

PR #1090, opened 2022-01-26
Build against Qt6 if it is available

argilo:
This is a continuation of the Qt modernization work started in #1079, which should eventually allow #1073 to be solved.

Here I've updated the CMake build to use Qt6 if it is available. I've tested this on Arch Linux, where Qt6 is available in the package manager (`pacman -S qt6-base qt6-svg`), and Gqrx works well. I haven't spotted any bugs yet.


LongnoseRob:
works also fine on ArchlinuxArm

argilo:
That's great! If you notice any bugs in the Qt6 build, please let me know.

Issue #1091, opened 2022-01-26
Regression Compilation broken


vladisslav2011:
Compilation is broken on my 'weak' test system (Atom N270) since https://github.com/gqrx-sdr/gqrx/pull/1088/commits/5e923578e7f50cfca36f1ede42d12f745a07ea27

Steps to reproduce:
1. Checkout master
2. Configure and make

Expected results:
gqrx builds and runs

Observed results:
```
make -j36
[  1%] Automatic moc for target gqrx
[  1%] Built target gqrx_automoc
[  2%] Building CXX object src/CMakeFiles/gqrx.dir/applications/gqrx/remote_control.cpp.o
/home/vlad/gqrx-my/src/applications/gqrx/remote_control.cpp: In member function 'void RemoteControl::acceptConnection()':
/home/vlad/gqrx-my/src/applications/gqrx/remote_control.cpp:174:21: error: 'class QHostAddress' has no member named 'isEqual'
         if (address.isEqual(QHostAddress(allowed_host)))
                     ^
src/CMakeFiles/gqrx.dir/build.make:255: recipe for target 'src/CMakeFiles/gqrx.dir/applications/gqrx/remote_control.cpp.o' failed
make[2]: *** [src/CMakeFiles/gqrx.dir/applications/gqrx/remote_control.cpp.o] Error 1
CMakeFiles/Makefile2:135: recipe for target 'src/CMakeFiles/gqrx.dir/all' failed
make[1]: *** [src/CMakeFiles/gqrx.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

Conditions:
```
dpkg -l |grep qt5
ii  libqt5concurrent5:i386                5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 concurrent module
ii  libqt5core5a:i386                     5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 core module
ii  libqt5dbus5:i386                      5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 D-Bus module
ii  libqt5gui5:i386                       5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 GUI module
ii  libqt5network5:i386                   5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 network module
ii  libqt5printsupport5:i386              5.5.1+dfsg-16ubuntu7.7                   i386         Qt 5 print support module
[...]

PR #1092, opened 2022-01-26
Fix compilation on Qt < 5.8

vladisslav2011:
Use QHostAddress::operator==(const QHostAddress &) instead of isEqual(const QHostAddress &)on Qt < 5.8

Fix regression introduced in 5e92357.
Close https://github.com/gqrx-sdr/gqrx/issues/1091

### DIFF
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -171,7 +171,11 @@ void RemoteControl::acceptConnection()
 
     for (auto allowed_host : rc_allowed_hosts)
     {
+#if QT_VERSION < QT_VERSION_CHECK(5, 8, 0)
+        if (address == QHostAddress(allowed_host))
+#else
         if (address.isEqual(QHostAddress(allowed_host)))
+#endif
         {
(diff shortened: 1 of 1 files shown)


vladisslav2011:
Another possible fix is to convert both addresses to IPv6:
```
diff --git a/src/applications/gqrx/remote_control.cpp b/src/applications/gqrx/remote_control.cpp
index ef98839..04a4b3e 100644
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -171,7 +171,7 @@ void RemoteControl::acceptConnection()
 
     for (auto allowed_host : rc_allowed_hosts)
     {
-        if (address.isEqual(QHostAddress(allowed_host)))
+        if (memcmp(address.toIPv6Address().c, QHostAddress(allowed_host).toIPv6Address().c, 16) == 0)
         {
             connect(rc_socket, SIGNAL(readyRead()), this, SLOT(startRead()));
             return;
```

argilo:
As mentioned in #1091, Gqrx no longer supports Ubuntu 16.04 and I'd rather not carry around extra code to support it. I'd suggest upgrading your system to a supported version of Ubuntu.

PR #1094, opened 2022-01-27
Build against Qt6 on macOS

### LINKED ISSUE #1073: macOS Monterey 12.1 UI Bugs

argilo:
Fixes #1073.

Now that Gqrx has been updated with Qt6 support (#1079, #1081, #1082, #1083, #1084, #1085, #1086, #1087, #1090), we should be able to switch over the macOS build.

For some reason, macdeployqt fails with a bunch of errors on the first run:
```
ERROR: Cannot resolve rpath "@rpath/QtGui.framework/Versions/A/QtGui"
ERROR:  using QSet("/Users/runner/work/gqrx/gqrx/lib")
ERROR: Cannot resolve rpath "@rpath/QtCore.framework/Versions/A/QtCore"
ERROR:  using QSet("/Users/runner/work/gqrx/gqrx/lib")
...
```
Running it a second time works around that problem.

Also, macdeployqt fails to copy in the libbrotlicommon.1.dylib dependency, so I've added a step to manually copy it over and sign it.

I've added reminders to remove both of these workarounds once macdeployqt is updated or a better solution is found.

I'd appreciate help testing the new build, which is available at the bottom of this page: https://github.com/gqrx-sdr/gqrx/actions/runs/1758484287

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew update
-          brew install airspy airspyhf boost dylibbundler gnuradio hackrf libbladerf librtlsdr libserialport portaudio pybind11 uhd
+          brew install airspy airspyhf boost dylibbundler gnuradio hackrf libbladerf librtlsdr libserialport portaudio pybind11 uhd qt@6
           brew tap pothosware/homebrew-pothos
           brew install soapysdr soapyremote
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew update
-          brew install airspy boost gnuradio hackrf libbladerf librtlsdr pybind11 uhd
+          brew install airspy boost gnuradio hackrf libbladerf librtlsdr pybind11 uhd qt@6
 
           cd /tmp
           git clone git://git.osmocom.org/gr-osmosdr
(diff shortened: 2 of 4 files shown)


kgarrels:
BTW, this also works for me with qt6 from macports, very nice.

Thanks,
kai

argilo:
I haven't heard any negative feedback, so I'll merge this. If any bugs come up, we can fix forward.

argilo:
I just noticed a bug: the dropdowns to change the FFT size, FFT rate, and antenna input do not work in Qt6. This is because the obsolete `currentIndexChanged(const QString &text)` signal has been removed from `QComboBox`. I'll change them to use `currentIndexChanged(int index)`.

argilo:
Fixed. A new build will be available here shortly: https://github.com/gqrx-sdr/gqrx/actions/runs/1769237597

maxbundscherer:
@argilo - thanks for the great work and quick fix! I use the build https://github.com/gqrx-sdr/gqrx/actions/runs/1769237597 on my intel macbook and macbook m1.

PR #1096, opened 2022-01-29
New rx_agc implementation (4/4) Bugfixes and performance

vladisslav2011:
Move common code from nbrx/wfmrx to receiver_base
Move all agc processing into block class rx_agc_2f and eliminate CAgc class
Remove CurrentGainDb and move processing to get_current_gain
Rename SetParameters to set_parameters
Remove ProcessData and move all processing to work
Fix tag propagation setting sample delay
Workaroung GNU Radio < 3.8 tag propagation bug: manually calculate new tag offsets
Small cleanup
Set audio mute by disconnecting audio_snk
Switch to using history to improve performance

Flowgraph structure (master):
gqrx-master

Flowgraph structure (this PR):
gqrx-rx_agc

Unfortunately github does not play well with SVG images. Right-click the image and choose "Save image as...", then open saved image to get it rendered correctly.


csete:
The whole purpose of the AGC is to provide demodulators with a stable signal under fading conditions. This is rather important for AM, in fact, one of the distinguishing features of GQRX for shortwave listeners.

vladisslav2011:
Hello.
Thanks for a wonderful application. Gqrx was the one SDR application, that actually worked, when I tried to use chinese TV tuner as a SDR frontend for the first time.

I agree, that in case of a hardware AM receiver the purpose of the AGC is to provide silicon/germanium diode with enough voltage, so it will operate as a semi-linear amplitude detector. But that's not the case if demodulation is done in software. complex_to_mag block does not have a minimum detection threshold, nor non-linear region. We can put AGC after the demodulator and get exactly the same result.
This AGC implementation gives almost the same results as the original implementation when the DCR option of the AM/AMSync demodulator is disabled. So the result mostly depends on placing the AGC before or after DC-removing high-pass filter. If we place it before, it takes the carrier level into account and outputs too quiet sound as the modulation is never 100%. If we place it after, it does not take the carrier level into account and normalizes the signal to full scale. At least this implementation works this way.
In case of SSB both implementations are giving almost similar results, but original implementation makes it hard to get non-clipped full-scale sound out of the receiver. I'm getting either clipped scratching sound or clipped too quiet sound with the original AGC implementation. This may be a bug...
In case of FM demodulator, putting the AGC before it does not help at all as the ideal FM demodulator (GNU Radio tries to be as ideal, as possible) does not take the magnitude into account and operates on a phase only. Putting the AGC after the FM demodulator helps when different stations on the same frequency are using different deviation settings. It normalizes the sound and makes listening more comfortable. Commercial WFM stations are sometimes using different deviation settings too. Even the same station is sometimes transmitting with different deviation on different days in my region. So setting the maximum AGC gain to a low value (0...5db), to not amplify the noise in pauses, may help too. See the discussion here: https://github.com/gqrx-sdr/gqrx/pull/1056

PR #1124, opened 2022-04-29
Improving readme content

ei8fdb:
I've made the following improvements to the readme:

- expanded installation instructions, broken down primarily by platform, then by install method: there is some duplication (i.e. conda-forge) but this is needed because people will search for instructions mainly by the platform they want to install on *or* by installation method
- moved "install by source" to locate it with the other installation instructions
- broke out "get help, report bugs etc to it's own section and created sub-headings: help to break up the content to make readability easier
- added link to gqrx matix channel
- broke credits and license out to its own section and added link to GPL
- broke iconography out to its own section and added links to icons: helpful for those who might want to contribute new icons/design

If anyone things the README is getting long/too long, I can break the installation instructions out into it's own INSTALL.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,79 +4,73 @@ Gqrx
 [![CI](https://github.com/gqrx-sdr/gqrx/workflows/CI/badge.svg)](https://github.com/gqrx-sdr/gqrx/actions?query=workflow%3ACI+branch%3Amaster)
 [![Build](https://github.com/gqrx-sdr/gqrx/workflows/Build/badge.svg)](https://github.com/gqrx-sdr/gqrx/actions?query=workflow%3ABuild+branch%3Amaster)
 
-Gqrx is an open source software defined radio (SDR) receiver implemented using
-[GNU Radio](https://gnuradio.org) and the [Qt GUI toolkit](https://www.qt.io/).
-Currently it works on Linux and Mac with hardware supported by gr-osmosdr,
-including Funcube Dongle, RTL-SDR, Airspy, HackRF, BladeRF, RFSpace, USRP and
-SoapySDR.
+Gqrx is an open source software defined radio (SDR) receiver implemented using [GNU Radio](https://gnuradio.org) and the [Qt GUI toolkit](https://www.qt.io/).
+Currently it works on Linux and Mac with hardware supported by gr-osmosdr, including Funcube Dongle, RTL-SDR, Airspy, HackRF, BladeRF, RFSpace, USRP and SoapySDR.
 
-Gqrx can operate as an AM/FM/SSB receiver with audio output or as an FFT-only
-instrument. There are also various hooks for interacting with external
+Gqrx can operate as an AM/FM/SSB receiver with audio output or as an FFT-only instrument. There are also various hooks for interacting with external
 applications using network sockets.
 
 ![Screenshot of the main Gqrx window](resources/screenshots/gqrx-main.png)
 
 
 Download
---------
+========
 
-Gqrx is distributed as a source code package and binaries for Linux and Mac.
-Many Linux distributions provide gqrx in their package repositories.
-Alternate Mac support is available through [MacPorts](https://ports.macports.org/port/gqrx/summary) and [Homebrew](https://formulae.brew.sh/cask/gqrx).
+Gqrx is distributed as a source code package and binaries for Linux and Mac. Many Linux distributions provide gqrx in their package repositories.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,79 +4,73 @@ Gqrx
...
+The simplest installation for Raspberry-Pi devices is via the Raspberry-Pi software repository.
 
-Getting help and reporting bugs
--------------------------------
+```
+open the Applications menu
+select Preferences
+Add / Remove Software
+search for Gqrx
+After installation, Gqrx will be available in the Internet section of the Applications menu
```
LongnoseRob, line 55: I would like to recommend to rephase this point, on ArchLinux(Arm) with [hamrado-menu](https://aur.archlinux.org/packages/hamradio-menus) installed, gqrx shows up in the hamradio section..
I dont know how it is on other distributions.

README.md
```diff
@@ -4,79 +4,73 @@ Gqrx
...
-If you don't see your device listed in the drop-down list it could be because:
-- The driver has not been included in a binary distribution
-- The udev rule has not been properly configured
-- Linux kernel driver is blocking access to the device
+Linux distributions
+-------------------
 
-You can test your device using device specific tools, such as rtl_test,
-airspy_rx, hackrf_transfer, qthid, etc.
+The simplest installation for Linux distributions is by our [official AppImage release](https://github.com/gqrx-sdr/gqrx/releases). Many Linux distributions provide gqrx in their package repositories. Linux distribution versions can be out-of-date.
```
LongnoseRob, line 34: Here it might be worth to hint to give feedback to maintainers to push for newer releases..

README.md
```diff
@@ -148,6 +142,70 @@ Use Qt Creator as before
...
+
+The first time you start gqrx it will open a device configuration dialog.
+Supported devices that are connected to the computer are discovered
+automatically and you can select any of them in the drop-down list.
+
+If you don't see your device listed in the drop-down list it could be because:
+- The driver has not been included in a binary distribution
+- The udev rule has not been properly configured
+- Linux kernel driver is blocking access to the device
+
```
LongnoseRob, line 163: One cause could also be that the hardware is not connected..

README.md
```diff
@@ -148,6 +142,70 @@ Use Qt Creator as before
...
+Tutorials and howtos are being written and published on the website
+https://gqrx.dk/
+
+
+Known problems
+--------------
+
+See the bug tracker on Github: https://github.com/gqrx-sdr/gqrx/issues
+
+
```
LongnoseRob, line 184: It might be worth to give some details on which data should be provided for bug-reports

LongnoseRob (commented):
Some small comments which could improve this already good readme

README.md
ei8fdb: Yup. I think best approach to this would be to create a bug report issue template? Thoughts? Before I go create one?

README.md
ei8fdb: Ok so either specify these instructions are for Ras-Pi OS, or make them generic, ala "then search for gqrx to start it."

I think the recommended way would be to specify each OS, but this will make the readme longer.

In which case I think it'd be better to break these instructions out into an INSTALL.md?

README.md
ei8fdb: Yup. I think best approach to this would be to create a bug report issue template? Thoughts? Before I go create one?

README.md
ei8fdb: Ok, I'll add that to the list too.

README.md
LongnoseRob: Maybe we could create a small script which would collect most of the required info?
This cript could run uname-a, etc

vladisslav2011: I think, good bug report should look like this:
Title: `BUG: Gqrx freezes when hackrf (selected as an input device) is disconnected from USB port.`
Body:
```
GUI freeze when HackRF is selected as an input device and USB connection breaks.

Conditions:
gqrx built from git master
Platform: Linux
Distribution: Ubuntu ...
GNU Radio v3.9...
Qt: 5.9.5...
Libhackrf: ...

Steps to reproduce:
1. Connect HackRF to a USB port
2. Start gqrx
3. Select HackRF as an input device and start the DSP
4. Disconnect HackRF from the USB port

Expected behavior:
Gqrx should stop producing audio samples and should continue to show last FFT result. Gqrx should allow to select another input device using the "Configure IO devices" dialog.

Observed behavior:
Gqrx GUI freezes. No reaction to "Configure IO devices" button.

Screenshots, videos, additional notes, stack traces...
```

PR #1130, opened 2022-06-30
Save FFT Zoom state

va7acq:
Adds the ability to save and load the FFT Zoom state in the configuration file.


argilo:
Thanks! This appears to work well.

I changed the default value to 1, to match the lower limit of the zoom slider. Once CI passes, I'll get this merged.

Issue #1140, opened 2022-08-11
BUG: input decimation setting breaks IQ Tool playback


vladisslav2011:
Steps to reproduce:
1. Get an IQ recording (Start DSP, open IQ tool, press Record, wait, press record again, close IQ tool)
2. Open "Configure I/O devices" dialog
3. Set "Decimation" to a value, larger than 1.
4. Press "OK"
5. Open IQ tool
6. Choose an IQ recording and press "Play"

Expected behavior:
Normal waterfall/pandapter and frequency display, normal sound, clicking a frequency on a waterfall/panadapter tunes to it. No decimation applied.

Observed behavior:
Waterfall/panadapter do not match actual chosen frequency. Decimation is partially applied. Some receiver blocks get wrong sampling rate. Distorted sound.

PR #1141, opened 2022-08-11
Fix IQ Tool playback when input decimation is enabled

vladisslav2011:
Disable the input decimation decimation before starting IQ playback.
This should make waterfall/panadapter frequency display and fix distorted sound.
Close https://github.com/gqrx-sdr/gqrx/issues/1140

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1603,6 +1603,7 @@ void MainWindow::startIqPlayback(const QString& filename, float samprate, qint64
 
     qDebug() << __func__ << ":" << devstr;
 
+    rx->set_input_decim(1);
     rx->set_input_device(devstr.toStdString());
     updateHWFrequencyRange(false);
 


willcode:
It's sort of ambiguous whether the user would mean to have decimation apply to playback. I agree that in general, the input device setting should apply to the input device only and not to playback, so I think this would be a good change. I assume the config reset after playback puts the decim back to what it was before?
If decim is desired for playback, it should be specified in the IQ tool.
Stuff like this is part of the reason I'd like to rearchitect the input portion to make files, zmq, etc. more equal to hardware devices.

vladisslav2011:
Yes. Config reset resets all settings, including decimation to its' initial state.
Adding some sort of input device abstraction is a good idea. gr-osmosdr was such an abstraction for a long time, but it looks, like nobody wants to maintain it...
There is one feature, that should be kept in mind while adding an input abstraction: timeshift. Gqrx should be able to record an IQ file and switch between direct processing and playing the file from selected point while recording (in the future).
I can try to implement this, but I want to have some bugs fixed first (like this one) and to finish my multi_vfo implementation: put the UDP streamer into receiver channel's hier_block2 and implement multichannel data receiver (AFSK1200) - put data processing blocks into receiver channel's hier_block2 and update the GUI.

willcode:
Doesn't look like things get restored after playback stops with this one-liner. I had to implement this for other stuff I'm working on ... will submit when I can.

vladisslav2011:
I've retested and can confirm, that this does not work.
Let's restore my previous fix: https://github.com/gqrx-sdr/gqrx/pull/1064

PR #1142, opened 2022-08-15
Iq recording can now be controlled via remote control interface.

stefanino-ch:
Remote control interface extension to control also iq recording. 
One method of iq recording renamed to get a symmetrical solution with iaudio recording without name clashes.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.17.7: In progress...
+
+       NEW: Start/stop I/Q recording via remote control.
+
+
     2.17.6: Released November 29, 2024
 
        NEW: Fetch RDS Program Service & RadioText via remote control.
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -37,6 +37,10 @@ Supported commands:
     Get status of audio recorder
  U RECORD <status>
     Set status of audio recorder to <status>
+ u IQRECORD
+    Get status of IQ recorder
+ U IQRECORD <status>
+    Set status of IQ recorder to <status>
  u DSP
     Get DSP (SDR receiver) status
  U DSP <status>
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -319,7 +319,9 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
 
     // I/Q playback
     connect(iq_tool, SIGNAL(startRecording(QString, QString)), this, SLOT(startIqRecording(QString, QString)));
+    connect(iq_tool, SIGNAL(startRecording(QString)), remote, SLOT(startIqRecorder(QString)));
     connect(iq_tool, SIGNAL(stopRecording()), this, SLOT(stopIqRecording()));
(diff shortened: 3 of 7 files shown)


willcode:
You might want to submit the audio recorder fix separately.
Is the shortening of a couple function names for style, or was there a conflict?
General idea looks good. Haven't tried it out.

stefanino-ch:
Audio fix: just in https://github.com/gqrx-sdr/gqrx/pull/1142/commits/ac4dd91287388d534bd3bdc2977221d8f41a070d#diff-2e56865e037f0cdab2b444f8dcc08143f82b9cde9b9734fed5d05bf7eedad8d3 lines 737 until 746. Next time I will try to seperate the changes in different pulls. 

Function names adapted: audio and  iq recording are very similar in their implementation, but the naming of methods doing the same  (once for audio and once for iq recording) used different schemes. So in principle it's more style (and future maintability) than real functional conflict.

stefanino-ch:
Changes for audio recording has been moved into another Pull Request.

willcode:
My opinion: if you could change the function names back for now, that would be helpful. There are plenty of things that could use renaming in the future.

`iq_recorder_status` is not wired up such that it knows if a recording has been started via the IQ dialog. Any thoughts on whether/how the wiring should be completed? I know the Gqrx signal/slot usage gets complicated with all the routing and the fact that it's all manually specified.

stefanino-ch:
> is not wired up
Yep only the syncronisation from remote control to the iq tool was implemented. The other way around there was two connects missing, the methods was already there.

stefanino-ch:
From my point of view the development is done. 

Question: is there a way to flag this somehow? After all research I have not found an answer.

Question: do I need to add the checks in the "Checks" tab by myself? How?

argilo:
Sorry for the very long delay in looking at this.

It seems there is plenty of interest in this feature since #1403 and #1405 were recently opened.

I rebased this branch and squashed the commits, and also undid the renaming of `startIqRecording` / `stopIqRecording`. I'll take the changes for a spin soon and have a look over the rest of the code.

argilo:
It looks like a few additional code changes may be needed to account for the fact that there are now two I/Q recording formats (raw & SigMF); that change was made in #1182. Perhaps it should be possible to specify the recording format in the remote control command.

stefanino-ch:
> It looks like a few additional code changes may be needed to account for the fact that there are now two I/Q recording formats (raw & SigMF); that change was made in #1182. Perhaps it should be possible to specify the recording format in the remote control command.

I used the remote control in a way, that gqrx was always running. Via my [scheduler](https://github.com/stefanino-ch/gqrx-apscheduler) running as seperate programm I just started/ stopped the recording. If you operate in this way, there's no need to set the recording format. I personally think, spefifying the recording format via remote control would be nice but not needed.

mhassell:
Hi, I tested this as well and it looks okay to me.  I was able to build and run the command

> U IQRECORD 1

and confirmed I captured IQ data, which I then played back to confirm it worked.

I also confirmed that 

> U IQRECORD 0 

stops the IQ recorder.

My only suggestion would be to open the IQ recorder window so it is clear to the user that something is being recorded.  Otherwise, if the telnetlib disconnects we will wind up recording IQ data quietly.

argilo:
A few more changes were needed to make the code compatible with the current I/Q recorder. All looks to be working well now, so I'll merge this.

Issue #1143, opened 2022-08-16
Segmentation fault when working with bookmarks


yuvadm:
When working with bookmarks at some point this crash happens.

Versions: Arch Linux 5.18.16 + Qt 6.3.1

```
Thread 1 "gqrx" received signal SIGSEGV, Segmentation fault.
0x00007ffff67de8da in QString::operator=(QString const&) () from /usr/lib/libQt6Core.so.6
(gdb) bt
#0  0x00007ffff67de8da in QString::operator=(QString const&) () at /usr/lib/libQt6Core.so.6
#1  0x00007ffff67dec16 in QString::append(QString const&) () at /usr/lib/libQt6Core.so.6
#2  0x000055555565a4ee in BookmarksTableModel::data(QModelIndex const&, int) const ()
#3  0x00007ffff68eb421 in QAbstractItemModel::multiData(QModelIndex const&, QModelRoleDataSpan) const () at /usr/lib/libQt6Core.so.6
#4  0x00007ffff7832ff3 in QStyledItemDelegate::initStyleOption(QStyleOptionViewItem*, QModelIndex const&) const () at /usr/lib/libQt6Widgets.so.6
#5  0x00007ffff7836c30 in QStyledItemDelegate::paint(QPainter*, QStyleOptionViewItem const&, QModelIndex const&) const () at /usr/lib/libQt6Widgets.so.6
#6  0x00007ffff7859cb1 in  () at /usr/lib/libQt6Widgets.so.6
#7  0x00007ffff785f8af in QTableView::paintEvent(QPaintEvent*) () at /usr/lib/libQt6Widgets.so.6
#8  0x00007ffff75b82c0 in QWidget::event(QEvent*) () at /usr/lib/libQt6Widgets.so.6
#9  0x00007ffff7643b37 in QFrame::event(QEvent*) () at /usr/lib/libQt6Widgets.so.6
#10 0x00007ffff6711c88 in QCoreApplicationPrivate::sendThroughObjectEventFilters(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
#11 0x00007ffff75749cc in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt6Widgets.so.6
#12 0x00007ffff6713088 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt6Core.so.6
[...]

PR #1145, opened 2022-08-23
Switch to using shared pointers for TagInfo

vladisslav2011:
Avoid crash on Qt6 due to use after free.
Fixes https://github.com/gqrx-sdr/gqrx/issues/1143
Needs testing.


yuvadm:
Pulled and tested this fix, looks good AFAICT. There are some preexisting bugs with the tag lists, but no more seg faults.

pinkavaj:
I'm afraid this only hides the symptoms but does not solve the problem. The question is, how it is even possible the objects are destroyed, when eg. new tag is added. My guess is the QList stores the objects and relocates them once new object is added. So we cannot combine QList and reference its objects by pointers.

vladisslav2011:
> So we cannot combine QList and reference its objects by pointers.

That's exactly the issue. Qt6 QList copies it's members during internal housekeeping and invalidates direct pointers. Accessing invaludated pointers results in bookmarks corruption, segfaults, etc... And using shared pointer solves it. We can copy/create/move shared pointers to an object, but the object remains the same and other shared pointer does not get invalidated. So QList stops breaking things and everything starts working as expected.

pinkavaj:
You are completely right, I have missed the important line, my claim above is invalid. Sorry for confusion.

Just thinking if it really makes sense to use relatively expansive `shared_ptr`. The objects should not be referenced after deletion from the list, if it is used it is likely a bug. Using `shared_ptr` will ensure the program will work, except it will use old data. I can imagine this might lead to wierd behaviouir (like using and displaying value from already deleted object).
With `unique_ptr`, the program will probably crash or it will be possible to identify such a think using `valgrind`

pinkavaj:
@argilo Are we expected to also add line to news?

src/qtgui/bookmarks.cpp
```diff
@@ -315,10 +313,10 @@ const QColor BookmarkInfo::GetColor() const
 {
     for(int iTag=0; iTag<tags.size(); ++iTag)
     {
-        TagInfo& tag = *tags[iTag];
-        if(tag.active)
+        TagInfo::sptr tag = tags[iTag];
```
pinkavaj, line 316: This is completely valid and straightforward but a bit inefficient due to the copy-construction of the shared pointer. This assignment here is for convenience only, to avoid more complicated expressions bellow. Using a reference style (as before the change) is IMHO most convenient. Please consider dereferencing the pointer to reference.

src/qtgui/bookmarks.h
```diff
@@ -30,26 +30,27 @@
...
+    using sptr = std::shared_ptr<TagInfo>;
     QString name;
     QColor color;
     bool active;
 
     static const QColor DefaultColor;
     static const QString strUntagged;
 
-    TagInfo()
+    TagInfo(QString name = "")
```
pinkavaj, line 45: May-be simplify this by by #1176 first (if it makes sense to you, can you help me to get the MR in?)

But definitely a blocker, I can easily rebase. (I would suggest not to merge the change from #1176 to avoid mixing thinks together).

src/qtgui/bookmarks.cpp
```diff
@@ -158,21 +158,20 @@ bool Bookmarks::save()
         stream << QString("# Tag name").leftJustified(20) + "; " +
                   QString(" color") << '\n';
 
-        QSet<TagInfo*> usedTags;
+        QMap<QString, TagInfo::sptr> usedTags;
```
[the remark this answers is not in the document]
vladisslav2011: Using raw pointers here makes Gqrx larger and makes code more complicated. You may try it yourself...

src/qtgui/bookmarks.cpp
vladisslav2011: Thanks. Switching to using a const reference simplifies code a bit. Tests pending...

src/qtgui/bookmarks.h
vladisslav2011: #1176 looks fine to me. Approved. I'll rebase this PR if #1176 gets merged first.

src/qtgui/bookmarks.cpp
pinkavaj: By dereferencing I ment somethink like `TagInfo &tag = *tags[iTag];` :)

src/qtgui/bookmarks.cpp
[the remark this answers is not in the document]
pinkavaj: To make sure we understand each other, the proposal was to keep the old code (for now) except the necessary changes to make it work with shared pointer, eg. the only change necessary would be somethink like:

```
             BookmarkInfo& info = m_BookmarkList[iBookmark];
             for(int iTag = 0; iTag < info.tags.size(); ++iTag)
             {
-              TagInfo& tag = *info.tags[iTag];
-              usedTags.insert(&tag);
+              usedTags.insert(info.tags[iTag].get());
             }
         }
```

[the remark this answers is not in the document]
vladisslav2011: I can't guarantee, that QSet<TagInfo*> will work correctly here (collect tags with unique names). It may be better to change `Bookmarks::m_TagList` type form `QList<TagInfo>` to `QMap<QString,TagInfo::sptr>` and get rid of this loop.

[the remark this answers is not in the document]
pinkavaj: Yep, the non-uniqueness of tags should be solved too, but it seems to me this tries to solve more thinks in one MR. Mey-be separate commit if it is necessary to fix #1143 or separate MR if it can be done independently would help me.

src/qtgui/bookmarks.cpp
vladisslav2011: No. My latest commit is wrong. I'll remove it.
We should never create a reference to a smart pointer/a reference to a managed object as it defeats the purpose of using smart pointers to prevent use-after-free.

src/qtgui/bookmarks.cpp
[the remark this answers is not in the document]
vladisslav2011: Switching to a better container may be a different PR.
I do not know, how other people are using tags. Tag management should be improved if someone is using thousands of tags (adding a quick filter to help finding tags/bookmarks may help in this case).

src/qtgui/bookmarks.cpp
pinkavaj: I does not see any way how use-after-free can happend within this function, is it?

vladisslav2011: 1. I'm thinking about offloading plotter rendering to a background thread.
2. https://schneide.blog/2012/01/30/dont-mix-c-smart-pointers-with-references/

pinkavaj: 1. Unfortunately I cannot see how this might affect this, because there is no code yet, so I'm afraid it's hard to discuss.
2. Fine article, but it applies to situation where the smart pointer value is passed to another function/location. I belive bad usages like this can be caught during code-review, so using it strictly in local context is IMHO fine.

Nevertheless the situation still can be avoided just by replacing the `tag` by the whole `tags[iTag]`, so this option can be considered too.

argilo:
I rebased and resolved merge conflicts so I can test this out.

argilo:
This seems like a sensible approach, and I verified that it resolves the invalid memory accesses.

PR #1148, opened 2022-09-11
Add FFT control to remote control commands

dl3jop:
For applications were gqrx is controlled remotely, I missed the option to control some options of the FFT window.
Example: Using gqrx without mouse & keyboard but external hardware which changes frequencies and keeps the grey demod box centered in the middle of the FFT screen.

Therefore I added some new remote commands. I tried to stay within the name convention:

- SDC: Centers the demodulator area in the middle of the FFT window
- RFZ: Resets the FFT zoom
- SFZ: Sets the FFT zoom
- SFC: Centers the whole FFT window


willcode:
The code looks reasonable.

I haven't set up a system where Gqrx is driven completely by another program. What kind of setup do you have that makes this useful?

dl3jop:
**In my case** I want to use it for amateur radio satellite transponders where a standard VHF/UHF transceiver is used for uplink an a SDR for downlink. As this setup should be portable I needed something which automatically zooms/pans around the spectrum when changing frequency. Using this I can zoom as far into the waterfall just to see the transponder passband and nothing more.
In my case an external software handles the radio and gqrx as well as physical buttons and keeps the demodulation spectrum centered as well es automatically zooms to the chosen transponders pass-band width

**In general** it can be used in applications where GQRX is used as a automatic pandandapter or monitoring system where no user input is expected, wanted or feasible.

PR #1151, opened 2022-10-08
Improve AGC efficiency

### LINKED ISSUE #1055: rx_agc_cc sometimes eats the whole CPU core

argilo:
Fixes #1055.

I had a look into why Gqrx's AGC is so inefficient, and it's because the code that searches for the magnitude peak within a sliding window searches the entire window for a new peak every time the previous peak leaves the window. The loop on lines 238-243 causes very high CPU utilization:

https://github.com/gqrx-sdr/gqrx/blob/26c12e84f21b5bb7305ed8aff51c2ae975447de6/src/dsp/agc_impl.cpp#L223-L245

When the squelch is closed, this situation occurs on every sample, thus triggering the AGC's worst-case behaviour.

Peak detection can be performed much more efficiently using a deque, as described here: https://www.geeksforgeeks.org/sliding-window-maximum-maximum-of-all-subarrays-of-size-k/

After this change, the performance of the peak detector no longer depends on the input samples.


vladisslav2011:
Hello.
This implementation solves the problem with AGC hogging the CPU when the squelch is closed.
But the CPU usage is still somewhat higher (but acceptable), than when alternative implementation is used.

argilo_agc
alt_agc

Maybe that's because this algorithm is not constant complexity and still sometimes creates spikes of CPU usage (when a lot of indexes are discarded from tracking deque on a new peak)...

argilo:
Thanks for having a look.

My goal with this pull request was to keep the AGC output identical to the current implementation while reducing CPU utilization, thus making it suitable for a bug-fix release. I'm still open to swapping out the AGC for a better one in the next feature release.

vladisslav2011 (approved):

vladisslav2011:
I think, quiet sound can't be fixed by the AGC before the demodulator.
I'll keep my PRs (and resolve merge conflicts when this gets merged).
CPU load comparison may be unfair as it is possible, that I have `set_output_multiple` on some block before the AGC and this may be a game-changer. I'll try to find and eliminate it and retest.

argilo:
> Maybe that's because this algorithm is not constant complexity and still sometimes creates spikes of CPU usage (when a lot of indexes are discarded from tracking deque on a new peak)...

That's correct. The total number of deque operations is limited to 2 times the number of input samples, but some samples will use more operations than others.

vladisslav2011:
I've found and removed `set_output_multiple`s, that were in action (in resampler_xx.cpp)
```
vlad@lucid:~/warez/gqrx-my/build$ grep -r set_output_multiple ../src
../src/dsp/rds/decoder_impl.cc:	set_output_multiple(104);  // 1 RDS datagroup = 104 bits
../src/dsp/rx_rnnoise.cpp:    set_output_multiple(d_frame_size);
../src/dsp/resampler_xx.cpp:    d_filter->set_output_multiple(4096);
../src/dsp/resampler_xx.cpp:    d_filter->set_output_multiple(4096);
../src/dsp/rx_fft.cpp:    set_output_multiple(8192);
vlad@lucid:~/warez/gqrx-my/build$ vim ../src/dsp/resampler_xx.cpp
vlad@lucid:~/warez/gqrx-my/build$ grep -r set_output_multiple ../src
../src/dsp/rds/decoder_impl.cc:	set_output_multiple(104);  // 1 RDS datagroup = 104 bits
../src/dsp/rx_rnnoise.cpp:    set_output_multiple(d_frame_size);
../src/dsp/rx_fft.cpp:    set_output_multiple(8192);
```
The other two in rx_rnnoise.cpp (disabled) and rx_fft.cpp (fft_channelizer is disabled too) should not change anything.
All CPU cores were set to minimum possible frequency and turbo was disabled during this and previous tests.

alt_rx_agc_retest
alt_rx_agc2

rx_agc_2f CPU load fluctuates between 0.3 and 0.7 % even when the AGC is set to maximum 1000ms attack time (maximum peak tracking buffer size), so the constant complexity implementation is still a bit faster.

vladisslav2011:
No. Maybe I'm wrong again. rnnoise noise reduction block is always connected even when bypassed. So it should be removed completely.

vladisslav2011:
With all `set_output_multiple`s eliminated alternative implementation is a bit slower:
agc_no_rnnoise

PR #1156, opened 2022-10-30
Bandplan stacking and dynamic bookmark levels

shawnj007:
When plotter is initialized with a small area, the number of bookmark label levels cannot change with resize. Removing the "static const" fixes this issue.

The bandplan now stacks ranges that overlap frequencies. This allows for more complicated bandplans and for all overlapping ranges to be identified.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1298,7 +1298,7 @@ void CPlotter::drawOverlay()
         static const int fontHeight = fm.ascent() + 1;
         static const int slant = 5;
         static const int levelHeight = fontHeight + 5;
-        static const int nLevels = h / (levelHeight + slant);
+        int nLevels = h / (levelHeight + slant);
         if (m_BookmarksEnabled)
         {
             tags = Bookmarks::Get().getBookmarksInRange(m_CenterFreq + m_FftCenter - m_Span / 2,
@@ -1375,20 +1375,45 @@ void CPlotter::drawOverlay()
     {
         QList<BandInfo> bands = BandPlan::Get().getBandsInRange(m_CenterFreq + m_FftCenter - m_Span / 2,
                                                                 m_CenterFreq + m_FftCenter + m_Span / 2);
-
+        int bandLevel = 0;
+        qint64 last_max = 0;
+        qint64 last_max_0 = 0;
+        
+        std::sort(bands.begin(), bands.end(), [=]( const BandInfo& band1 , const BandInfo& band2 )->bool {
+            if (band1.minFrequency == band2.minFrequency) {
+                if (band1.maxFrequency != band2.maxFrequency) 
+                    return band1.maxFrequency > band2.maxFrequency;
+                else
+                    return band1.name < band2.name;
+            }
+            return band1.minFrequency < band2.minFrequency;
+        });
+        
         for (auto & band : bands)
         {
+            if (band.minFrequency >= last_max_0) {
+                bandLevel = 0;
+                last_max = band.maxFrequency;
+                last_max_0 = last_max;
+            } else if (band.minFrequency < last_max) {
+                --bandLevel;
+                last_max = band.maxFrequency;
+            } else if (band.minFrequency == last_max) {
+                last_max = band.maxFrequency;
+            }
+            
             int band_left = xFromFreq(band.minFrequency);
             int band_right = xFromFreq(band.maxFrequency);
             int band_width = band_right - band_left;
-            rect.setRect(band_left, xAxisTop - m_BandPlanHeight, band_width, m_BandPlanHeight);
+            int band_bottom = bandLevel * m_BandPlanHeight;
+            rect.setRect(band_left, xAxisTop - m_BandPlanHeight + band_bottom, band_width, m_BandPlanHeight);
             painter.fillRect(rect, band.color);
             QString band_label = band.name + " (" + band.modulation + ")";
             int textWidth = metrics.boundingRect(band_label).width();
             if (band_left < w && band_width > textWidth + 20)
             {
                 painter.setOpacity(1.0);
-                rect.setRect(band_left, xAxisTop - m_BandPlanHeight, band_width, metrics.height());
+                rect.setRect(band_left, xAxisTop - m_BandPlanHeight + band_bottom, band_width, m_BandPlanHeight);
                 painter.setPen(QColor(PLOTTER_TEXT_COLOR));
                 painter.drawText(rect, Qt::AlignCenter, band_label);
             }


shawnj007:
Here is a band plan I adapted from: [Original BandPlan.xml](https://github.com/Arrin-KN1E/SDR-Band-Plans/blob/master/US/SDR%23/BandPlan.xml)
A version that works in gqrx can be found: [Modified bandplan.csv](https://github.com/shawnj007/gqrx-touch/blob/master/resources/bandplan.csv)

src/qtgui/plotter.cpp
```diff
@@ -1375,20 +1375,45 @@ void CPlotter::drawOverlay()
...
+            if (band.minFrequency >= last_max_0) {
+                bandLevel = 0;
+                last_max = band.maxFrequency;
+                last_max_0 = last_max;
+            } else if (band.minFrequency < last_max) {
+                --bandLevel;
+                last_max = band.maxFrequency;
+            } else if (band.minFrequency == last_max) {
+                last_max = band.maxFrequency;
+            }
```
pinkavaj, line 1403: The `bandLevel` can reach arbitrary value, but the algorithm currently works only up to level 2 (see attached bandplan bellow to reproduce the issue). Although this will be rare case I would suggest either to limit the `bandLevel` up to some level or make this implementation generic.

There is also the edgecase where the bandplans can get out of the screen (the stack is too high). But this is really not a practical concern.

```
100000000,       102000000,       CW,  1,   #40BAFD00, Power A
100000000,       101000000,       CW,  1,   #40B00D00, Power D
101000000,       102000000,       CW,  1,   #40B00D00, Power E
100500000,       101500000,       CW,  1,   #40B00D00, Power G
```

pinkavaj (changes requested):

src/qtgui/plotter.cpp
```diff
@@ -1375,20 +1375,45 @@ void CPlotter::drawOverlay()
...
+        
+        std::sort(bands.begin(), bands.end(), [=]( const BandInfo& band1 , const BandInfo& band2 )->bool {
+            if (band1.minFrequency == band2.minFrequency) {
+                if (band1.maxFrequency != band2.maxFrequency) 
+                    return band1.maxFrequency > band2.maxFrequency;
+                else
+                    return band1.name < band2.name;
+            }
+            return band1.minFrequency < band2.minFrequency;
+        });
```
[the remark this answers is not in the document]
shawnj007: Agreed, [=] is not necessary.

As far as the sort, I don't think it increases the resources needed by very much, as it is only sorting a subset of the band plan each time the function is called. I think it's a pretty quick code path for the sort as it's a maximum of 3 comparisons for each band and usually only 2. Putting it in the load may save some resources if the overlay is drawn repeatedly. Doing so would require modification of the "m_bandPlanFile" variable, which is feasible, but I question if it is necessary.

src/qtgui/plotter.cpp
```diff
@@ -1298,7 +1298,7 @@ void CPlotter::drawOverlay()
         static const int fontHeight = fm.ascent() + 1;
         static const int slant = 5;
         static const int levelHeight = fontHeight + 5;
-        static const int nLevels = h / (levelHeight + slant);
+        int nLevels = h / (levelHeight + slant);
```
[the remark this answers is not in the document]
shawnj007: Correct, this is a small, cosmetic fix that will make "excess" bookmark labels disappear not simply be written beyond the bottom of the plotter.

I was not able to reproduce the original problem for which I made this change. It can be disregarded, but I think it should remain part of the codebase so that when the plotter window does have a small height the labels don't overwrite other things (such as the band plan, although this awareness remains to be implemented).

src/qtgui/plotter.cpp
shawnj007: You are correct, it will "stair-step" for certain band plans. It still is a better implementation than no stacking at all, which simply writes one band over the top of others. It works as intended when a new band starts after the lowest level max has been exceeded.

To make it generic would require a "last_max" record for each level. Again, feasible, but is it worth the time to implement and the resources to check in order to avoid the "stair-step"?

src/qtgui/plotter.cpp
```diff
@@ -1375,20 +1375,45 @@ void CPlotter::drawOverlay()
     {
         QList<BandInfo> bands = BandPlan::Get().getBandsInRange(m_CenterFreq + m_FftCenter - m_Span / 2,
                                                                 m_CenterFreq + m_FftCenter + m_Span / 2);
-
+        int bandLevel = 0;
+        qint64 last_max = 0;
+        qint64 last_max_0 = 0;
+        
+        std::sort(bands.begin(), bands.end(), []( const BandInfo& band1 , const BandInfo& band2 )->bool {
```
pinkavaj, line 1382: There is already `bool operator<(const BandInfo &other) const` operator, consider extending it and use here instead of the lambda.

src/qtgui/plotter.cpp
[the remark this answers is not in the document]
pinkavaj: It looks like pre-sorting in `load()` this will be overcomplicatin now. I'm abandoning the idea for now.

src/qtgui/plotter.cpp
```diff
@@ -1375,20 +1375,45 @@ void CPlotter::drawOverlay()
...
+            int band_bottom = bandLevel * m_BandPlanHeight;
+            rect.setRect(band_left, xAxisTop - m_BandPlanHeight + band_bottom, band_width, m_BandPlanHeight);
             painter.fillRect(rect, band.color);
             QString band_label = band.name + " (" + band.modulation + ")";
             int textWidth = metrics.boundingRect(band_label).width();
             if (band_left < w && band_width > textWidth + 20)
             {
                 painter.setOpacity(1.0);
-                rect.setRect(band_left, xAxisTop - m_BandPlanHeight, band_width, metrics.height());
+                rect.setRect(band_left, xAxisTop - m_BandPlanHeight + band_bottom, band_width, m_BandPlanHeight);
```
pinkavaj, line 1416: The change of  `metrics.height()` to `m_BandPlanHeight` will change behavior in case the font size does not fit into the band rect. Before it was getting out of the bound, now it will be cropped.

The line is now redundant with the same line above. Either should be removed or reverted to the original ` `metrics.height()`.

src/qtgui/plotter.cpp
```diff
@@ -751,12 +751,12 @@ void CPlotter::zoomStepX(float step, int x)
     qint64 new_FftCenter;
     if( limit_hit ) // cannot keep fixed_hz fixed
     {
-    	new_FftCenter = qRound64((f_min + f_max) / 2.0) - m_CenterFreq;
+        new_FftCenter = qRound64((f_min + f_max) / 2.0) - m_CenterFreq;
```
shawnj007, line 754: I don't know why the tabs/spaces changed, but it's consistent now. Same as below.

src/qtgui/plotter.cpp
shawnj007: Done.

src/qtgui/plotter.cpp
pinkavaj: I see this as a great improvement. What I'm trying to achieve is to help with getting it into the codebase, in case you are willing to give it a try, I will do my best to provide support to make the implementation generic, which I believe might help.

So my opinion is it is worth the time and I'm willing to invest here.

shawnj007:
Sorry for the force pushes, I'm obviously having issues with local branch management.

src/qtgui/plotter.cpp
shawnj007: Cool, any hints on making it generic would be nice, but maybe the better option (at least for now) is to limit it to 2 levels? My entire bandplan (linked above) never uses more than 2 levels. What do you think is more appropriate?

pinkavaj: Limiting it to level 2 might also help to get into situation where the tags will go over the top of the bar (which is also unlikely to happend edgecase). Honestly both solutions (and even the staircase) are acceptable for my in the end,   so @argilo my question will be what have the better chance to get into the codebase?

The worst what can happend from my side is I will create another pull request to make the "next step ..". :)

pinkavaj:
> Sorry for the force pushes, I'm obviously having issues with local branch management.

FYI: I use force-push for pull requests regularly if not told not to do. So correct behaviour is project dependent :)

PR #1157, opened 2022-10-31
Make it possible to control audio gain via the remote interface

vladisslav2011:
New gain control "AF"
Use "l AF" command to get current audio gain.
Use "L AF 10" to set current audio gain to 10.0 db.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -2,6 +2,7 @@
       2.16: In progress...
 
        NEW: Restore AM & AM-Sync settings between sessions.
+       NEW: Set/get audio gain via remote control.
      FIXED: Loading of narrow FM tau setting.
    REMOVED: Support for GNU Radio 3.7.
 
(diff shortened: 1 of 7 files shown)


kukabu:
Hello

I think it must be 'AF' instead of 'AUDIO' according https://github.com/Hamlib/Hamlib/blob/master/include/hamlib/rig.h#L957 for better hamlib compatible and this command must be added to documentation https://github.com/gqrx-sdr/gqrx/blob/master/resources/remote-control.txt

vladisslav2011:
@kukabu Thanks. The branch is updated.

argilo:
I noticed a bug: If Gqrx is started with the audio gain at the default value of -6.0, then `l AF` incorrectly returns 0.0 because `audioGainChanged` never occurs. I fixed that by initializing the remote control's `audio_gain` member variable to -6.0 to match the default value.

PR #1161, opened 2022-11-06
Simplify setFrequencyFocus()

pinkavaj:
- Replace deprecated QMouseEvent constructor (Deprecated since Qt 6.4)
- Focus only if not already focuses
- Keep the current position if focused

### DIFF
--- a/src/qtgui/freqctrl.cpp
+++ b/src/qtgui/freqctrl.cpp
@@ -876,13 +876,12 @@ void CFreqCtrl::cursorEnd()
 
 void CFreqCtrl::setFrequencyFocus()
 {
-    uint8_t position = floor(log10(m_freq));
-    position = (uint8_t)fmax(position, 4);      // restrict min to 100s of kHz
+    // Select last digit or 5th digit (100s of kHz), whatever is bigger.
+    int position = std::max(int(log10(m_freq)), 5);
 
-    QMouseEvent mouseEvent(QEvent::MouseMove,
-                           m_DigitInfo[position].dQRect.center(),
-                           Qt::NoButton,
-                           Qt::NoButton,
-                           Qt::NoModifier);
-    mouseMoveEvent(&mouseEvent);
+    if (!hasFocus()) {
+        setFocus(Qt::ShortcutFocusReason);
+    }
+
+    setActiveDigit(position);
 }


argilo:
CI fails:
```
[ 82%] Building CXX object src/CMakeFiles/gqrx.dir/qtgui/iq_tool.cpp.o
/__w/gqrx/gqrx/src/qtgui/freqctrl.cpp: In member function 'void CFreqCtrl::setFrequencyFocus()':
/__w/gqrx/gqrx/src/qtgui/freqctrl.cpp:885:48: error: no matching function for call to 'CFreqCtrl::mapToGlobal(const QPointF&)'
                            mapToGlobal(localPos),
                                                ^
In file included from /usr/include/x86_64-linux-gnu/qt5/QtWidgets/qframe.h:44:0,
                 from /usr/include/x86_64-linux-gnu/qt5/QtWidgets/QFrame:1,
                 from /__w/gqrx/gqrx/src/qtgui/freqctrl.h:6,
                 from /__w/gqrx/gqrx/src/qtgui/freqctrl.cpp:35:
/usr/include/x86_64-linux-gnu/qt5/QtWidgets/qwidget.h:296:12: note: candidate: QPoint QWidget::mapToGlobal(const QPoint&) const
     QPoint mapToGlobal(const QPoint &) const;
            ^~~~~~~~~~~
/usr/include/x86_64-linux-gnu/qt5/QtWidgets/qwidget.h:296:12: note:   no known conversion for argument 1 from 'const QPointF' to 'const QPoint&'
make[2]: *** [src/CMakeFiles/gqrx.dir/qtgui/freqctrl.cpp.o] Error 1
src/CMakeFiles/gqrx.dir/build.make:1263: recipe for target 'src/CMakeFiles/gqrx.dir/qtgui/freqctrl.cpp.o' failed
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:135: recipe for target 'src/CMakeFiles/gqrx.dir/all' failed
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
Error: Process completed with exit code 2.
```

argilo:
I guess the problem is that `QWidget::mapToGlobal` is new in 6.0.

vladisslav2011:
This can be fixed: https://github.com/gqrx-sdr/gqrx/commit/354a9f2db229aa5e7849f7c7bfc75094a6b1a8c9

src/qtgui/freqctrl.cpp
```diff
@@ -879,8 +879,10 @@ void CFreqCtrl::setFrequencyFocus()
     uint8_t position = floor(log10(m_freq));
     position = (uint8_t)fmax(position, 4);      // restrict min to 100s of kHz
 
+    const QPointF localPos = m_DigitInfo[position].dQRect.center();
     QMouseEvent mouseEvent(QEvent::MouseMove,
-                           m_DigitInfo[position].dQRect.center(),
+                           localPos,
+                           mapToGlobal(localPos),
```
vladisslav2011, line 885:
```
mapToGlobal(QPoint(localPos.x(),localPos.y())),
```
`mapToGlobal` expects `QPoint` instead of `QPointF` as an argument.

vladisslav2011 (changes requested):

argilo:
This code looks unnecessary. Instead of synthesizing a fake mouse event and passing it into `mouseMoveEvent`, we could just set the highlighted digit directly.

argilo:
This seems to work fine:
```diff
diff --git a/src/qtgui/freqctrl.cpp b/src/qtgui/freqctrl.cpp
index f5b2c158ad..f02de46ab6 100644
--- a/src/qtgui/freqctrl.cpp
+++ b/src/qtgui/freqctrl.cpp
@@ -879,10 +879,8 @@ void CFreqCtrl::setFrequencyFocus()
     uint8_t position = floor(log10(m_freq));
     position = (uint8_t)fmax(position, 4);      // restrict min to 100s of kHz
 
-    QMouseEvent mouseEvent(QEvent::MouseMove,
-                           m_DigitInfo[position].dQRect.center(),
-                           Qt::NoButton,
-                           Qt::NoButton,
-                           Qt::NoModifier);
-    mouseMoveEvent(&mouseEvent);
+    if (!hasFocus())
+        setFocus(Qt::MouseFocusReason);
+
+    setActiveDigit(position);
 }
```

vladisslav2011:
I think `Qt::ShortcutFocusReason` may fit better.

argilo:
Yes, I imagine so.

While we're here, it might also be worth cleaning up the `position` calculation, which unnecessarily bounces between `uint8_t` and `float`. The `setActiveDigit` argument is an `int` anyway.

argilo:
The `// restrict min to 100s of kHz` comment also doesn't match reality: 4 should be changed to 5 if we want the cursor to appear at the 100 kHz column.

src/qtgui/freqctrl.cpp
pinkavaj: Great the tests are here, I really did not paid attention to Qt5 compatibility.

pinkavaj:
Great the tests are here, I really did not paid attention to Qt5 compatibility.

argilo:
@pinkavaj Let me know if you'd like to make these changes. If not, I can clean up `setFrequencyFocus` tonight or tomorrow.

src/qtgui/freqctrl.cpp
pinkavaj: > @pinkavaj Let me know if you'd like to make these changes. If not, I can clean up `setFrequencyFocus` tonight or tomorrow.

I'm already on it, I have taken the liberty to change it a bit more, opinion welcome.

src/qtgui/freqctrl.cpp
```diff
@@ -876,13 +878,11 @@ void CFreqCtrl::cursorEnd()
 
 void CFreqCtrl::setFrequencyFocus()
 {
-    uint8_t position = floor(log10(m_freq));
-    position = (uint8_t)fmax(position, 4);      // restrict min to 100s of kHz
+    if (!hasFocus() || m_ActiveEditDigit == -1) {
+        // Select last digit or 5th digit (100s of kHz), whatever is bigger.
+        uint8_t position = MAX(int(log10(m_freq)), 5);
```
argilo, line 883: I'd suggest using `std::max` instead of a macro from `sys/param.h` (which is probably not available on all platforms, Windows in particular).

Also, `position` can be an `int`.

src/qtgui/freqctrl.cpp
vladisslav2011: That's outdated. You have pushed a better commit.

src/qtgui/freqctrl.cpp
pinkavaj: Done, thx!

argilo (approved):
The changes look good. I took this for a spin, and it works well on Linux at least.

vladisslav2011 (approved):

argilo:
> Focus only if not already focuses

This was already the case before.

> Keep the current position if focused

I guess it's debatable whether it should work that way. Prior to this change, pressing <kbd>F</kbd> would always jump to the first digit, and I don't mind that behaviour. Is it worth changing?

pinkavaj:
> > Focus only if not already focuses
> 
> This was already the case before.
> 
> > Keep the current position if focused
> 
> I guess it's debatable whether it should work that way. Prior to this change, pressing F would always jump to the first digit, and I don't mind that behaviour. Is it worth changing?

Honestly not sure, I would expect this behaviour but some users might already by accustomed to the old one. Lets revert it.

argilo:
> I would expect this behaviour but some users might already by accustomed to the old one. Lets revert it.

I think that's best. If you think the behaviour should be changed, that could be debated in a separate issue or pull request. Let's keep this one to just cleanup.

vladisslav2011 (approved):

Issue #1037, opened 2021-12-31
I/Q tool file selector randomly jumps to current selection


argilo:
Steps to reproduce:

1. Create enough I/Q recording files such that they take up more than one page in the I/Q tool.
2. Open the I/Q tool and click on the first recording file.
3. Scroll down to the bottom of the list of I/Q files.

Within one second, you'll be returned to the top of the list. This makes it very difficult to select the I/Q files you want.

This happens because `CIqTool::timeoutFunction` executes once per second and calls `CIqTool::refreshDir`, which clears and rebuilds the list, then calls `QListWidget::setCurrentRow`. I imagine there's a better way to keep the file list up to date.

vladisslav2011:
I've noticed this issue too.
It may fixed by merging https://github.com/gqrx-sdr/gqrx/pull/1011

Issue #1017, opened 2021-12-14
BUG: IQ Tool playback slider runaway when the DSP is stopped


vladisslav2011:
The IQ tool playback slider keeps running even when the DSP is stopped. After DSP restart it does not return to correct position. The time display starts showing wrong time too.

Steps to reproduce:
1. Start IQ file playback
2. Before playback finishes, stop the DSP
3. Wait some time
4. Start the DSP

Expected results:
The DSP stops, the slider stops. IQ file playback pauses. After restart, the file continues to play from the position, at which it was stopped. The slider continues to follow the actual position. The time counter shows correct elapsed time.

Observed results:
The DSP stops, the slider continues to run. After restart, the file continues to play from the position, at which it was stopped. The slider does not show the correct position. The time counter shows incorrect elapsed time.

+If the file playback does not start immediately due to buffering, the slider (and time display) do not show actual played position straight from start. Seeking the file to some previous point resolves this for one time.

PR #1013, opened 2021-12-10
Iq tool formats

vladisslav2011:
Close #881.
Includes some IQ Tool bug fixes.

PR #1011, opened 2021-12-03
Iq tool gui fixes

vladisslav2011:
Fix IQ Tool file list jumpy scrollbar.
Always choose correct sample rate when starting IQ playback.
Disable GUI controls, that have no use or may break things during playback/recording of an IQ file.

Previous in series https://github.com/gqrx-sdr/gqrx/pull/1010
Next in series https://github.com/gqrx-sdr/gqrx/pull/1013
Closes https://github.com/gqrx-sdr/gqrx/issues/1037

PR #1162, opened 2022-11-06
Allow recording and playing whole BW also to/from WAV file

pinkavaj:


vladisslav2011:
1. Have you tested recording/playback of files larger, than 2(4) Gb?
2. Isn't it better to implement different sample formats in a RAW file first #881 before adding container support? PRs #1139 #1013 or provide your better solution.
3. Isn't it better to fix horrible IQ Tool bugs (that affect usability) first #1037 #1017 ? PR #1011 or provide your better solution.

PR #1016, opened 2021-12-14
Iq tool buffered

vladisslav2011:
Implement buffered file_source/file_sink to improve IQ recording/playback success rate in case of slow media/heavy background activity.
Includes https://github.com/gqrx-sdr/gqrx/pull/1013.
This is the last PR, created from splitting https://github.com/gqrx-sdr/gqrx/pull/1008 into smaller parts.

PR #1163, opened 2022-11-06
Do not block during save()

pinkavaj:
Just do not block when not necessary, this reduced a number of "hickups" on my system a bit.

### DIFF
--- a/src/applications/gqrx/receiver.cpp
+++ b/src/applications/gqrx/receiver.cpp
@@ -1217,14 +1217,12 @@ receiver::status receiver::stop_iq_recording()
     }
 
     tb->lock();
-    iq_sink->close();
-
     if (d_decim >= 2)
         tb->disconnect(input_decim, 0, iq_sink, 0);
     else
         tb->disconnect(src, 0, iq_sink, 0);
-
     tb->unlock();
+
     iq_sink.reset();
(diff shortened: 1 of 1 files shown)


argilo:
Does this actually make a difference? I recently noticed that the `close()` function doesn't actually close the file: https://github.com/gnuradio/gnuradio/issues/6277

vladisslav2011:
The file is always getting closed in the destructor, so `iq_sink->close()` call may be omitted here.
On my system the only solution to "hickups" is buffering up to 20...30 seconds of IQ data (https://github.com/gqrx-sdr/gqrx/pull/1016)

pinkavaj:
> Does this actually make a difference? I recently noticed that the `close()` function doesn't actually close the file: [gnuradio/gnuradio#6277](https://github.com/gnuradio/gnuradio/issues/6277)

Can it be that is is calling the `fflush(d_fp);` which is blocking? If we does not insist on the `flush()` the `close()` can be really removed as suggested above. Any reason to keep it?

pinkavaj:
But it is possible the reduction in number of "hickups" was just a coincidence.

argilo:
I'm going to close this off for now, because there doesn't seem to be any evidence that it's a performance improvement.

PR #1165, opened 2022-11-08
Implement tag renaming

pinkavaj:


vladisslav2011:
Isn't it better to just actually rename the tag (update `TagInfo::name`, this should not invalidate pointers) instead of looping through all bookmarks updating the pointer to a tag?
I think, the CI wants `#include <QDebug>`.

vladisslav2011:
I've fixed the CI https://github.com/vladisslav2011/gqrx/tree/pi-rename-tag, but it does not work as expected. It tries to create a new tag instead of renaming and does wrong things when a tag is renamed to an existing name. I think, renaming should fail if the name, we are trying to rename to, already exists...

vladisslav2011:
I've added another commit to https://github.com/vladisslav2011/gqrx/tree/pi-rename-tag 
It works like it should now.
I think, we should remove 'Create New Tag' menu item from tag filter window as it does not work as expected (it does not actually create a tag, but adds it to the list and allows to filter using it).

vladisslav2011:
Adding a tag and not assigning it to a bookmark looks pointless to me. It is still possible to create a 'floating' tag, that is not assigned to a bookmark (add a tag to a single bookmark, then remove it from the bookmark), but this tag will always get filtered out in `Bookmarks::save()`.

vladisslav2011:
I have no idea why this warning is emitted...
You have not picked the CI fix.

src/qtgui/bookmarks.cpp
```diff
@@ -289,6 +289,32 @@ bool Bookmarks::removeTag(QString tagName)
...
+bool Bookmarks::renameTag(QString oldName, QString newName)
+{
+    oldName = oldName.trimmed();
+    newName = newName.trimmed();
+
+    if(oldName == TagInfo::strUntagged || oldName == "")
+        return false;
+
+    if (oldName == newName)
+        return false;
```
vladisslav2011, line 301: Lines 300..310 can be removed. I missed the same check at the line 254.
Or it may be better to remove check on line 254 as it looks better to compare trimmed strings.

pinkavaj:
> You have not picked the CI fix.

The `qDebug()` is removed now.

It is more convenient for me to point at code rather than on the person, it usually creates less friction :)

vladisslav2011:
Sorry :-)

PR #1166, opened 2022-11-08
Ensure tags are not duplicates during bookmark load

pinkavaj:
Duplicates can be introduced in several ways. User can add duplicate
tag now, but it seems tath past and current GQRX contains some bugs
related to bookmarks/tags and the saved content is sometimes crappy.


pinkavaj:
Is it ok to add such refactors, or should I create separate MR? (I would do it for bigger ones for sure).

vladisslav2011:
I think, it is better to switch to using shared pointers instead of raw pointers. It solves tag corruption  when Gqrx is built with Qt6. Take a look at this PR https://github.com/gqrx-sdr/gqrx/pull/1145

vladisslav2011:
Could you open an issue and provide exact steps to reproduce tag duplication bug?

pinkavaj:
> Could you open an issue and provide exact steps to reproduce tag duplication bug?

https://github.com/gqrx-sdr/gqrx/issues/1167 Thats for the manual method, I was able to create really a mess in tags, but I guess that might be accounted to the bug which #1145 is addressing.

pinkavaj:
Sure #1145 can be solved first, I does not mind to rebase.

PR #1168, opened 2022-11-10
Set resolution for tunning on SSB demodulator to 50 Hz

### LINKED ISSUE #920: Suggestion to change USB/LSB step size

pinkavaj:
Close #920

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1144,7 +1144,7 @@ void MainWindow::selectDemod(int mode_idx)
         rx->set_demod(receiver::RX_DEMOD_SSB);
         ui->plotter->setDemodRanges(-40000, -100, -5000, 0, false);
         uiDockAudio->setFftRange(0,3000);
-        click_res = 100;
+        click_res = 50;
         break;
 
     case DockRxOpt::MODE_USB:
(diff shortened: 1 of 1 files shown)


pinkavaj:
I'm personally not using this so much, so maybe someone with more insight may want to evaluate this?

vladisslav2011:
It may be better to make tuning step user configurable.
Add a checkbox to switch from demodulator default to user preferred tuning step and add an input (spinbox?) to set user preferred tuning step.

pinkavaj:
> It may be better to make tuning step user configurable. Add a checkbox to switch from demodulator default to user preferred tuning step and add an input (spinbox?) to set user preferred tuning step.

#928 is already trying to do this. IMHO the actual choice depends on how complex the SW should be, for what audience is intended.

vladisslav2011:
I think, it is better to have an option than to have no option...
I had a request to set SSB tuning step to 500Hz to match common ham band frequencies in my fork... So, I'll implement this anyways and maybe open a PR.

argilo:
I had a look into the code history, and left my findings and opinions here: https://github.com/gqrx-sdr/gqrx/issues/920#issuecomment-1510029578

PR #1169, opened 2022-11-10
IQ Tool: Fix jumpy scrollbar of IQ player

vladisslav2011:
When file list grows long and the user tries to scroll the list to a new position, the list continues to return it's scroll to the last selected item. That's really annoying.
Return the list to last seen position after refreshing it's contents.
Demonstration: https://www.youtube.com/watch?v=DW3LpVRxGoA

Closes https://github.com/gqrx-sdr/gqrx/issues/1037

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -4,6 +4,7 @@
        NEW: Restore FFT zoom level between sessions.
        NEW: Restore peak detect & peak hold between sessions.
      FIXED: Remove empty frame from bottom of I/Q tool window.
+     FIXED: Sudden scrolling of file list in I/Q tool window.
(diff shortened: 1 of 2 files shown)


pinkavaj (approved):
@argilo Works like a charm for me, can we get throught?

argilo:
This is definitely an improvement. Thanks!

pinkavaj:
@argilo Should we add the news line too or is it somethink you would preffer to do your-selves?

argilo:
I usually add the news items myself just before merging. Including that earlier would sometimes lead to merge conflicts, so I think it's best not to include it in pull requests.

PR #1172, opened 2022-11-11
Use swap_iq implementation from GnuRadion

pinkavaj:
This should be more efficient.

### DIFF
--- a/src/dsp/correct_iq_cc.cpp
+++ b/src/dsp/correct_iq_cc.cpp
@@ -98,6 +98,7 @@ iq_swap_cc::iq_swap_cc(bool enabled)
           gr::io_signature::make(1, 1, sizeof(gr_complex)))
 {
     d_enabled = enabled;
+#if GNURADIO_VERSION < 0x30900
     d_c2f = gr::blocks::complex_to_float::make();
     d_f2c = gr::blocks::float_to_complex::make();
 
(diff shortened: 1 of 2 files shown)


argilo:
As you noticed, the Swap IQ block is only available in GNU Radio 3.9 or later. I don't think it's worth adding conditionally-compiled code to get around this though, since #1063 seems like a better approach. I expect that IQ swapping is a rarely-used feature, so it makes sense to prioritize performance in the case where it is turned off.

vladisslav2011:
Disconnecting `iq_swap` completely gives slightly bigger performance boost, than replacing the processing block(s) with `copy`.
You may take a look at this PR: https://github.com/gqrx-sdr/gqrx/pull/1063

argilo:
Gqrx does use `#if GNURADIO_VERSION` in some other places, but only where absolutely necessary. I prefer to avoid it, because conditionally-compiled code is prone to bugs, especially when future modifications are made.

vladisslav2011:
Agree.
I thought of moving almost all conditional compilation directives to a dedicated include file at cost of slightly longer compilation due to include overhead. This may simplify maintenance. But the change is big, it will break many pending PRs and it should be discussed.

argilo:
> I thought of moving almost all conditional compilation directives to a dedicated include file

It's probably not worth the trouble.

Most of the `#ifdef`s are trivial (header file renames, `boost::shared_ptr` -> `std::shared_ptr`, new arguments added to constructors), so they're not particularly likely to cause bugs. The three exceptions are:

* `receiver`'s workaround for https://github.com/gnuradio/gnuradio/pull/3184: https://github.com/gqrx-sdr/gqrx/blob/3e3b22f28444640599c213ed0ad1aff58a1b7343/src/applications/gqrx/receiver.cpp#L211-L222
* `rx_rds`'s alternate implementation for GNU Radio 3.7: https://github.com/gqrx-sdr/gqrx/blob/3e3b22f28444640599c213ed0ad1aff58a1b7343/src/dsp/rx_rds.cpp#L85-L103
* `udp_sink`'s changes for the new UDP sink in GNU Radio 3.10: https://github.com/gqrx-sdr/gqrx/blob/3e3b22f28444640599c213ed0ad1aff58a1b7343/src/interfaces/udp_sink_f.cpp#L98-L119

GNU Radio 3.10 is shaping up to be a long-lived release, so 3.7-3.9 will fade out over the next few years and we can remove most of the `#ifdef`s. In fact, I think support for 3.7 (and about 20% of the `ifdef`s) could be dropped in the next feature release (2.16) of Gqrx.

PR #1106, opened 2022-02-17
Multi vfo upstream

vladisslav2011:
Initial multivfo implementation

Move ddc into receiver_base_cf
Move defines and modulation type to dedicated files
Implement config v4
Implement vfo selection by clickin the plotter widget
Implement stereo panning feature to improve listening experience
Implement locking vfos to a frequency and automatically delete locked
vfos when moved out of bandwidth
Add global lock/unlock context menu
Add global squelch auto/reset popup menu
Implement settings loading from a bookbark by right-clicking bookmark
tag on a plotter
Implement new vfo creation from a bookbark by middle-clicking bookmark
tag on a plotter
Add bookmarks context menu actions
1. To tune current demod to a frequency
2. To tune current demod to a frequency and load params from the
bookmark
3. To create a new demod with the bookmark params

Prevent new vfo creation over already active bookmark
Replace selected bookmark instead of adding new one at the same frequency
Update plotter overlay on bookmark delete
Update plotter overlay on bookmarks save
Save bookmarks on tag delete
Store all receiver parameters to a bookmark
Add a button to copy recording settings to all VFOs
Sort bookmarks by any column
Validate and correct filter params on input
Make Modulations class fully static
Allow to choose bookmarks table visible columns
Select a bookmark on exact freq match only to avoid a situation when
2 bookmarks have overlapping filter passbands and wrong bookmark with
wider filter passband is activated.
Automatically create demodulators when a bookmark, marked as 'auto' gets
visible on the panadapter.
Add a switch to enable/disable this function.

PR #1174, opened 2022-11-17
Allow activate bookmark by doubleclik in plotter

### LINKED ISSUE #214: Selecting a bookmark doesn't retrieve all saved settings
When selecting a bookmark that is within the currently sampled range, not all settings stored in the bookmark are retrieved. e.g., If 154.025MHz is a stored bookmark that is within the frequency range I am currently watching on the waterfall, and I select it, the filter bandwidth and mode are not set as stored within the bookmark. Instead, they remain at whatever they were set to before. However, if I select a bookmark that is outside this range, the settings are retrieved.

pinkavaj:
Close #214

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -275,6 +275,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
 
     // Bookmarks
     connect(uiDockBookmarks, SIGNAL(newBookmarkActivated(qint64, QString, int)), this, SLOT(onBookmarkActivated(qint64, QString, int)));
+    connect(ui->plotter, SIGNAL(newBookmarkActivated(qint64, QString, int)), this, SLOT(onBookmarkActivated(qint64, QString, int)));
     connect(uiDockBookmarks->actionAddBookmark, SIGNAL(triggered()), this, SLOT(on_actionAddBookmark_triggered()));
 
     //DXC Spots
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -176,6 +176,23 @@ QSize CPlotter::sizeHint() const
     return {180, 180};
 }
 
+void CPlotter::mouseDoubleClickEvent(QMouseEvent* event)
+{
+    if (m_CursorCaptured == TAG)
+    {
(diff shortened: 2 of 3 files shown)


vladisslav2011:
Isn't it better to use right-click, like it is done here: https://github.com/gqrx-sdr/gqrx/commit/ff69939426af331e53925e4c4a6b1302a45719ff ?

pinkavaj:
> Isn't it better to use right-click, like it is done here: [ff69939](https://github.com/gqrx-sdr/gqrx/commit/ff69939426af331e53925e4c4a6b1302a45719ff) ?

It is definitely an option. My conclusion was doubleclick will produce more consistent behavior. AFAIK rightclick is now used to either pop-up menus or to reset number to zero in frequency selection box. Doubleclick is used in bookmarks dock to activate the bookmark. This is not menu but does aktivate the bookmark.

Doing UI might get tricky and honestly I did not get through whole app to find out every usage of double/righ-click. I'm open to discussion but for now confident this good choice.

vladisslav2011:
Doubleclick in a bookmarks dock by default switches the cell to editing state, so using it to activate a bookmark is not the best possible option. It may better to use a context menu like it is done here: https://github.com/gqrx-sdr/gqrx/commit/f8c78569d44bd85e4e5baf42e76fcdad00b584c5
It may be better to add a context menu to a plotter. I decided to not do so and leave default right click to reset zoom/offset action and override right click over a bookmark tag only.
Consider taking a look at this branch https://github.com/vladisslav2011/gqrx/tree/multi_vfo_unsquashed and suggesting commits, that worth opening a separate PR to make https://github.com/gqrx-sdr/gqrx/pull/1106 smaller and easier to review.

pinkavaj:
Wow, it seems to be even more complicated. On build from latest master (`53bd3c50`) I observe this behavior in bookmark dock:

#1175 

Do you observe the same behavior? @vladisslav2011 

My gut feeling is this should be make consistent in the first place.

vladisslav2011:
I can confirm the same behavior.
This behavior is fine to me, so I have not opened an issue about it.

There may be at least 2 possible actions, that may be assigned to a doubleclick: just tune to a bookmarked frequency and tune to a bookmarked frequency and load settings. There may be the third action in the future: to create a VFO at a bookmarked frequency.

Which action should be executed by default?
Should we add an alternate action (or both) to a context menu?

PR #1185, opened 2023-01-18
Fix MacOS CI

### LINKED ISSUE #1184: The MacOS CI script fails to install dependencies
Steps to reproduce:

1. Fork repository
2. Enable the CI
3. Create a branch
4. Push a commit with a no-breaking change (fix code formatting somewhere for example) to a newly created branch to trigger the CI

Expected behavior:

CI passes with all green checkmarks

Observed behavior:

MacOS CI failes with the following error
```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.11
Target /usr/local/bin/2to3-3.11
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.11'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.11

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.11
```

vladisslav2011:
Relink both python versions overwriting old symlinks.
This should be fixed on the brew side and this fix should be removed afterwards.
Fix #1184

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,8 +77,11 @@ jobs:
       - name: Install dependencies
         run: |
           brew update
+          brew install python@3.11 || true
+          brew link --overwrite python@3.11
+          brew install python@3.10 || true
+          brew link --overwrite python@3.10
           brew install airspy boost gnuradio hackrf libbladerf librtlsdr pybind11 uhd qt@6
-
           cd /tmp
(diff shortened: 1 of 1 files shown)


argilo:
It appears that Homebrew has fixed the issue, as CI now runs on the master branch without issue.

vladisslav2011:
Good news.

Issue #263, opened 2015-09-04
Edited bookmark name is not applied immediately to spectrum display


darrenfreeman:
Using the latest source from git, under Linux/KDE.

Tune in to a bookmark.

Edit the name of the bookmark.

The spectrum display shows the old bookmark name until the frequency is changed. Then it updates to the new name.

csete:
This is also the case if a bookmark is deleted.

darrenfreeman:
Yes, I discovered that later too :)

kolbma:
Would be fixed in https://github.com/kolbma/gqrx/tree/bookmarks-ext

Issue #965, opened 2021-07-04
Deselecting bookmark tags should hide them from panadapter


Rua:
When unchecking bookmark tags, they disappear from the bookmarks list, but they stay visible above the panadapter. It would be nice if they were hidden from the panadapter too. That would help when there's many bookmarks.

argilo:
Fixed in #1222.

PR #1188, opened 2023-02-09
Redraw panadapter bookmark marks on bookmark adding/deletion/tag selection

vladisslav2011:
Update plotter overlay on bookmarks change (connect signal `Bookmarks::BookmarksChanged()` to `CPlotter::updateOverlay()`).
Emit the signal on bookmarks deletion/adding.
Implement panadpter bookmark mark tag filtering.
May close https://github.com/gqrx-sdr/gqrx/issues/263 and https://github.com/gqrx-sdr/gqrx/issues/965

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -5,6 +5,7 @@
        NEW: Set/get audio gain via remote control.
      FIXED: Loading of narrow FM tau setting.
      FIXED: Crash when adding or removing bookmark tags.
+     FIXED: Redraw bookmarks immediately after changes.
    REMOVED: Support for GNU Radio 3.7.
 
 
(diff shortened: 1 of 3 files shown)


argilo:
> Implement panadpter bookmark mark tag filtering.

That change looks unrelated to this PR. Could you move it to its own PR?

By the way, it seems the original intent was to just use the `IsActive` method for this purpose: 459f5552c9c6848e7082270b9751f5e300bd4e0f. I'm not sure why `if(info.IsActive())` was left commented out, but we should probably use the existing method, cleaning it up if necessary.

vladisslav2011:
> Could you move it to its own PR?

No problem. https://github.com/gqrx-sdr/gqrx/pull/1222

> I'm not sure why `if(info.IsActive())` was left commented out

I think, it did not work correctly without changes from this PR.

PR #1193, opened 2023-03-11
Add min hold line on plotter

willcode:
min-hold


vladisslav2011 (approved):
I don't know, how this may be used, but it builds and works (even rebased on top of heavy-modified branches).
min-hold-test

willcode:
My use is (1) seeing the noise floor more clearly in some cases, and (2) seeing things like the pilot tone in the FM signal pictured. It's not a major win, but on the other hand it was just a few lines of code to put it in.

argilo:
I suppose this might also help to distinguish between continuous and intermittent transmitters.

It does add some clutter to the display though, so maybe it would be a good idea to have a separate button for it.

willcode:
Added a second button. I tried to make the smallest possible change to the GUI that made sense.

Gqrx set records for the number of changes required to add a single button. If I were to start from scratch (it has occurred to me) a common settings bus would be much better than all this inter-component signal/slot wiring.

PR #1195, opened 2023-03-15
Increase max fft size to 4M and reduce scope of locking in rx_fft

willcode:
Larger ffts would cause overruns (testing done with a B205mini) and the choke point was the write into the fft buffer. While the buffer data structures need to be protected by a mutex, the fft execute() does not, since it uses a separate buffer.

Further, since it is not critical which data gets fft'd (the timer is based on system time anyway), apply_window() can be done outside the critical section. There is some chance that the reader and writer will collide while apply_window() is being executed. I've made the assumption that the results won't be visible to the user and speed/overruns are of more interest.

Increasing the number of threads had some effect on the achievable fft rate for larger fft size, but was not the reason for the drops, so settable nthreads can be left to another commit.

### DIFF
--- a/src/dsp/rx_fft.cpp
+++ b/src/dsp/rx_fft.cpp
@@ -94,17 +94,23 @@ int rx_fft_c::work(int noutput_items,
     const gr_complex *in = (const gr_complex*)input_items[0];
     (void) output_items;
 
-    /* just throw new samples into the buffer */
-    std::lock_guard<std::mutex> lock(d_mutex);
+    gr_complex *write_pointer;
 
+    /* just throw new samples into the buffer */
     int items_to_copy = std::min(noutput_items, (int)d_writer->bufsize());
     if (items_to_copy < noutput_items)
         in += (noutput_items - items_to_copy);
 
-    if (d_writer->space_available() < items_to_copy)
-        d_reader->update_read_pointer(items_to_copy - d_writer->space_available());
-    memcpy(d_writer->write_pointer(), in, sizeof(gr_complex) * items_to_copy);
-    d_writer->update_write_pointer(items_to_copy);
+    {
+        std::lock_guard<std::mutex> lock(d_mutex);
+
+        if (d_writer->space_available() < items_to_copy)
+            d_reader->update_read_pointer(items_to_copy - d_writer->space_available());
+        write_pointer = (gr_complex *)d_writer->write_pointer();
+        d_writer->update_write_pointer(items_to_copy);
+    }
+
+    memcpy(write_pointer, in, sizeof(gr_complex) * items_to_copy);
 
     return noutput_items;
 }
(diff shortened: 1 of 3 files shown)


vladisslav2011:
Hello.

> Larger ffts would cause overruns (testing done with a B205mini) and the choke point was the write into the fft buffer.

What is your input rate? Does it happen at low sample rates too?

I can't confirm. Just increasing the `MAX_FFT_SIZE` to 4M and adding new values to a combo box worked to me in my fork. No dropouts on B200/B210 @56Msps (with proper buffering) and RX888 @128Msps.
I had an issue with insufficient buffering and fixed it this way: https://github.com/vladisslav2011/gqrx/commit/8e07fccc18169e2b54bc0d669a5a45dbf247abee . This may or may not be related to dropouts at larger FFT sizes. I'll rebase https://github.com/vladisslav2011/gqrx/commit/cee77684c6ab1e4353dd4ecaca8736e9a024c547 on top of master and test.

B210 always shows some dropped samples (000 in console) regardless of FFT size, but this can be fixed by adding `num_recv_frames=256` to a device line in "Configure I/O devices" dialog.

> There is some chance that the reader and writer will collide while apply_window() is being executed.

I think, this may be a problem... How often does this occur?

willcode:
I haven't seen any evidence that the reader and writer collide. The typical cases would be (1) reader keeps up and is waiting on the writer, or (2) the buffer is always full and the writer doesn't lap the reader in the time it takes the reader to process. Given that the FFT is only used for visuals, I'm going with an empirical test. If we want to be formally correct, `apply_window()` can go back in the critical section, but it takes time.

`num_recv_frames` did help - thanks - now running at 61.44 MS/s with almost no drops on the B205mini, 4M FFT runs at about 5Hz. This is where threading might start to help.

willcode:
I've got demod off, so resampler wouldn't affect anything.

willcode:
With the `num_recv_frames` set, I still get one overflow when the dsp starts, without the locking change, and none with it. This is not a huge difference, of course.

The PR with `apply_window()` moved back into the critical section also resulted in an overrun on startup ... no overrun when it is moved back outside.

vladisslav2011:
Test results with demod off, cherry-picked https://github.com/vladisslav2011/gqrx/commit/cee77684c6ab1e4353dd4ecaca8736e9a024c547 on top of current master:
https://www.youtube.com/watch?v=hrcKJ4E2IZU

I've found one more commit, that may help to avoid dropouts, not related to Gqrx.
https://github.com/vladisslav2011/gr-osmosdr/commit/1888972533cb57dfa1ee0fa9f5e18fda06040a9b
It just allocates some buffer space regardless of what next block wants. Quick, dirty fix. It may be better to add a configuration option to turn it on/off.
This was the first attempt to get rid of dropouts during 4 major GNSS satellites constellations reception with GNSS-SDR... Then I discovered `num_recv_frames` and got continuous tracking times of several hours.

vladisslav2011:
> With the `num_recv_frames` set, I still get one overflow when the dsp starts, without the locking change, and none with it.

Hmmm. I have no overflows in both cases. This test was done with unmodified gr-osmosdr:

https://www.youtube.com/watch?v=jPNPTxUhw-E

Several dropouts at DSP start are not a big problem. Maybe your system is slightly less powerful, than mine...

PR #1196, opened 2023-03-17
Keep 2D plot above frequency labels

willcode:
Frequency labels could be hard to read when the plot overwrote them.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -905,6 +905,8 @@ void CPlotter::draw()
     int     h;
     int     xmin, xmax;
 
+    QFontMetrics metrics(m_Font);
+
     if (m_DrawOverlay)
     {
         drawOverlay();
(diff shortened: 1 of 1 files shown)


vladisslav2011:
Hello.
I have tested it and it does not work well...
plot-above-freq-labels
Wasting vertical screen space just to improve numbers readability does not look like a good idea.
It may be better to draw the overlay over the plot instead of drawing the plot over the overlay copy. 
Something like drawing `m_2DPixmap` before drawing `m_OverlayPixmap` in `CPlotter::paintEvent`.

willcode:
Open to better ways. I wasn't thinking of re-ordering the drawing of layer, though. That seems like it could cause other problems. Personally, I like being able to read the numbers. Other people might have small screens, though.

Hmm, maybe there are some fixed constants in there somewhere that don't match up with the QFontMetrics I used.

willcode:
Had a git accident with this branch ... submitted a new PR, hopefully without the problem shown above.

PR #1197, opened 2023-03-17
dk.gqrx.gqrx.desktop: fix typo in Czech translation

AsciiWolf:
The correct term for Receiver in Czech is "Př**i**jímač", not "Př**í**jímač". See [this Wikipedia page](https://cs.wikipedia.org/wiki/P%C5%99ij%C3%ADma%C4%8D) and [this Wiktionary page](https://cs.wiktionary.org/wiki/p%C5%99ij%C3%ADma%C4%8D).

/cc @jiri-pinkava @argilo


argilo:
I don't speak Czech, but this appears to be correct and the number of Google hits is much higher with the updated spelling. Merging.

PR #1198, opened 2023-03-18
When peak hold is selected, peak detect uses peak hold data

willcode:
Goal is to make it easier to select peaks from intermittent signals (packet, PTT, etc.)

A sliding window is used for peak detection when using peak hold. The original algorithm is still used when peak hold is off.

Peak detection is moved to after peak hold drawing, so the peak hold data is available for use.

There was also an off-by-one in the drawing and detection of peaks. While I don't understand the exact nature, adding 1 to the detected bin seems to have fixed it.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1053,7 +1053,7 @@ void CPlotter::draw()
                         if (j != i && detectSource[j] < maxInWindow)
                             maxInWindow = detectSource[j];
                     if (d < maxInWindow)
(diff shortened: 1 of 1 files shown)


willcode:
Removed the off-by-one hack, as the cause is fixed in https://github.com/gqrx-sdr/gqrx/pull/1203.

vladisslav2011:
Interesting idea.
Tested. Works.
peak-hold-detect
Some users may find it useful.

vladisslav2011 (approved):

PR #1200, opened 2023-03-19
Shortcut key "z" zeros the rx filter offset

willcode:
Instead of entering all zeroes in the filter offset field, hit "z" to re-center the cursor.


argilo:
This seems like a useful feature. I do sometimes find myself going to the "Receiver Options" pane just to type in zeroes.

I dug into the CI failure, and it seems to be an issue with the gnuradio formula in Homebrew: https://github.com/Homebrew/homebrew-core/issues/126118

I'll merge once that's sorted out.

vladisslav2011 (approved):

argilo:
I noticed a small bug with this: if "Z" is pressed while DSP is stopped, the plotter does not redraw.

PR #1201, opened 2023-03-19
Make zoom reset more reliable

willcode:
Especially at higher zooms, right click on the freq axis did not fully reset the zoom level on the fft dock. Add an emit to resetHorizontalZoom() as is done already in zoomStepX().


vladisslav2011 (approved):

argilo:
Thanks! This works as expected.

PR #1202, opened 2023-03-21
Add AB markers, measurement, and selection from current/peak fft

willcode:
Markers A and B can be set from the filter frequency location (red line) using the "->" buttons and cleared with the "x" buttons. When both A and B are active, delta and center frequency are displayed.

Shift-click directly underneath the plot selects the span around the cursor, at the dB level of the cursor. Ctrl-shift-click does the same thing using the peak hold data, if enabled.

Markers can be dragged.

Also added a delta to filter frequency in the cursor text.

markers

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -90,6 +90,16 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
 
     setWindowTitle(QString("Gqrx %1").arg(VERSION));
 
+    // Set fixed widths for labels so they don't move around when set
+    QFontMetrics metrics(font);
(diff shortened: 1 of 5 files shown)


vladisslav2011:
Adding manual cursors to a plot... good idea.
Tested. Works.
ab_markers
Some thoughts:
It's another waste of plotter vertical screen space that I personally don't like. Isn't it better to add marker controls to a dock (FFT?)
The color of markers is hardcoded. Isn't it better to add an option to change it (pink for A, blue for B for example) like the plot color?
Marker frequencies are not saved into a config file. Another possible improvement.

I have already implemented "sort-of-markers" in multi_vfo branch: VFOs with "Demod OFF" are just markers. So, I'll steal an idea from here and add some information to a VFO tooltip :-)

PR #1203, opened 2023-03-22
Perform double precision math in xFromFreq() and freqFromX()

willcode:
An on-screen off-by-one pixel problem turned out to be rounding errors in one or both of these routines. Force all math to be double precision and round to integer results.

There's probably some more elegant way to code this, but I don't think the compiler cares, and it's easier to compare to the original computations this way.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1565,9 +1565,11 @@ void CPlotter::makeFrequencyStrs()
 // Convert from frequency to screen coordinate
 int CPlotter::xFromFreq(qint64 freq)
 {
-    qint64 w = width();
-    qint64 StartFreq = m_CenterFreq + m_FftCenter - m_Span / 2;
-    int x = (int) (w * (freq - StartFreq) / m_Span);
+    double w = width();
+    double startFreq = (double)m_CenterFreq
+                       + (double)m_FftCenter
+                       - (double)m_Span / 2.0;
+    int x = qRound(w * ((double)freq - startFreq) / (double)m_Span);
     if (x < 0)
         return 0;
     if (x > (int)w)
@@ -1579,7 +1581,8 @@ int CPlotter::xFromFreq(qint64 freq)
 qint64 CPlotter::freqFromX(int x)
 {
     double ratio = (double)x / (double)width();
-    qint64 f = (m_CenterFreq + m_FftCenter - m_Span / 2) + ratio * m_Span;
+    qint64 f = qRound64((double)m_CenterFreq + (double)m_FftCenter
+                        - (double)m_Span / 2.0 + ratio * (double)m_Span);
     return f;
 }
 


vladisslav2011:
Tested with a lot of bookmarks. Works.
Funny thing - it's actually a bit faster, than previous fixed-point implementation on my machine.

vladisslav2011 (approved):

src/qtgui/plotter.cpp
```diff
@@ -1579,7 +1581,8 @@ int CPlotter::xFromFreq(qint64 freq)
 qint64 CPlotter::freqFromX(int x)
 {
     double ratio = (double)x / (double)width();
-    qint64 f = (m_CenterFreq + m_FftCenter - m_Span / 2) + ratio * m_Span;
+    qint64 f = llround((double)m_CenterFreq + (double)m_FftCenter
+                       - (double)m_Span / 2 + ratio * (double)m_Span);
```
pinkavaj, line 1585: mey-be, just for consistenci, use `2.0` here as well (used in the above).

willcode: It really makes no difference since `m_Span` is already cast to double. I could make them all the same to make it prettier.

pinkavaj: thats it, it is just NIT to make ti prettier, LGTM from me.

argilo:
One nitpick: Qt provides `qRound` and `qRound64` functions which return `int` and `qint64` respectively. Those might perhaps be better choices than `lround` and `llround`. I see that `qRound64` is used in `CPlotter::zoomStepX`.

willcode:
Changed to `qRound[64]()` and `2.0`.

argilo (changes requested):
After thinking about this some more, I realized that we probably shouldn't be representing the plotter's x coordinate as an integer anymore. Qt has supported high-DPI displays for a while, and switched to using floating point numbers to represent screen coordinates in terms of device-independent pixels. Much of Gqrx has been updated to the new system and its associated APIs (e.g. `QPointF` in place of `QPoint`), but integer coordinates still remain in a few places, including in `xFromFreq()` and `freqFromX()`.

I think the right approach would be to represent `x` as a `float` in these methods, and update the callers to use floating-point coordinates as well.

@willcode If you want to take a crack at this, you're welcome to do so. Otherwise I'll try to look at it soon.

willcode:
I'll take a look at it. I'm pretty deep into the plotter code right now, and have also gotten a pretty good performance improvement. At the lower end fft sizes, 250 fps works. 4M will run at 10 fps. It turns out that VOLK is not needed in many places, and sometimes is slower, than plain old C++, if written for performance.
It wouldn't hurt to merge this fix in the mean time.
gqrx-fft-features

argilo:
Agreed that this is still an improvement, even without switching to float now. I'll merge once I've had a chance to test, and will throw in a note to remind myself that this code still needs to be updated with high DPI support.

argilo:
I took this for a spin, and noticed a bug: bookmarks are sometimes one pixel off from frequency grid lines even when their frequencies are equal, because the grid line positions are calculated (in various places) without rounding, for instance:

https://github.com/gqrx-sdr/gqrx/blob/803b90e1babc694d906d20ccc9afe93a621b27b2/src/qtgui/plotter.cpp#L1421

Since it would take some more work to convert everything to rounded integer coordinates, maybe it would be best to just skip straight to high-DPI-friendly floating point coordinates everywhere.

willcode:
I'm working on the conversion to floating point in the plotter-work PR so this one can be absorbed. It's a bit of a process ... getting there.

PR #1206, opened 2023-03-23
iq_swap_cc: implement directly instead of as a hier block

### LINKED ISSUE #1205: iq_swap_cc expensive when off
The `iq_swap` block is always inline. When swap is turned off, samples are sent through a c2f then f2c block, to no effect. These blocks, together, also count for a large portion of the processor utilizaiton, whether or not they are acting as a pass-through.

https://github.com/gqrx-sdr/gqrx/blob/5588106e8ee2c8c20ada759a2199876e8f54803d/src/dsp/correct_iq_cc.cpp#L112

Make it possible to bypass the `iq_swap` block. The `dc_corr` already requires reconfiguration when the DC Cancel box is checked, so why not save a lot of processor in the most common "not swaped" case.

One simple solution: since the c2f and f2c blocks are just wrappers around single volk functions, use the volk functions directly in `iq_swap_cc`. This saves 2 of 3 buffers, and in the "not swapped" case it just a single copy.

willcode:
Much faster, especially when off, and fewer buffers.

Saves about a half a core on my system at 56 MB/s.

Fixes #1205

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -10,6 +10,7 @@
      FIXED: Aliasing when input rate is higher than 2 Msps.
   IMPROVED: AGC performance.
   IMPROVED: WFM stereo & OIRT performance.
+  IMPROVED: I/Q swap performance.
   IMPROVED: Apply amplitude normalization to FFT window functions.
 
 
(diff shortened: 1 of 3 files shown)


argilo (approved):
I verified that this reduces CPU utilization, whether I/Q swap is on or off. It's also great that the new version is shorter.

PR #1207, opened 2023-03-23
stereo_demod: reduce CPU consumption

vladisslav2011:
Increase pilot FIR filter rolloff to reduce the order and taps size. This should not increase noise as pll_refout is itself a very narrow filter.
With this fix my small old laptop (Samsung NC10) is able to receive FM stereo transmissions even at 2.4Msps with almost no skipping. Without this fix 240ksps max stable.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -8,6 +8,7 @@
      FIXED: Sudden scrolling of file list in I/Q tool window.
      FIXED: Reset zoom slider after right click on panadapter / waterfall.
   IMPROVED: AGC performance.
+  IMPROVED: WFM stereo & OIRT performance.
   IMPROVED: Apply amplitude normalization to FFT window functions.
 
 
--- a/src/dsp/stereo_demod.cpp
+++ b/src/dsp/stereo_demod.cpp
@@ -68,12 +68,12 @@ stereo_demod::stereo_demod(float input_rate, float audio_rate, bool stereo, bool
     if (!d_oirt)
     {
         d_tone_taps = gr::filter::firdes::complex_band_pass(
-                                       20.0,         // gain,
+                                       1.0,          // gain,
 		                                   d_input_rate, // sampling_freq
                                        18980.,       // low_cutoff_freq
                                        19020.,       // high_cutoff_freq
-                                       1000.);       // transition_width
-        pll = gr::analog::pll_refout_cc::make(0.001,    // loop_bw FIXME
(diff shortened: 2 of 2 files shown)


argilo:
The number of taps does appear to be excessive: 579 for WFM stereo and 5781 (!!!) for WFM oirt. I previously reduced the number of taps for WFM stereo in 93dcce14b4ed28a206e242edae8c94e1f845865c, but it makes sense to go further and apply the same change to OIRT. :+1:

I think 3000 would be a safer choice for the transition width. At 5000 I'm seeing some of the audio making it into the filtered signal, which might degrade the signal quality.

The gain was intentionally changed to 20.0 in 93dcce14b4ed28a206e242edae8c94e1f845865c to normalize the pilot to an amplitude of 1.0; some GNU Radio blocks expect signals to be normalized to this level (although I'm not sure if the PLL block is one of them). It would probably make sense to apply this change to OIRT as well.

The cutoff frequencies should probably be left as they are, since widening them increases the width of the overall filter unnecessarily.

What is the reason for the change to the loop bandwidth? Have you measured the impact of this change? It might be best to leave this to a separate PR, unless it's somehow linked to the FIR filter change.

It may be possible to get a further performance improvement by switching from an FIR filter to an FFT filter. If you'd like to try this, I'd suggest doing it in a separate PR so that it can be investigated on its own.

vladisslav2011:
> some GNU Radio blocks expect signals to be normalized to this level

The `pll_refout_cc` does not require input signal normalization. It can lock to a very weak signal and it's output is always normalized to full scale.

> What is the reason for the change to the loop bandwidth?

This improves PLL filtering performance. I think, we can remove the FIR filter completely without noticeable SNR degradation if PLL loop bandwidth is set to 0.0002...0.0001. It may be better to set it to 0.00001, but this results in locking issues independent of FIR width. Maybe broadcast stations' pilot jitter is too high in my location.

> Have you measured the impact of this change?

Yes. With this flowgraph:
[hackrf-fm-stereo2.grc.tar.gz](https://github.com/gqrx-sdr/gqrx/files/11205160/hackrf-fm-stereo2.grc.tar.gz)

I agree, that FIR bandwidth may be left intact as the number of taps depends only on transition width.
I'll update the PR with less changes.

argilo:
Thanks for the updates. This seems to work well, and reduces CPU utilization noticeably.

PR #1208, opened 2023-03-23
Space freezes the plotter

willcode:
The drawing routines keep running, so when Space is hit again, nothing has been lost.

Currently, various controls do not work as expected when the plotter is frozen. Area for improvement - would likely require a third drawing layer (not a bad idea if performance allows).

Two possible uses:
- Look at something for a second but don't want to miss new stuff or shut off rx
- Save bandwidth when operating VNC/RDP


vladisslav2011:
This breaks too many things.
It may be better to increase `receiver::iq_fft` history to maximum possible value, take a snapshot to a buffer when space is pressed and render waterfall/panadapter from captured buffer on resize/retune/FFT size/window changes/etc., like it is done in my ' WF redraw' commit sequence (not extracted to a branch/PR from 'fft_channelizer' branch yet).

willcode:
Freezing the waterfall is still useful. I've often tried to put the cursor on a moving burst. This is still a WIP. It was starting to take too much code to make the entire plot consistent while frozen, so I gave up on freezing the 2D for now.

PR #1211, opened 2023-03-27
Scale FFT to dB/MHz

willcode:
FFT power was scaled using 1/(fftsize^2). This works out approximately for fftsize 1M, but was causing the power levels to shift at other sizes.

The plotter shows the maximum value of all those mapped to a pixel on the screen. This is useful, but does not represent the average power for the bins mapped to that pixel. The code was probably written to make the answer on the screen look correct in this situation, and show the noise floor averaging down at higher fftsize.

Separate modes to show the average power in a fft bin, and the minimum power, are the subjects of a separate commit.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1375,9 +1375,8 @@ void MainWindow::iqFftTimeout()
         return;
     }
 
-    // NB: without cast to float the multiplication will overflow at 64k
-    // and pwr_scale will be inf
-    pwr_scale = 1.0 / ((float)fftsize * (float)fftsize);
+    // Scale to dB/MHz
+    pwr_scale = 1.0 / 1000000.0 / (float)fftsize;
 
     /* Normalize, calculate power and shift the FFT */
     volk_32fc_magnitude_squared_32f(d_realFftData, d_fftData + (fftsize/2), fftsize/2);


vladisslav2011:
Hello.

> but was causing the power levels to shift at other sizes.

It depends on the signal and window type:

4m
1m

128k
32k
8k

Different FFT sizes, but the level is the same.

See resolved discussion here: https://github.com/gqrx-sdr/gqrx/pull/1190

willcode:
This is a separate bug, in addition to your window normalization, which was also necessary.

willcode:
Yes, it changes the max(bin value for each pixel). But that's the way it's supposed to be. The problem is that max() is not always the right measurement. If you use avg(), the power levels stay the same. You might want to see max() to see the narrowest signals with a large fft, or you may want to see avg() to see average power levels.

vladisslav2011:
Hmmm. This PR makes Gqrx unusable for measurements to me (it shows completely wrong levels), so I'll revert it in my fork if it will get merged :-)

vladisslav2011:
I have seen this code.
These 2 lines https://github.com/vladisslav2011/gqrx/commit/d466c7c4106a74b68278135ad0525ebfa2fa95ee#diff-fcf4675691a139e7f1a9f0884be68dd18e1daa93b3da5517d039e0c6c935d852R94-R95 do the same thing.
See resolved discussion here: https://github.com/gqrx-sdr/gqrx/pull/1190

willcode:
If you're interested, try that change along wiht https://github.com/gqrx-sdr/gqrx/pull/1212 in Sampling=Average mode. That uses the actual average power over bins mapped to a pixel, and I think it works for measurement.

vladisslav2011:
I tried. It shows peak level of -40dbfs when the signal level is at -3dbfs. Increasing the frontend gain by 1..2db makes clipping products appear, so -3 dbfs is correct and -40dbfs is wrong.

I don't think that wasting CPU cycles to calculate wide FFT first and wasting CPU cycles again to calculate an average is a correct way to get average power per plot pixel... Just selecting flattop window and reducing FFT size to be close to the screen width does the same thing at much less cost.
On the other hand averaging gives a nice clean FFT plot with all narrow weak signals removed... I'll rebase it and test with IQ Tool "plot on seek on pause" feature to see if it will be useful to me.

willcode:
Removing the 1000000.0 would fix that.

willcode:
Ugh, GNU Radio's Frequency Sink gives different amplitudes for different fft sizes. Now, I have to go figure out what's going on there too.

willcode:
I'll close this one - not really sure what the solution should be at the moment.

PR #1212, opened 2023-03-27
User selectable max/avg/min sampling modes for fft scaling

willcode:
When the fftsize is greater than the width of the plotter (pixels) the contents of the fft are resampled. Currently, the maximum of all the fft bins mapped to a pixel is shown. This commit adds user-controllable max/avg/min.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -65,6 +65,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     QMainWindow(parent),
     configOk(true),
     ui(new Ui::MainWindow),
+    d_display_dbm(false),
     d_lnb_lo(0),
     d_hw_freq(0),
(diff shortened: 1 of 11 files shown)


vladisslav2011:
I've rebased it on top of "fft_channelizer"+some performance counters. Had to made a way through a number of merge conflicts and segfaults... https://github.com/vladisslav2011/gqrx/tree/fft_channelizer%2Bfft-sampling-mode
Test results: https://www.youtube.com/watch?v=blpKZIrmzKc
It works. No noticeable  performance degradation. But "min" and "avg" modes are not very useful when hunting for some weak signals...

willcode:
True, I think avg would be useful for power measurements once we figure out the "right" way to do power calculations. The min mode doesn't seem all that useful, but I thought I'd try it out.

Did you find any segfaults due to the feature, or all rebase challenges?

vladisslav2011:
The PR is fine.
`*xmin`, `*xmax` updates were lost during rebase. That's why I encountered some segfaults.

PR #1213, opened 2023-03-28
Power scaling for dBm/Hz into 50 ohms

willcode:
Scale PSD as dBm/Hz into a 50 ohm load. The current code assumes that bandwidth is same as FFT size, which is not usually the case. Also, most work with spectrum analyzers in radio is done using units of dBm rather than dBW. A spectrum analyzer compensates for load, typically 50 ohms in radio.

Playing white noise at 1MS/s, a level of -60 dBW is expected. Compensating for dBm into 50 ohms (+13dB), levels are seen to be -37dB on Gqrx at higher at higher FFT sizes. At lower sizes, max hold shows -37dB with averaging turned all the way down.

Experiments with a DC "tone" looked "about right". A tone always manages to be spread over bins or averaged in a bin, so the best test I could find was a 262144 bin FFT with a DC file played at 262144 S/s. A level of -10dB was shown. I was hoping for -7dB. The tone may have been split between bins, even though I thought DC was a single bin.

We don't really have a concept of RBW, other than rate/fftsize, which is different than a spectrum analyzer.

scale-factor-dbm-50

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1364,7 +1364,7 @@ void MainWindow::iqFftTimeout()
 {
     unsigned int    fftsize;
     unsigned int    i;
-    float           pwr_scale;
+    double          quad_rate = rx->get_input_rate() / rx->get_input_decim();
 
     // FIXME: fftsize is a reference
     rx->get_iq_fft_data(d_fftData, fftsize);
@@ -1375,9 +1375,8 @@ void MainWindow::iqFftTimeout()
         return;
     }
 
-    // NB: without cast to float the multiplication will overflow at 64k
-    // and pwr_scale will be inf
-    pwr_scale = 1.0 / ((float)fftsize * (float)fftsize);
+    // Scale for dBm into 50 ohms
+    const double pwr_scale = 1000.0 / (2.0 * (double)fftsize * quad_rate * 50.0);
 
     /* Normalize, calculate power and shift the FFT */
     volk_32fc_magnitude_squared_32f(d_realFftData, d_fftData + (fftsize/2), fftsize/2);


vladisslav2011:
Hello.
Different devices have different gains and different input levels for 0 dBfs (it may depend on frequency too as generic SDR devices are not calibrated and may have input impedance far from 50+0j Ohm), so it's impossible to make S-meter/FFT show levels in dBm into 50 Ohm load in generic case.
I think, it will be better to leave FFT scaling as is and add a calibration option to let the user shift the Y-axis and S-meter display to make panadapter/S-meter levels match dBm/dBW when using specific device.
It will be even better to implement frequency-dependent calibration to make Gqrx show correct levels independent of center frequency. And temperature-dependent gain compensation with temperature sensor interface...
That's why professional spectrum analyzers cost so much.

willcode:
It's not just for noise. If you look at anything that's not a carrier (FM band, LTE, TV stations), Gqrx currently shows different levels depending on FFT size and bandwidth (input rate/decimation). It's not possible to measure anything, except very narrow signals. For a good demo of that, look at a TV broadcast. The pilot tone stays at about the same level, but the whole rest of the plot drops 30dB between the smallest and largest FFT sizes. That gives Gqrx a bit of a believability problem.

vladisslav2011:
> Gqrx currently shows different levels depending on FFT size and bandwidth

That's correct behavior.
Less bandwidth - less noise, more bandwidth - more noise.
It is not possible to measure complex signal level with FFT if it's bandwidth is larger, than FFT bin width.
And it is not possible to emulate arbitrary RBW just by scaling FFT results.
It's almost impossible to emulate arbitrary RBW at arbitrary step with FFT without oversampling/overlapping and extra filtering.
It is possible to do "instant sweep" whit a user-chosen RBW/step with a bank of frequency-xlating FIR filters, but this will be too heavy for realtime processing... FFT may help to reduce the FIR order decimating the signal before the FIR if RBW is small, compared to input bandwidth, but it will still be a heavy task.
So if someone wants to change "RBW" in Gqrx, someone changes FFT size/input rate.
I think, the FFT bin level should match the S-meter if the receiver frequency is equal to the bin center frequency, the receiver's filter width is equal to RBW and the FFT window is set to rectangular.

PR #1218, opened 2023-04-05
Plotter features and refactoring

willcode:
- Add a min hold line to the plotter
- Use peak hold data for peak detect when available
- Data is kept in linear units (Vrms or dBm) until display
- Expensive operations moved to plotter, where they operate
  only on decimated fft data to be displayed
- New peak detect algorithm (works on raw data vs y-coordinates)
- Markers with freq, center, and span measurements
  - Checkbox or Ctrl-K toggles markers and data in plotter
  - Select a span above or below the plot using
    - Shift-click (usually)
    - Ctrl-click for max data (vs avg) on plots with both
    - Ctrl-shift-click for max hold data when enabled
  - Can be set from current red line position
  - Can be dragged (Shift-drag to move both)
- Increased max fft size (4M) and reduced scope of locks
- P key temporarily pauses and restores the waterfall
- Time on waterfall is correct and stable (unless frame rate
  can not be met in Auto mode)
- Multiple visualizations make use of decimated fft data
  - Max fft bin, same as current display
  - Avg fft bin
  - Max/avg with fill
  - Histogram
- Selectable scaling
  - dBFS - current behavior
  - dBV - interprets signal as calibrated Vrms
  - dBm - same, and assumes 50 ohm load
- For dBm, selectable denominator
  - /RBW - current behavior
  - dBV/√Hz and dBm/Hz - power spectral density
- Energy normalized FFT window for dBV/√Hz and dBm
- Averaging slider
  - Effects are same at different frame rates
  - Gamma curve makes setting feel more linear
- Waterfall can show
  - Max fft bin - current behavior
  - Avg fft bin
  - Sync to plotter (includes any IIR averaging)
- Cursor overlay text shows delta freq from red line
- Plotter performance improvents, and frame rates up to 250/s
  heavily dependent on fft size, visualization, resolution, window size ...
- Overlay is now a composited overlay (it was an underlay)
- Drop shadow on axis text, peak circles
- HiDPI enabled, all drawing uses reals (TODO: Save Waterfall)
- X/Y zoom works on last plot when stopped
- FFT controls reworked
- Frame drops indicated in fft control panel (red !! before "Rate")

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -44,6 +44,7 @@
 #include <QTextStream>
 #include <QtGlobal>
 #include <QTimer>
+#include <QFuture>
 #include <QVBoxLayout>
 #include <QSvgWidget>
 #include "qtgui/ioconfig.h"
@@ -68,6 +69,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     d_lnb_lo(0),
     d_hw_freq(0),
     d_fftAvg(0.25),
+    d_fftNormalizeEnergy(false),
     d_have_audio(true),
     dec_afsk1200(nullptr)
 {
@@ -90,6 +92,17 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
 
     setWindowTitle(QString("Gqrx %1").arg(VERSION));
 
+    // Set fixed widths for labels so they don't move around when set
+    QFontMetrics metrics(font);
+    QRect markerRect = metrics.boundingRect("99,999,999.999 kHz");
+    ui->markerLabelA->setFixedWidth(markerRect.width());
+    ui->markerLabelB->setFixedWidth(markerRect.width());
+    QRect deltaFreqRect = metrics.boundingRect("Δ99,999,999.999 kHz   ⨏99,999,999.999 kHz");
+    ui->deltaFreqLabel->setFixedWidth(deltaFreqRect.width());
+    setMarkerA(MARKER_OFF);
+    setMarkerB(MARKER_OFF);
+    d_show_markers = true;
+
     /* frequency control widget */
     ui->freqCtrl->setup(0, 0, 9999e6, 1, FCTL_UNIT_NONE);
     ui->freqCtrl->setFrequency(144500000);
@@ -108,17 +121,17 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     connect(meter_timer, SIGNAL(timeout()), this, SLOT(meterTimeout()));
 
     /* FFT timer & data */
+    d_iqFftData.resize(receiver::DEFAULT_FFT_SIZE);
     iq_fft_timer = new QTimer(this);
+    iq_fft_timer->setTimerType(Qt::PreciseTimer);
     connect(iq_fft_timer, SIGNAL(timeout()), this, SLOT(iqFftTimeout()));
+    d_last_fft_ms = 0;
+    d_avg_fft_rate = 0.0;
+    d_frame_drop = false;
 
+    d_audioFftData.resize(receiver::DEFAULT_FFT_SIZE);
     audio_fft_timer = new QTimer(this);
     connect(audio_fft_timer, SIGNAL(timeout()), this, SLOT(audioFftTimeout()));
-
-    d_fftData = new std::complex<float>[MAX_FFT_SIZE];
-    d_realFftData = new float[MAX_FFT_SIZE];
-    d_iirFftData = new float[MAX_FFT_SIZE];
-    for (int i = 0; i < MAX_FFT_SIZE; i++)
-        d_iirFftData[i] = -140.0;  // dBFS
 
     /* timer for data decoders */
     dec_timer = new QTimer(this);
@@ -156,6 +169,12 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     // zero cursor (rx filter offset)
     auto *rx_offset_zero_shortcut = new QShortcut(QKeySequence(Qt::Key_Z), this);
     QObject::connect(rx_offset_zero_shortcut, &QShortcut::activated, this, &MainWindow::rxOffsetZeroShortcut);
+    // freeze plotter on P
+    auto *toggle_freeze_shortcut = new QShortcut(QKeySequence(Qt::Key_P), this);
+    QObject::connect(toggle_freeze_shortcut, &QShortcut::activated, this, &MainWindow::toggleFreezeShortcut);
+    // toggle markers on/off
+    auto *toggle_markers_shortcut = new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_K), this);
+    QObject::connect(toggle_markers_shortcut, &QShortcut::activated, this, &MainWindow::toggleMarkers);
 
     setCorner(Qt::TopLeftCorner, Qt::LeftDockWidgetArea);
     setCorner(Qt::TopRightCorner, Qt::RightDockWidgetArea);
@@ -248,35 +267,46 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     connect(uiDockAudio, SIGNAL(audioPlayStarted(QString)), this, SLOT(startAudioPlayback(QString)));
     connect(uiDockAudio, SIGNAL(audioPlayStopped()), this, SLOT(stopAudioPlayback()));
     connect(uiDockAudio, SIGNAL(fftRateChanged(int)), this, SLOT(setAudioFftRate(int)));
+
+    // FFT Dock
     connect(uiDockFft, SIGNAL(fftSizeChanged(int)), this, SLOT(setIqFftSize(int)));
     connect(uiDockFft, SIGNAL(fftRateChanged(int)), this, SLOT(setIqFftRate(int)));
     connect(uiDockFft, SIGNAL(fftWindowChanged(int)), this, SLOT(setIqFftWindow(int)));
     connect(uiDockFft, SIGNAL(wfSpanChanged(quint64)), this, SLOT(setWfTimeSpan(quint64)));
     connect(uiDockFft, SIGNAL(fftSplitChanged(int)), this, SLOT(setIqFftSplit(int)));
-    connect(uiDockFft, SIGNAL(fftAvgChanged(float)), this, SLOT(setIqFftAvg(float)));
+    connect(uiDockFft, SIGNAL(fftAvgChanged(float)), ui->plotter, SLOT(setFftAvg(float)));
     connect(uiDockFft, SIGNAL(fftZoomChanged(float)), ui->plotter, SLOT(zoomOnXAxis(float)));
+    connect(uiDockFft, SIGNAL(waterfallModeChanged(int)), ui->plotter, SLOT(setWaterfallMode(int)));
+    connect(uiDockFft, SIGNAL(plotModeChanged(int)), ui->plotter, SLOT(setPlotMode(int)));
+    connect(uiDockFft, SIGNAL(plotScaleChanged(int, bool)), ui->plotter, SLOT(setPlotScale(int, bool)));
+    connect(uiDockFft, SIGNAL(plotScaleChanged(int, bool)), this, SLOT(plotScaleChanged(int, bool)));
     connect(uiDockFft, SIGNAL(resetFftZoom()), ui->plotter, SLOT(resetHorizontalZoom()));
     connect(uiDockFft, SIGNAL(gotoFftCenter()), ui->plotter, SLOT(moveToCenterFreq()));
     connect(uiDockFft, SIGNAL(gotoDemodFreq()), ui->plotter, SLOT(moveToDemodFreq()));
-    connect(uiDockFft, SIGNAL(bandPlanChanged(bool)), ui->plotter, SLOT(toggleBandPlan(bool)));
+    connect(uiDockFft, SIGNAL(bandPlanChanged(bool)), ui->plotter, SLOT(enableBandPlan(bool)));
+    connect(uiDockFft, SIGNAL(markersChanged(bool)), ui->plotter, SLOT(enableMarkers(bool)));
+    connect(uiDockFft, SIGNAL(markersChanged(bool)), this, SLOT(enableMarkers(bool)));
     connect(uiDockFft, SIGNAL(wfColormapChanged(const QString)), ui->plotter, SLOT(setWfColormap(const QString)));
     connect(uiDockFft, SIGNAL(wfColormapChanged(const QString)), uiDockAudio, SLOT(setWfColormap(const QString)));
-
     connect(uiDockFft, SIGNAL(pandapterRangeChanged(float,float)),
             ui->plotter, SLOT(setPandapterRange(float,float)));
     connect(uiDockFft, SIGNAL(waterfallRangeChanged(float,float)),
             ui->plotter, SLOT(setWaterfallRange(float,float)));
+    connect(uiDockFft, SIGNAL(fftColorChanged(QColor)), this, SLOT(setFftColor(QColor)));
+    connect(uiDockFft, SIGNAL(fftFillToggled(bool)), this, SLOT(enableFftFill(bool)));
+    connect(uiDockFft, SIGNAL(fftMaxHoldToggled(bool)), ui->plotter, SLOT(enableMaxHold(bool)));
+    connect(uiDockFft, SIGNAL(fftMinHoldToggled(bool)), ui->plotter, SLOT(enableMinHold(bool)));
+    connect(uiDockFft, SIGNAL(peakDetectToggled(bool)), ui->plotter, SLOT(enablePeakDetect(bool)));
+    connect(uiDockRDS, SIGNAL(rdsDecoderToggled(bool)), this, SLOT(setRdsDecoder(bool)));
+
+    // Plotter
     connect(ui->plotter, SIGNAL(pandapterRangeChanged(float,float)),
             uiDockFft, SLOT(setPandapterRange(float,float)));
     connect(ui->plotter, SIGNAL(newZoomLevel(float)),
             uiDockFft, SLOT(setZoomLevel(float)));
     connect(ui->plotter, SIGNAL(newSize()), this, SLOT(setWfSize()));
-
-    connect(uiDockFft, SIGNAL(fftColorChanged(QColor)), this, SLOT(setFftColor(QColor)));
-    connect(uiDockFft, SIGNAL(fftFillToggled(bool)), this, SLOT(setFftFill(bool)));
-    connect(uiDockFft, SIGNAL(fftPeakHoldToggled(bool)), this, SLOT(setFftPeakHold(bool)));
-    connect(uiDockFft, SIGNAL(peakDetectionToggled(bool)), this, SLOT(setPeakDetection(bool)));
-    connect(uiDockRDS, SIGNAL(rdsDecoderToggled(bool)), this, SLOT(setRdsDecoder(bool)));
+    connect(ui->plotter, SIGNAL(markerSelectA(qint64)), this, SLOT(setMarkerA(qint64)));
+    connect(ui->plotter, SIGNAL(markerSelectB(qint64)), this, SLOT(setMarkerB(qint64)));
 
     // Bookmarks
     connect(uiDockBookmarks, SIGNAL(newBookmarkActivated(qint64, QString, int)), this, SLOT(onBookmarkActivated(qint64, QString, int)));
@@ -411,9 +441,6 @@ MainWindow::~MainWindow()
     delete uiDockRDS;
     delete rx;
     delete remote;
-    delete [] d_fftData;
-    delete [] d_realFftData;
-    delete [] d_iirFftData;
     delete qsvg_dummy;
 }
 
@@ -872,6 +899,74 @@ void MainWindow::setNewFrequency(qint64 rx_freq)
     uiDockRxOpt->setHwFreq(d_hw_freq);
     ui->freqCtrl->setFrequency(rx_freq);
     uiDockBookmarks->setNewFrequency(rx_freq);
+}
+
+// Update delta and center (of marker span) when markers are updated
+void MainWindow::updateDeltaAndCenter()
+{
+    if (d_marker_a != MARKER_OFF && d_marker_b != MARKER_OFF)
+    {
+        qint64 delta = d_marker_b - d_marker_a;
+        qint64 center = (d_marker_a + d_marker_b) / 2;
+        ui->deltaFreqLabel->setText(QString("Δ%1 kHz   ⨏%2 kHz")
+                            .arg(locale().toString(delta/1.e3, 'f', 3))
+                            .arg(locale().toString(center/1.e3, 'f', 3)));
+    }
+    else {
+        ui->deltaFreqLabel->setText("");
+    }
+}
+
+// Set marker via slots
+void MainWindow::setMarkerA(qint64 freq)
+{
+    d_marker_a = freq;
+    if (freq != MARKER_OFF)
+    {
+        ui->markerLabelA->setText(QString("%1 kHz").arg(locale().toString(freq/1.e3, 'f', 3)));
+    }
+    else {
+        ui->markerLabelA->setText("");
+    }
+    ui->plotter->setMarkers(d_marker_a, d_marker_b);
+    updateDeltaAndCenter();
+}
+
+void MainWindow::setMarkerB(qint64 freq)
+{
+    d_marker_b = freq;
+    if (freq != MARKER_OFF)
+    {
+        ui->markerLabelB->setText(QString("%1 kHz").arg(locale().toString(freq/1.e3, 'f', 3)));
+    }
+    else {
+        ui->markerLabelB->setText("");
+    }
+    ui->plotter->setMarkers(d_marker_a, d_marker_b);
+    updateDeltaAndCenter();
+}
+
+// Set marker via buttons
+void MainWindow::on_setMarkerButtonA_clicked()
+{
+    auto center_freq = d_hw_freq + (qint64)rx->get_filter_offset();
+    setMarkerA(center_freq + d_lnb_lo);
+}
+
+void MainWindow::on_setMarkerButtonB_clicked()
+{
+    auto center_freq = d_hw_freq + (qint64)rx->get_filter_offset();
+    setMarkerB(center_freq + d_lnb_lo);
+}
+
+void MainWindow::on_clearMarkerButtonA_clicked()
+{
+    setMarkerA(MARKER_OFF);
+}
+
+void MainWindow::on_clearMarkerButtonB_clicked()
+{
+    setMarkerB(MARKER_OFF);
 }
 
 /**
@@ -1363,17 +1458,10 @@ void MainWindow::meterTimeout()
     remote->setSignalLevel(level);
 }
 
-#define LOG2_10 3.321928094887362
-
 /** Baseband FFT plot timeout. */
 void MainWindow::iqFftTimeout()
 {
-    unsigned int    fftsize;
-    unsigned int    i;
-    float           pwr_scale;
-
-    // FIXME: fftsize is a reference
-    rx->get_iq_fft_data(d_fftData, fftsize);
+    const unsigned int fftsize = rx->iq_fft_size();
 
     if (fftsize == 0)
     {
@@ -1381,39 +1469,38 @@ void MainWindow::iqFftTimeout()
         return;
     }
 
-    // NB: without cast to float the multiplication will overflow at 64k
-    // and pwr_scale will be inf
-    pwr_scale = 1.0 / ((float)fftsize * (float)fftsize);
+    // Track the frame rate and warn if not keeping up. Since the interval is ms, the timer can
+    // not be set exactly to all rates.
+    const quint64 now_ms = QDateTime::currentMSecsSinceEpoch();
+    const float expected_rate = 1000.0 / (float)iq_fft_timer->interval();
+    const float last_fft_rate = 1000.0 / (float)(now_ms - d_last_fft_ms);
+    const float alpha = std::pow(expected_rate, -0.75);
+    if (d_avg_fft_rate == 0.0)
+        d_avg_fft_rate = expected_rate;
+    else
+        d_avg_fft_rate = (1.0 - alpha) * d_avg_fft_rate + alpha * last_fft_rate;
 
-    /* Normalize, calculate power and shift the FFT */
-    volk_32fc_magnitude_squared_32f(d_realFftData, d_fftData + (fftsize/2), fftsize/2);
-    volk_32fc_magnitude_squared_32f(d_realFftData + (fftsize/2), d_fftData, fftsize/2);
-    volk_32f_s32f_multiply_32f(d_realFftData, d_realFftData, pwr_scale, fftsize);
-    volk_32f_log2_32f(d_realFftData, d_realFftData, fftsize);
-    volk_32f_s32f_multiply_32f(d_realFftData, d_realFftData, 10 / LOG2_10, fftsize);
-
-    for (i = 0; i < fftsize; i++)
-    {
-        /* FFT averaging */
-        d_iirFftData[i] += d_fftAvg * (d_realFftData[i] - d_iirFftData[i]);
+    const bool drop = d_avg_fft_rate < expected_rate * 0.95;
+    if (drop != d_frame_drop) {
+        if (drop) {
+            uiDockFft->setActualFrameRate(d_avg_fft_rate, true);
+        }
+        else {
+            uiDockFft->setActualFrameRate(d_avg_fft_rate, false);
+        }
+        d_frame_drop = drop;
     }
+    d_last_fft_ms = now_ms;
 
-    ui->plotter->setNewFftData(d_iirFftData, d_realFftData, fftsize);
+    rx->get_iq_fft_data(d_iqFftData.data());
+
+    ui->plotter->setNewFftData(d_iqFftData.data(), fftsize);
 }
 
 /** Audio FFT plot timeout. */
 void MainWindow::audioFftTimeout()
 {
-    unsigned int    fftsize;
-    unsigned int    i;
-    float           pwr;
-    float           pwr_scale;
-    std::complex<float> pt;             /* a single FFT point used in calculations */
-
-    if (!d_have_audio || !uiDockAudio->isVisible())
-        return;
-
-    rx->get_audio_fft_data(d_fftData, fftsize);
+    const unsigned int fftsize = rx->audio_fft_size();
 
     if (fftsize == 0)
     {
@@ -1422,28 +1509,12 @@ void MainWindow::audioFftTimeout()
         return;
     }
 
-    pwr_scale = 1.0 / (fftsize * fftsize);
+    if (!d_have_audio || !uiDockAudio->isVisible())
+        return;
 
-    /** FIXME: move post processing to rx_fft_f **/
-    /* Normalize, calculate power and shift the FFT */
-    for (i = 0; i < fftsize; i++)
-    {
-        /* normalize and shift */
-        if (i < fftsize/2)
-        {
-            pt = d_fftData[fftsize/2+i];
-        }
-        else
-        {
-            pt = d_fftData[i-fftsize/2];
-        }
+    rx->get_audio_fft_data(d_audioFftData.data());
 
-        /* calculate power in dBFS */
-        pwr = pwr_scale * (pt.imag() * pt.imag() + pt.real() * pt.real());
-        d_realFftData[i] = 10.0 * log10f(pwr + 1.0e-20);
-    }
-
-    uiDockAudio->setNewFftData(d_realFftData, fftsize);
+    uiDockAudio->setNewFftData(d_audioFftData.data(), fftsize);
 }
 
 /** RDS message display timeout. */
@@ -1703,15 +1774,17 @@ void MainWindow::seekIqFile(qint64 seek_pos)
 void MainWindow::setIqFftSize(int size)
 {
     qDebug() << "Changing baseband FFT size to" << size;
+    d_iqFftData.resize(size);
+    d_iqFftData.shrink_to_fit();
     rx->set_iq_fft_size(size);
-    for (int i = 0; i < size; i++)
-        d_iirFftData[i] = -140.0;  // dBFS
 }
 
 /** Baseband FFT rate has changed. */
 void MainWindow::setIqFftRate(int fps)
 {
     int interval;
+
+    d_fps = fps;
 
     if (fps == 0)
     {
@@ -1727,15 +1800,34 @@ void MainWindow::setIqFftRate(int fps)
             ui->plotter->setRunningState(true);
     }
 
-    if (interval > 9 && iq_fft_timer->isActive())
+    // Limit to 250 fps
+    if (interval > 3 && iq_fft_timer->isActive())
         iq_fft_timer->setInterval(interval);
 
     uiDockFft->setWfResolution(ui->plotter->getWfTimeRes());
+
+    // Invalidate average frame rate
+    d_avg_fft_rate = 0.0;
 }
 
 void MainWindow::setIqFftWindow(int type)
 {
-    rx->set_iq_fft_window(type);
+    d_fftWindowType = type;
+    rx->set_iq_fft_window(d_fftWindowType, d_fftNormalizeEnergy);
+}
+
+void MainWindow::plotScaleChanged(int type, bool perHz)
+{
+    // PLOT_SCALE_DBFS (0) always uses amplitude normalization.
+
+    // PLOT_SCALE_DBV (1) requires energy normalization for /sqrt(Hz) (1), but
+    // not for RBW (0).
+
+    // PLOT_SCALE_DBM (2) requires energy normalization of FFT window whether
+    // or not perHz is specified.
+
+    d_fftNormalizeEnergy = (type == 2) || (type == 1 && perHz);
+    rx->set_iq_fft_window(d_fftWindowType, d_fftNormalizeEnergy);
 }
 
 /** Waterfall time span has changed. */
@@ -1761,12 +1853,6 @@ void MainWindow::setIqFftSplit(int pct_wf)
         ui->plotter->setPercent2DScreen(pct_wf);
 }
 
-void MainWindow::setIqFftAvg(float avg)
-{
-    if ((avg >= 0) && (avg <= 1.0))
-        d_fftAvg = avg;
-}
-
 /** Audio FFT rate has changed. */
 void MainWindow::setAudioFftRate(int fps)
 {
@@ -1787,20 +1873,10 @@ void MainWindow::setFftColor(const QColor& color)
 }
 
 /** Enable/disable filling the aread below the FFT plot. */
-void MainWindow::setFftFill(bool enable)
+void MainWindow::enableFftFill(bool enable)
 {
-    ui->plotter->setFftFill(enable);
+    ui->plotter->enableFftFill(enable);
     uiDockAudio->setFftFill(enable);
-}
-
-void MainWindow::setFftPeakHold(bool enable)
-{
-    ui->plotter->setPeakHold(enable);
-}
-
-void MainWindow::setPeakDetection(bool enabled)
-{
-    ui->plotter->setPeakDetection(enabled ,2);
 }
 
 /**
@@ -2447,4 +2523,21 @@ void MainWindow::frequencyFocusShortcut()
 void MainWindow::rxOffsetZeroShortcut()
 {
     uiDockRxOpt->setFilterOffset(0);
+}
+
+void MainWindow::toggleFreezeShortcut()
+{
+    ui->plotter->toggleFreeze();
+}
+
+void MainWindow::enableMarkers(bool enabled)
+{
+    d_show_markers = enabled;
+    ui->markerFrame->setVisible(d_show_markers);
+}
+
+void MainWindow::toggleMarkers()
+{
+    enableMarkers(!d_show_markers);
+    uiDockFft->setMarkersEnabled(d_show_markers);
 }
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -169,9 +169,6 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     // zero cursor (rx filter offset)
     auto *rx_offset_zero_shortcut = new QShortcut(QKeySequence(Qt::Key_Z), this);
     QObject::connect(rx_offset_zero_shortcut, &QShortcut::activated, this, &MainWindow::rxOffsetZeroShortcut);
-    // freeze plotter on P
-    auto *toggle_freeze_shortcut = new QShortcut(QKeySequence(Qt::Key_P), this);
-    QObject::connect(toggle_freeze_shortcut, &QShortcut::activated, this, &MainWindow::toggleFreezeShortcut);
     // toggle markers on/off
     auto *toggle_markers_shortcut = new QShortcut(QKeySequence(Qt::CTRL | Qt::Key_K), this);
     QObject::connect(toggle_markers_shortcut, &QShortcut::activated, this, &MainWindow::toggleMarkers);
@@ -2523,11 +2520,6 @@ void MainWindow::frequencyFocusShortcut()
 void MainWindow::rxOffsetZeroShortcut()
 {
     uiDockRxOpt->setFilterOffset(0);
-}
-
-void MainWindow::toggleFreezeShortcut()
-{
-    ui->plotter->toggleFreeze();
 }
 
 void MainWindow::enableMarkers(bool enabled)
--- a/src/qtgui/dockfft.ui
+++ b/src/qtgui/dockfft.ui
@@ -6,43 +6,51 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>368</width>
-    <height>444</height>
+    <width>444</width>
+    <height>644</height>
    </rect>
-  </property>
-  <property name="sizePolicy">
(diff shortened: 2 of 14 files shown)


argilo:
Not that it matters much, but I noticed a small side effect of the changes here:

Screenshot from 2023-04-17 12-39-44

The waterfall now paints white while an SDR device is starting up for the first time; it used to paint black.

argilo:
Valgrind complains about use of uninitialized variables in various places during startup:
```
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x200E7F: rx_fft_c::set_window_type(int) (rx_fft.cpp:225)
==3307788==    by 0x2015AC: rx_fft_c::rx_fft_c(unsigned int, double, int) (rx_fft.cpp:71)
==3307788==    by 0x20193E: make_rx_fft_c(unsigned int, double, int) (rx_fft.cpp:35)
==3307788==    by 0x1D9D85: receiver::receiver(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned int) (receiver.cpp:118)
==3307788==    by 0x1C6F89: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:111)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x27DE5C: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1097)
==3307788==    by 0x232896: setPercent2DScreen (plotter.h:192)
==3307788==    by 0x232896: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:66)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x27DE5C: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1097)
==3307788==    by 0x232896: setPercent2DScreen (plotter.h:192)
==3307788==    by 0x232896: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:66)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27B6EA: CPlotter::draw(bool) (plotter.cpp:1477)
==3307788==    by 0x27DE66: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1098)
==3307788==    by 0x232896: setPercent2DScreen (plotter.h:192)
==3307788==    by 0x232896: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:66)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x27DE66: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1098)
==3307788==    by 0x232896: setPercent2DScreen (plotter.h:192)
==3307788==    by 0x232896: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:66)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x27DE66: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1098)
==3307788==    by 0x232896: setPercent2DScreen (plotter.h:192)
==3307788==    by 0x232896: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:66)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27B6EA: CPlotter::draw(bool) (plotter.cpp:1477)
==3307788==    by 0x232903: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:72)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x232903: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:72)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x232903: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:72)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27B6EA: CPlotter::draw(bool) (plotter.cpp:1477)
==3307788==    by 0x23291C: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:73)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x23291C: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:73)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x27BBC0: CPlotter::draw(bool) (plotter.cpp:1694)
==3307788==    by 0x23291C: DockAudio::DockAudio(QWidget*) (dockaudio.cpp:73)
==3307788==    by 0x1C721A: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:144)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x200E7F: rx_fft_c::set_window_type(int) (rx_fft.cpp:225)
==3307788==    by 0x201A8F: rx_fft_c::set_params() (rx_fft.cpp:169)
==3307788==    by 0x1D0067: receiver::set_input_rate(double) (receiver.cpp:364)
==3307788==    by 0x1D90C1: receiver::set_input_device(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (receiver.cpp:235)
==3307788==    by 0x1C131D: MainWindow::loadConfig(QString const&, bool, bool) (mainwindow.cpp:535)
==3307788==    by 0x1C9119: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:353)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x200E7F: rx_fft_c::set_window_type(int) (rx_fft.cpp:225)
==3307788==    by 0x201A8F: rx_fft_c::set_params() (rx_fft.cpp:169)
==3307788==    by 0x1D0067: receiver::set_input_rate(double) (receiver.cpp:364)
==3307788==    by 0x1C2490: MainWindow::loadConfig(QString const&, bool, bool) (mainwindow.cpp:585)
==3307788==    by 0x1C9119: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:353)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x22FAFB: setSpanFreq (plotter.h:91)
==3307788==    by 0x22FAFB: DockAudio::setFftRange(unsigned long long, unsigned long long) (dockaudio.cpp:102)
==3307788==    by 0x1BEF10: MainWindow::selectDemod(int) (mainwindow.cpp:1199)
==3307788==    by 0x5E0C272: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x1AD7F4: DockRxOpt::demodSelected(int) (moc_dockrxopt.cpp:507)
==3307788==    by 0x25082B: DockRxOpt::readSettings(QSettings*) (dockrxopt.cpp:471)
==3307788==    by 0x1C1A3E: MainWindow::loadConfig(QString const&, bool, bool) (mainwindow.cpp:652)
==3307788==    by 0x1C9119: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:353)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x22FAFB: setSpanFreq (plotter.h:91)
==3307788==    by 0x22FAFB: DockAudio::setFftRange(unsigned long long, unsigned long long) (dockaudio.cpp:102)
==3307788==    by 0x1BEF10: MainWindow::selectDemod(int) (mainwindow.cpp:1199)
==3307788==    by 0x5E0C272: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x1AD7F4: DockRxOpt::demodSelected(int) (moc_dockrxopt.cpp:507)
==3307788==    by 0x25082B: DockRxOpt::readSettings(QSettings*) (dockrxopt.cpp:471)
==3307788==    by 0x1C1A3E: MainWindow::loadConfig(QString const&, bool, bool) (mainwindow.cpp:652)
==3307788==    by 0x1C9119: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:353)
==3307788==    by 0x1A69D1: main (main.cpp:156)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x200E7F: rx_fft_c::set_window_type(int) (rx_fft.cpp:225)
==3307788==    by 0x201A8F: rx_fft_c::set_params() (rx_fft.cpp:169)
==3307788==    by 0x5E0C272: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x1AD491: DockFft::fftSizeChanged(int) (moc_dockfft.cpp:578)
==3307788==    by 0x2392D4: DockFft::on_fftSizeComboBox_currentIndexChanged(int) (dockfft.cpp:477)
==3307788==    by 0x1B4D82: DockFft::qt_metacall(QMetaObject::Call, int, void**) (moc_dockfft.cpp:564)
==3307788==    by 0x5E0BF56: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x502F7D4: QComboBox::currentIndexChanged(int) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5031DC9: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5034F6C: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x50351FD: QComboBox::setCurrentIndex(int) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x2394A5: DockFft::setFftSize(int) (dockfft.cpp:129)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x200E7F: rx_fft_c::set_window_type(int) (rx_fft.cpp:225)
==3307788==    by 0x5E0C272: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x1AD554: DockFft::fftWindowChanged(int) (moc_dockfft.cpp:592)
==3307788==    by 0x1B4D82: DockFft::qt_metacall(QMetaObject::Call, int, void**) (moc_dockfft.cpp:564)
==3307788==    by 0x5E0BF56: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x502F7D4: QComboBox::currentIndexChanged(int) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5031DC9: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5034F6C: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x50351FD: QComboBox::setCurrentIndex(int) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x23A792: DockFft::readSettings(QSettings*) (dockfft.cpp:346)
==3307788==    by 0x1C1A66: MainWindow::loadConfig(QString const&, bool, bool) (mainwindow.cpp:653)
==3307788==    by 0x1C9119: MainWindow::MainWindow(QString const&, bool, QWidget*) (mainwindow.cpp:353)
==3307788== 
```

argilo:
```
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27988A: CPlotter::drawOverlay() (plotter.cpp:1975)
==3307788==    by 0x27DE5C: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1097)
==3307788==    by 0x4F46B2D: QWidget::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4FE2405: QFrame::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x4F3DEA5: QWidgetPrivate::sendPendingMoveAndResizeEvents(bool, bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AD6: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F45CAA: QWidgetPrivate::setVisible(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A37: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AF2: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A29: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27989B: CPlotter::drawOverlay() (plotter.cpp:1992)
==3307788==    by 0x27DE5C: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1097)
==3307788==    by 0x4F46B2D: QWidget::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4FE2405: QFrame::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x4F3DEA5: QWidgetPrivate::sendPendingMoveAndResizeEvents(bool, bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AD6: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F45CAA: QWidgetPrivate::setVisible(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A37: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AF2: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A29: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x27B6EA: CPlotter::draw(bool) (plotter.cpp:1477)
==3307788==    by 0x27DE66: CPlotter::resizeEvent(QResizeEvent*) (plotter.cpp:1098)
==3307788==    by 0x4F46B2D: QWidget::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4FE2405: QFrame::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x4F3DEA5: QWidgetPrivate::sendPendingMoveAndResizeEvents(bool, bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AD6: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F45CAA: QWidgetPrivate::setVisible(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A37: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42AF2: QWidgetPrivate::show_helper() (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F42A29: QWidgetPrivate::showChildren(bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x274833: CPlotter::msecFromY(int) (plotter.cpp:2214)
==3307788==    by 0x27E279: CPlotter::mouseMoveEvent(QMouseEvent*) (plotter.cpp:328)
==3307788==    by 0x4F460E7: QWidget::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4FE2405: QFrame::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EF62D6: QApplication::notify(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x4EF4DE1: QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F5880A: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F5B634: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788== 
==3307788== Conditional jump or move depends on uninitialised value(s)
==3307788==    at 0x274866: CPlotter::msecFromY(int) (plotter.cpp:2214)
==3307788==    by 0x27E279: CPlotter::mouseMoveEvent(QMouseEvent*) (plotter.cpp:328)
==3307788==    by 0x4F460E7: QWidget::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4FE2405: QFrame::event(QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EF62D6: QApplication::notify(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788==    by 0x4EF4DE1: QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F5880A: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4F5B634: ??? (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x4EECD35: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Widgets.so.6.2.4)
==3307788==    by 0x5DB3A57: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /usr/lib/x86_64-linux-gnu/libQt6Core.so.6.2.4)
==3307788== 
```

argilo:
The "Freq Zoom" slider doesn't work when the DSP is stopped.

argilo:
Dragging the frequency axis in from the left results in a corrupted waterfall:

Screenshot from 2023-04-19 12-56-39

willcode:
Added gamma to the slider behavior and it behaves a little more believably. We could write a paper on the meaning of a running IIR of the max of decimated FFT bins :smile: 

The other two should be quick to track down. You can't imagine the number of wildly artistic waterfall pictures I've come up with while debugging.

argilo:
> Added gamma to the slider behavior and it behaves a little more believably.

Nice! I like the new version. :+1:

argilo:
Oh, you're right. I guess I was testing the improvements that you made yesterday. Let me see now how the alpha works.

argilo:
It seems like there's a bug where, if you start up Gqrx in 250-fps mode, the averaging slider isn't calibrated for 250 fps. But if you switch to another frame rate and back, it works fine.

Otherwise it looks good.

vladisslav2011:
This branch looks a lot better now.
All optional features may be disabled.
The plotter text readability is improve without wasting the screen space.
I have seen about 2x performance improvement on my more powerful machine yesterday, but I have more or less the same performance on the laptop (i7-8550U downclocked to 400MHz).
The only remaining issue is dockfft minimum size increase ( https://github.com/gqrx-sdr/gqrx/commit/525890cb7533f9a46c38ad314bf1e79620e632b2 )
Sorry for large screenshots (this little monster is close to 4k)
fft_channelizer
willcode
fft_channelizer2
willcode2
I'll retest on a desktop and, maybe, tiny laptop and post results here.

willcode:
Manually edited the width/height ... I can't get designer to leave them alone. I also shuffled a couple controls around. The order didn't make sense, so all the 2D plot controls are together, with the waterfall controls below.

willcode:
> This branch looks a lot better now.
> All optional features may be disabled.

@vladisslav2011 Thanks for testing, and I'm glad you see some performance gain. If I ever sell a product - you're going to be in charge of marketing :grin:

argilo:
I tested with `QT_SCALE_FACTOR=1.5` and `QT_SCALE_FACTOR=2` on Linux, and it seems to leave the right part of the panadapter & waterfall empty.

I pushed the branch to gqrx-sdr/gqrx to trigger DMG & AppImage builds. Once a DMG is ready I'll see how high-DPI is working on Mac.

Unrelated to your work here, but I noticed that the frequency control and S-meter look horrible with `QT_SCALE_FACTOR=1.5`. I might have made some incorrect assumptions when updating those to high-DPI.

willcode:
Oh, I had that working right and I must have changed something - will fix. Same thing on my machine now.

vladisslav2011:
Desktop (Ryzen 5950X downclocked to 2.2GHz - minimum possible)
QT_QPA_PLATFORMTHEME=qt5ct 
QT_FONT_DPI=80
willcode
fft_channelizer
willcode2
fft_chan2
willcode1M
fft_chan1M
willcode1M2
fft_chan1M2


It's slower at small FFT sizes (64k), but faster at large FFT sizes (1M).

src/qtgui/dockfft.ui
```diff
@@ -6,242 +6,320 @@
...
   <widget class="QWidget" name="dockWidgetContents">
+   <property name="sizePolicy">
+    <sizepolicy hsizetype="MinimumExpanding" vsizetype="MinimumExpanding">
+     <horstretch>0</horstretch>
+     <verstretch>0</verstretch>
+    </sizepolicy>
+   </property>
    <layout class="QVBoxLayout" name="verticalLayout_2">
     <item>
-     <widget class="QScrollArea" name="scrollArea">
```
argilo, line 47: Removing the `QScrollArea` from the FFT Settings panel increases the minimum height of the Gqrx window. Since FFT Settings is the tallest panel, it might be good to leave the `QScrollArea` in.

vladisslav2011: Agree.

argilo: Prior to this change, the minimum height was a bit under 720 pixels. I think that's a good number to shoot for, to keep Gqrx usable on small screens.

vladisslav2011:
I tried to build it for a tiny laptop and got
```
$ make -j33
[  1%] Automatic moc for target gqrx
[  1%] Built target gqrx_automoc
[  2%] Building CXX object src/CMakeFiles/gqrx.dir/qtgui/plotter.cpp.o
/home/vlad/gqrx-my/src/qtgui/plotter.cpp: In member function 'virtual void CPlotter::resizeEvent(QResizeEvent*)':
/home/vlad/gqrx-my/src/qtgui/plotter.cpp:1059:31: error: 'devicePixelRatioF' was not declared in this scope
     m_DPR = devicePixelRatioF();
                               ^
src/CMakeFiles/gqrx.dir/build.make:1383: recipe for target 'src/CMakeFiles/gqrx.dir/qtgui/plotter.cpp.o' failed
make[2]: *** [src/CMakeFiles/gqrx.dir/qtgui/plotter.cpp.o] Error 1
CMakeFiles/Makefile2:135: recipe for target 'src/CMakeFiles/gqrx.dir/all' failed
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```
After fixing this
fft_channelizer
willcode

It's a bit slower even at "Large" 128k FFT size. And dockfft does not fit into 600px screen height...

argilo:
> Unrelated to your work here, but I noticed that the frequency control and S-meter look horrible with QT_SCALE_FACTOR=1.5. I might have made some incorrect assumptions when updating those to high-DPI.

Fixed in #1224.

argilo:
> error: 'devicePixelRatioF' was not declared in this scope

Is that because you're running Ubuntu 16.04? `qPixelRatioF` has been around since Qt 5.6, so it should be safe for Gqrx to use. (Perhaps it's what I should be using in #1224 as well? It seems that `devicePixelRatio` returns an `int` in Qt 5.)

vladisslav2011:
Yes. Ubuntu 16.04. So no GNU Radio 3.8 and Qt 5.5.

argilo:
> Ubuntu 16.04. So Qt 5.5

Gqrx dropped support for those back in 2.14 (November 2020).

vladisslav2011:
But it builds with minor fixes and works well.

willcode:
Scaling should be working right now. I put the scrollbars back in and left the dock size at around 250. I guess the first time you run the program on any machine, there's a bit of layout and sizing you have to do. The dock might be scolled at one DPI but not at another. Anyway, let me know if that's how you want scrolling to work.

willcode:
#1224 looks good here.

vladisslav2011:
> Yes, you want the "F" version of that

Only in case of fractional DPR. Not a common case on linux, but a common case on Mac.

willcode:
Gnome has a weird scaling thing, where I select 1.5 for font scaling and "most" apps work right. I think they use Xrandr or something, and only fonts actually get scaled? Then, some apps (mainly Qt) need to be scaled down to work right.

I may have done extra juggling with the DPR stuff, but I really can't get it to work the "simpler" way. That might be because we're taking advantage of DPR to draw denser data, rather than just scaling up the same amount of data. I made the assumption that that's what we want. Not totally sure.

argilo:
> That might be because we're taking advantage of DPR to draw denser data, rather than just scaling up the same amount of data. I made the assumption that that's what we want. Not totally sure.

I think that's the right thing to do, and the direction I had intended to go in. (Up until now, I had only done the bare minimum to get text & icons rendering well on retina displays.)

argilo:
> I put the scrollbars back in

The scroll pane seems to have a border around it now, which is taking up some extra space.

GUI elements in the FFT Settings pane also seem to be spread quite far apart on macOS. I'm not sure why that is.

argilo:
DMG and AppImage builds are available here, for anyone who wants to test this branch without building from source: https://github.com/gqrx-sdr/gqrx/actions/runs/4749138857

argilo:
Valgrind shows tons of warnings, but it appears they all stem from `tlast_wf_drawn_ms` being uninitialized upon startup. Adding `tlast_wf_drawn_ms = 0;` to the `CPlotter` constructor got rid of them.

willcode:
I really appreciate all the testing! I'll get those fixed tomorrow. Qt widget spacing requires stylesheets for best results I think. Or layout margins specified in em. Haven't looked into that.

argilo:
Dragging A/B markers seems to be broken when the device pixel ratio is not 1.

argilo:
The Y axis can become very densely packed when DPR = 2:

Screenshot from 2023-04-19 23-28-23

The choice of spacing there should probably be based on device-independent pixels.

willcode:
Latest batch of fixes are in.

argilo:
Looking good so far.

New DMG / AppImage builds are available here: https://github.com/gqrx-sdr/gqrx/actions/runs/4754743223

argilo:
On macOS, the FFT Settings pane still has a thick border around it.

argilo:
The grey center line and red demod line are now thinner than before (1 pixel and not `m_DPR` pixels). Is that intentional? The marker lines are also thin. I'm wondering whether these line should all be set to `m_DPR` for better visibility.

willcode:
Lines now look good when I switch to HiDPI. The data lines are not scaled due to performance. If this becomes a problem, we switch the lines to "hairline" mode and see if that helps.

Click selection of marker spans is a little more obvious. It's shift-click, unless you specifically want the max on the fancier plots (where max is harder to see anyway), or ctrl-shift click to use the max hold mask.

I can't see the effects of the frame/border you see on macOS. I turned off frame shadow, which I though only had an effect when there was a frame border.

I wish you could set up to have automatic width but scrolled vertical for the fft panel. Can't find a way to do that, so it's scrolled vertical and clipped width. Having the horizontal scroll bar to access control doesn't seem useful.

willcode:
The grid is a little overbearing this way, so I'll switch it hairline and leave everything else scaled. Ugh, broke something in marker select for HiDPI now.

willcode:
OK, took another run through what does/doesn't need to be DPR'd, and fixed a marker fill bug.

I've tried out a number of HiDPI settings. My main monitor is 3840x2160.
- usual settings: font scaling set to 1.4
- font scaling set to 1.0, with and without various QT_SCALE_FACTORs
- usual settings with various QT_SCALE_FACTORs

It looks good to me, given
- I don't have a Mac, so I don't know what it looks like there
- it's also a little difficult to separate out what Xrandr is doing from the other effects

Let me know if you see anything else. I'm going to mark this ready for review, since it seems to be stable. I'll keep looking for cleanups and performance tricks (for instance, the IIR is bypassed now if averaging is set to 0) but no need to hold up (unless more problems pop up).

argilo:
Amazing! I'll kick off builds again, and take the code for a spin later tonight.

argilo:
AppImage & DMG: https://github.com/gqrx-sdr/gqrx/actions/runs/4759235336

argilo:
Unfortunately, the extra border is still present on Mac:

Screenshot 2023-04-20 at 18 30 45

There's also still a lot of empty space between UI elements in the pane.

argilo:
Removing all margins has left text butting up against the right side of the window, which looks a bit ugly to me:

Screenshot from 2023-04-20 22-51-54

The peak detect / hold check boxes also look a bit confusing in my opinion, although I'm not sure how to improve them.

argilo:
On Mac, a few words are now getting cut off:

Screen Shot 2023-04-20 at 10 55 29 PM

argilo:
Also on Mac, double margins can still be seen between table rows. Maybe because the UI elements within default to having margins on Mac but not Linux? I assume the doubling is due to each row being contained in a QHBoxLayout.

willcode:
A few more margin tweaks and some reordering/grouping.

argilo:
Hmm... it seems like that on its own is sufficient in Qt Creator, but not when you actually run it.

argilo:
OK, it does work if you let Qt Creator update the minimum width automatically to 388.

argilo:
Looking better on Mac now, but there's still a bit of text overlap in the "Show" section. (Spacing too low?)

Screen Shot 2023-04-21 at 11 34 22 AM

Another minor issue: The "Freq Zoom" slider changes its width as the `zoomLevelLabel` text changes. It's more noticeable on Mac where different digits have different widths, and the text is right justified.

willcode:
You get to choose between a scroll area and the dock defending its space. Best I can tell, there's no way to have both, except to put a minimum size on the dock. And any size will be wrong for some platform.

argilo:
Clicking "→A" or "→B" to set the marker frequencies does not work correctly when "LNB LO" (in the Input controls panel) is nonzero.

argilo:
Automated builds: https://github.com/gqrx-sdr/gqrx/actions/runs/4773087570

argilo:
With "Show Peaks" enabled, if I zoom way in (using the mouse wheel), the following segfault occurs:
```
Thread 1 "gqrx" received signal SIGSEGV, Segmentation fault.
0x00005555556ca5b2 in CPlotter::draw (this=0x7fffce77e010, newData=<optimized out>) at /home/argilo/prefix_310/src/gqrx/src/qtgui/plotter.cpp:1701
1701	                    m_peakSmoothBuf[i + xmin] = 0; // m_peakSmoothBuf[xmin + pw];
(gdb) p xmin
$4 = -2147483648
(gdb) bt
#0  0x00005555556ca5b2 in CPlotter::draw(bool) (this=0x7fffce77e010, newData=<optimized out>)
    at /home/argilo/prefix_310/src/gqrx/src/qtgui/plotter.cpp:1701
#1  0x0000555555605b5d in MainWindow::iqFftTimeout() (this=0x7fffffffd610)
    at /home/argilo/prefix_310/src/gqrx/src/applications/gqrx/mainwindow.cpp:1479
#2  0x00007ffff65ac273 in  () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#3  0x00007ffff65ba4be in QTimer::timeout(QTimer::QPrivateSignal) () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#4  0x00007ffff65a062f in QObject::event(QEvent*) () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#5  0x00007ffff737fd36 in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
    at /lib/x86_64-linux-gnu/libQt6Widgets.so.6
#6  0x00007ffff6553a58 in QCoreApplication::notifyInternal2(QObject*, QEvent*) ()
    at /lib/x86_64-linux-gnu/libQt6Core.so.6
#7  0x00007ffff66bc203 in QTimerInfoList::activateTimers() () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#8  0x00007ffff677bbfc in  () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#9  0x00007ffff41dbd3b in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#10 0x00007ffff42306c8 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007ffff41d93e3 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff677beae in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /lib/x86_64-linux-gnu/libQt6Core.so.6
#13 0x00007ffff6560adb in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /lib/x86_64-linux-gnu/libQt6Core.so.6
#14 0x00007ffff655c0f3 in QCoreApplication::exec() () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#15 0x00005555555f2c98 in main(int, char**) (argc=<optimized out>, argv=<optimized out>)
    at /home/argilo/prefix_310/src/gqrx/src/applications/gqrx/main.cpp:161
```

argilo:
Also, I noticed that when "Show Peaks" is disabled, the FFT line disappears at extreme zoom levels, while it remains visible in 2.15.10.

argilo:
> reserve space for zoom level label

With this change, four-digit values begin to overlap the slider partially, and five-digit values (which are possible using the mouse wheel) are not readable. I guess space should be reserved for five digits, plus some spacing / margin.

Personally, I like the old left-justified look (in 2.15.10) a bit better than the new right-justified.

argilo:
On Mac, text is still overlapping in the "Show" section. I guess some spacing is needed there.

Screen Shot 2023-04-22 at 11 28 19 AM

argilo:
> Personally, I like the old left-justified look (in 2.15.10) a bit better than the new right-justified.

I guess it's meant to match the right-justified RBW / Overlap / Res though, so maybe it's better for consistency to leave it the same as those.

willcode:
Either way. I figured the average user is going to try to get the panel just the right width and things will look better right justified. If the panel is stretched, it could look goofy.

Will add another zoom digit. Funny, I ran across that and said to myself nobody will ever want 10000x 😀

argilo:
> Funny, I ran across that and said to myself nobody will ever want 10000x

With the new 4M FFT, a zoom of around 10000x is necessary to see individual bins on a narrow screen. So I can imagine it might get used in practice.

willcode:
Limited zoom to keep around 4 points on the screen, and changing fft size can cause a zoom out if needed.

Added an indicator that the plotter is unable to keep up. Waterfall will also be invalid if so, but I didn't know if we wanted to mark up the waterfall when time is lost. This shouldn't happen in manual span mode anyway, and that's where people will most likely care about accurate time.

More fixes/tweaks for histogram.

I think I'll stop the features and tweaks at this point and just do fixes if needed.

willcode:
Don't really like the way the right justfied text is working out. Will do something so it doesn't move around as much.

willcode:
Text is better now. Also, setting a stylesheet at the dock level
`QComboBox { padding: 0 0.2em 0 0; }`
gets rid of the need to do this on every widget. This can be done at the app level, too. I'd be interested in how that looks on other platforms. The `em` is a measure of character width, as opposed to px which might not translate well.

Current Gqrx keeps saves zoom level in the conf file. It doesn't save the pan offset, so this is only useful when the user is  zoomed in on center. Didn't change the behavior.

argilo:
At high zoom levels, the FFT is cut off at the right hand side.

Also, when dragging the frequency axis left and right, the axis moves smoothly but the FFT shifts one bucket at a time.

Before:

Screenshot from 2023-04-30 11-49-42

After:

Screenshot from 2023-04-30 11-47-51

willcode:
I started drawing bins on center, where they were previously drawn to the right of center. Let me see if adding one more bin the right gets rid of this.

argilo:
> I started drawing bins on center, where they were previously drawn to the right of center.

Nice. That sounds like the right way to do it.

willcode:
It's hard to know where a double is needed - not sure how the compiler figures it out. `qreal` is a double, so anything related to drawing might as well use double math. Some averaging and other math need double even though the input and output only need to be float.

argilo:
The marker area shading doesn't line up with the centred bins:

Screenshot from 2023-04-30 13-03-24

Also, the marker area is cut off to the right when zooming in:

Screenshot from 2023-04-30 13-03-41

argilo:
> It's hard to know where a double is needed

If this was in reference to #1232, there's no need to worry about it now. I'll sort out what should be `float` and what should be `double` later.

argilo:
When increasing the "Rate" setting, the red warning always lights up for a period of time, regardless of whether any frames are actually lost.

argilo:
Keyboard navigation no longer works for some of the elements in the FFT Settings panel. We should probably keep that for accessibility.

willcode:
The drawing errors when dragging freq scale uncovered some other stuff too. Might be a couple days before I get a chance to sort it all out.

willcode:
Time on the waterfall is invalid after pausing, so I cleared it. Up to you.

willcode:
Worked off the latest set. Rate warning is still there but much shorter. There's an IIR on the rate monitor so jitter doesn't constantly trigger it. If the initial 1/4 second flash when raising the rate is too annoying, I can find a way to reset the IIR on each rate change to something just above the desired rate.

Dockfft did not emit this fps on startup, so the rate-compensated alpha did not get set. Fixed.

Tab focus for widgets is back, and in order. Waterfall now pauses on P instead of SPACE, since the shortcut ate the checkbox toggle via spacebar.

argilo:
A very minor nitpick (feel free to ignore if it would be difficult to fix):

The first & last displayed FFT bins (-1023 and +1023 in the case of a 2048-bin FFT) are only partially displayed (due to some rounding / truncation?).

Screenshot from 2023-05-03 22-30-55

This also got me thinking about whether there's any sense in displaying bin -1024. Perhaps half could be displayed on each end. Again, not particularly important.

argilo:
Another minor issue: at high zoom levels, the "show peaks" circles seem to display in different locations within a bin depending on the zoom level, or even between bins. And occasionally an overlapping streak of circles are seen in a single bin. (The latter behaviour seems to be always present at high zoom levels on master.)

argilo:
So far things are looking very good with the latest commits. I haven't yet encountered any bugs that would preclude merging this. I kicked off a build in case others would like to take it for a spin using the DMG or AppImage: https://github.com/gqrx-sdr/gqrx/actions/runs/4878864125

willcode:
The Nyquist bin doesn't seem like useful data so I dropped it. Technically easy to add but do you think it's correct?

I can try to figure out the edges better. Lots of edge cases (😀) when projecting one array onto another and making sure bounds are valid. I had also tried interpolation at high zoom. Didn't like the results. Log of linear interpolations made funny curves.

Multiple peak circles is on my list. Peak finding for unknown signal shapes seems to be an art. It's not obvious which is the correct peak. The 2 pass algorithm seems to work for most spectrum shape.

argilo:
> The Nyquist bin doesn't seem like useful data so I dropped it. Technically easy to add but do you think it's correct?

After thinking about it some more, I agree with leaving it out. The outermost bins are always going to be a mess of aliases & filter roll-off anyway.

willcode:
Squashed.

argilo:
The FFT Settings panel is still looking ugly on macOS.

Before:

Screen Shot 2023-05-05 at 4 53 21 PM

After:

Screen Shot 2023-05-05 at 4 53 00 PM

Text is flowing out of combo boxes, check boxes are overlapping, and buttons are tiny. Also, combo box selections get cut off:

Screen Shot 2023-05-05 at 4 52 27 PM

I'm terrible with GUIs, but I at least have a Mac to test with, so let me know if you need help.

willcode:
Looks like they need some more padding back. It will make the Linux version a little looser, but we don't want OS ifdefs. Yes, glad you have something to test on!

argilo:
> we don't want OS ifdefs

Agreed. I want Gqrx to have as few `#ifdefs` as possible, since they're a frequent source of hard-to-diagnose bugs.

For GUI elements, I think we should also minimize the number of default properties we override. I'm open to overrides that help compress the FFT Settings dialog though, since the large number of elements would otherwise make Gqrx less usable on small screens.

willcode:
See if that's any better. It's still pretty good on my screen.

argilo:
Thanks. I kicked off a build and will take it for a spin & post screenshots when it's ready.

willcode:
Debugging by carrier pidgeon :grin:

argilo:
Getting better, but text is still overflowing in some places and buttons remain small.

Screen Shot 2023-05-05 at 6 49 26 PM

Screen Shot 2023-05-05 at 6 49 50 PM

willcode:
Stylesheet is completely removed from the dock. Let's see what that does.

argilo:
The plotter behaves erratically after Gqrx is launched, but before DSP is started. It seems like it rarely redraws.

willcode:
There seems to be a repaint paint/update missing somewhere. Should be a quick fix tomorrow.

willcode:
That was it.

argilo:
> Stylesheet is completely removed from the dock. Let's see what that does.

Looks better:

Screen Shot 2023-05-05 at 10 07 30 PM

There's still some overlap with the "Show" check boxes. Also, the remaining stylesheet for `QPushButton` is still making the buttons small. That happens in the other panels too:

Screen Shot 2023-05-05 at 10 07 44 PM

argilo:
Deleting the `QPushButton` stylesheet fixes the tiny buttons.

Changing the `spacing` parameter from 0 to -1 in `horizontalLayout_12` and `horizontalLayout_13` fixes the overlapping "Show" check boxes.

argilo:
I noticed that the macOS styling woes go away if Gqrx is started using the Fusion style (`gqrx -s Fusion`), which is the default on Linux. But I suppose we should probably try to design our UI layouts to be style-agnostic.

willcode:
All custom styling removed. I don't know how to deal with macOS being different. I tried making optional external stylesheets, but the `-stylesheet` param only applies to the main window, stylesheets as resources didn't work, etc.

src/applications/gqrx/mainwindow.cpp
```diff
@@ -44,6 +44,7 @@
 #include <QTextStream>
 #include <QtGlobal>
 #include <QTimer>
+#include <QFuture>
```
argilo, line 47: Leftover?

willcode:
That was part of a concurrency experiment. AFK at the moment.

PR #1219, opened 2023-04-07
use Blackman-Harris window in ddc to reduce aliasing

dl2kcd:
At input sampling frequencies of 2 MHz and above, GQRX uses a downconverter to decimate the sampling frequency before demodulation is performed. However, the aliasing filter used has poor stopband attenuation. This results in ghost signals that can be heard clearly at frequencies where the spectrum only shows a flat baseline, typically 1 MHz above or below a strong signal.
To reproduce the problem, set the input sampling frequency to 2 MHz or a higher integer multiple of 1 MHz. Then search a strong signal where the spectrum is quiet 1 MHz above or below that signal. Compare the S-meter readings when tuning to the strong signal and when tuning to the clear frequency 1 MHz below or above the signal. Tune in such a way that the strong signal itself remains in the range of the main spectrum. At 2 MHz input sample rate, the difference of S-meter readings is only about 33 dB, and the strong signal is clearly demodulated at the frequency that is clear in the spectrum. The S-meter reading when tuned to that frequency is also much higher than at other frequencies that show the same noise level in the main spectrum.
A simple fix is to use the Blackman-Harris window instead of the default Hamming window when creating the filter taps. This drastically increases stopband attenuation at a moderate increase of the number of filter taps, without significant impact on overall CPU load.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -7,6 +7,7 @@
      FIXED: Remove empty frame from bottom of I/Q tool window.
      FIXED: Sudden scrolling of file list in I/Q tool window.
      FIXED: Reset zoom slider after right click on panadapter / waterfall.
+     FIXED: Aliasing when input rate is higher than 2 Msps.
   IMPROVED: AGC performance.
   IMPROVED: WFM stereo & OIRT performance.
   IMPROVED: Apply amplitude normalization to FFT window functions.
--- a/src/dsp/downconverter.cpp
+++ b/src/dsp/downconverter.cpp
@@ -94,7 +94,12 @@ void downconverter_cc::update_proto_taps()
     {
         double out_rate = d_samp_rate / d_decim;
         filt->set_taps(gr::filter::firdes::low_pass(1.0, d_samp_rate, LPF_CUTOFF, out_rate - 2*LPF_CUTOFF,
-            gr::fft::window::WIN_BLACKMAN_HARRIS));
+#if GNURADIO_VERSION < 0x030900
+            gr::filter::firdes::WIN_BLACKMAN_HARRIS
+#else
+            gr::fft::window::WIN_BLACKMAN_HARRIS
+#endif
+        ));
     }
 }
 
--- a/src/dsp/downconverter.cpp
+++ b/src/dsp/downconverter.cpp
@@ -93,7 +93,8 @@ void downconverter_cc::update_proto_taps()
     if (d_decim > 1)
     {
         double out_rate = d_samp_rate / d_decim;
-        filt->set_taps(gr::filter::firdes::low_pass(1.0, d_samp_rate, LPF_CUTOFF, out_rate - 2*LPF_CUTOFF));
+        filt->set_taps(gr::filter::firdes::low_pass(1.0, d_samp_rate, LPF_CUTOFF, out_rate - 2*LPF_CUTOFF,
+            gr::fft::window::WIN_BLACKMAN_HARRIS));
     }
 }
 


argilo:
I did some calculations, and you are correct that the filter in the decimator is poor, providing only 33.3 dB of stopband attenuation with 7 taps. Switching to a Blackman-harris window does improve performance significantly, but increases the number of taps to 11. I played around with GNU Radio's filter design tool, and found that an equiripple filter can achieve 112 dB of stopband attenuation with only 7 taps:

Screenshot from 2023-04-07 21-11-35

This seems like it would be a better choice, since it won't have any effect on CPU utilization. I'll take a crack at updating the code to use this approach.

The current code also doesn't compile on earlier GNU Radio versions, probably because of recent API changes.

dl2kcd:
Hi Clayton,
thanks, that looks great!
Best regards, Jo

argilo:
After looking into this some more, it seems that equiripple filters are not so easy to compute at runtime. So I had a look at windowed filters again, and found that the Kaiser window gives a decent improvement:

Screenshot from 2023-04-10 00-10-05

Perhaps we should use that.

dl2kcd:
Kaiser window with beta = 8 was actually one of the options I had considered before making the PR. That still would increase the number of taps, though. But do you have to keep that number fixed at all cost? When running GQRX on my PC and looking at the CPU load using the "top" utility, I find it hard to tell the difference. Before making a decision, we should probably quantify the impact on work load for the various options.

argilo:
> Before making a decision, we should probably quantify the impact on work load for the various options.

Agreed. Options that don't change the number of tops would obviously have no impact on performance, but the current number of taps is not necessarily the best choice so it could be changed.

> When running GQRX on my PC and looking at the CPU load using the "top" utility, I find it hard to tell the difference.

You can use `top -H` to break out individual threads, which should make it easier to measure the impact of the decimation block in isolation.

dl2kcd:
With -H and taking the %CPU value of the thread with "command" freq_xlating_..., I obtained the following values on my computer (amd64), for sampling frequencies of 2, 5, and 10 MHz:

```
Hamming:
rate   taps     DDC    total
 2 MHz    7    7.7%    81.8%
 5 MHz   15   10.5%   141.5% 
10 MHz   31   16.3%   230.0%

Kaiser(8):
rate   taps     DDC    total
 2 MHz    9    8.1%    82.5%
 5 MHz   25   12.1%   144.2% 
10 MHz   49   20.5%   240.1%

Blackman-Harris:
rate   taps     DDC    total
 2 MHz   11    8.4%    83.0%
 5 MHz   27   12.9%   146.0% 
10 MHz   55   21.7%   233.0%
```

DDC is the %CPU for the filter thread, and total is the total %CPU. Strangely, the latter is lower for 55 taps than for 49 taps at 10 MHz sampling frequency. I double checked that.

I would argue that at least in the range that affects narrowband modes, the target should be at least 80 dB of attenuation. A signal at S9+30dB level would then create an alias below S1.

dl2kcd:
Another idea: What about a lookup table for equiripple filter taps, created e.g. by a Python script, and then hardcoded into GQRX?
If you design an equiripple filter at some sampling frequency f1 such that it has 120 kHz passband width and 120 kHz distance between start of stopband and decimated frequency (as the current implementation does), then this filter will have even larger passband width and larger distance of stopband edge to decimated frequency when used at some sampling frequency f2 > f1. So the filter design at the low end of a sampling frequency interval f1 to f2 will cover the worst case in that interval. The number of taps could be chosen to be just like those of the corresponding firdes design with Hamming window at sampling frequency f1. This would have no impact on performance or even improve performance at sampling frequencies between f1 and f2 where firdes uses more taps. And as your example with the 7 tap filter shows, it would probably still result in much better anti aliasing.
Since one filter is designed for one decimation factor, the intervals may simply be the multiples of 1 MHz where the decimation factor changes.
And the table would be finite, so some limit on the sampling frequency for this solution has to be chosen. 100 MHz might be enough for everybody. As a fallback, one could still use a firdes filter in case someone hits that limit.
I have not tried this yet, and there may be some pitfall that I have overlooked or some error in my thoughts, or the equiripple filters may turn out to be not as good as expected at higher frequencies. I will look further into this.

(And, by the way, there may be other tweaks for optimization as well. It seems that for SSB and CW reception on my PC, not using a downconverter at a sampling frequency of 2 MHz uses less %CPU than the current implementation. Also, the target of 1 MHz decimated frequency or a fixed target anyways may not be optimal in terms of performance. The geometric mean between the sampling frequency and some frequency characteristic for the demodulator would a priori be plausible as an optimum. But such optimizations could be a second step.)

vladisslav2011:
1 Msps as first decimation stage target may be too much. I have changed TARGET_QUAD_RATE to 400ksps in my fork. It works well with much less aliasing and consumes a lot less CPU.
I have tested different windows and it looks, like Hann window gives ~20 better attenuation and has less taps, than default Hamming widow.
Another interesting option is Kaiser window with beta=2.26.
I have tested `pm_remez`-generated taps too - much worse results. Maybe, I'm doing something wrong and `pm_remez` should be called with different arguments...)
There is a branch, with GUI widgets to test different window types and an option to show FIR output on the plotter:
https://github.com/vladisslav2011/gqrx/tree/fir_playground
Feel free to play :-)
hamming
hann
kaiser2 26
kaiser6 7

dl2kcd:
I find the claim that Hann window with target 400 ksps performs so much better than Hanning window hard to believe. Below is a plot of the filter curves for Hamming and Hann windows at input rate 2 MHz and decimation 5. The vertical line at 0.4 MHz is where the attenuation is most critical to prevent aliasing. I have not yet looked at your code, though.
sim-hann-2-5

vladisslav2011:
Hann window is better in 3rd Nyquist zone and is about equal (slightly worse) in 2nd Nyquist zone. That's expected.
The Blackman-harris window is the best, the Blackman is good too, but both are heavy...
Kaiser with the same number of taps a Hamming is a lot better, but may be still not enough with some 16bit frontends and quiet place conditions.
I had hard times making a test IQ recording...
Test results (all basic windows at 1M and 400k target rates, Kaiser with different beta, FFT channelizer): https://www.youtube.com/watch?v=sVDAYAGSPYI

You may pick the latest commit from fir_playground branch and rebase it on top of current master, if you don't want to review all that crazy stuff, I have written.

argilo:
I added another commit which fixes compilation with GNU Radio 3.8 and earlier.

After playing a bit more with equiripple and windowed filters, it does seem that the number of taps will need to be increased to achieve good anti-aliasing performance. So I think we can accept the additional CPU cost. Blackman-harris does seem to perform best among the windowed filters.

> What about a lookup table for equiripple filter taps

I thought about this as well. I'd only want to do this as a last resort if there's no better way to dynamically generate taps. I'd expect it's possible using either GNU Radio's `pm_remez` or another implementation like https://github.com/sfilip/firpm.

argilo:
> I have tried pm_remez - it does not generate better filters :-(

GNU Radio's Filter Design Tool uses `pm_remez` under the covers, and I was able to generate some good filters with it (like the 7-tap filter in https://github.com/gqrx-sdr/gqrx/pull/1219#issuecomment-1500751640. But I haven't tested it over a wide range of conditions.

argilo:
I rebased to pick up facfcc04759847971be9d9c32267a78bac528b42, which hopefully will get CI passing on macOS as well.

vladisslav2011:
I should take a look at GNU Radio's Filter Design Tool sources and learn, how to use `pm_remez` correctly...

argilo:
> I should take a look at GNU Radio's Filter Design Tool sources and learn, how to use `pm_remez` correctly...

Helper functions that convert filter specifications into Parks-McClellan parameters are here:

https://github.com/gnuradio/gnuradio/blob/main/gr-filter/python/filter/optfir.py

vladisslav2011:
Thanks, but I have already found optfir.py, extracted deviations calculation, made Gqrx call `pm_remez` with exactly same arguments, as gr_filter_design does and it works. 7 taps and ~90 dB attenuation at 2M=>1M decimation. Works well at higher sample rates too. The only issue is that I have not used the 'number of taps' calculation from optfir. I just used number of taps from `gr::filter::firdes::low_pass_2` with 65dB attenuation and Hamming window...

argilo:
I'm happy with the switch to Blackman-harris, so I'll merge this and include it with the upcoming 2.15.10 release.

If it's possible to reduce CPU utilization while retaining the same anti-aliasing performance (for instance, by using equiripple filters or changing the `TARGET_QUAD_RATE`) we can consider that for a future release.

PR #1220, opened 2023-04-14
Drop support for GNU Radio 3.7

argilo:
GNU Radio 3.7 had its final release in February 2020, and has been fully unsupported since November 2021. Ubuntu 18.04, which ships with GNU Radio 3.7, goes out of support at the end of this month. So as we begin work on the upcoming Gqrx 2.16 feature release, I think it's time to say goodbye to GNU Radio 3.7 and remove all the conditionally-compiled code that was needed to support it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ jobs:
     strategy:
       matrix:
         image:
-          - "ubuntu:18.04"
           - "ubuntu:20.04"
           - "ubuntu:22.04"
         backend:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,10 +148,6 @@ math(EXPR GNURADIO_BCD_VERSION
 )
 add_definitions(-DGNURADIO_VERSION=${GNURADIO_BCD_VERSION})
 
-if(Gnuradio_VERSION VERSION_LESS "3.8")
-    find_package(Boost COMPONENTS system REQUIRED)
-endif()
-
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD")
     if(NOT LINUX_AUDIO_BACKEND)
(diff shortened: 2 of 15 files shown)


willcode (approved):

vladisslav2011:
OK. I'll keep 3.7 support in my fork.

argilo:
You're certainly welcome to continue maintaining 3.7 support in a fork. :+1: 

My motivations for removing support from the master branch were to reduce the testing & support burden, and to make it easier for contributors since pull requests will only need to support three GNU Radio versions instead of four.

Looking ahead to the future, my inclination is to keep 3.8 and 3.9 support in place for another couple years until Ubuntu 20.04 goes out of support.

argilo:
> my inclination is to keep 3.8 and 3.9 support in place for another couple years until Ubuntu 20.04 goes out of support

On second thought, it might become necessary to drop GNU Radio 3.8 support earlier if we switch from gr-osmosdr to gr-soapy (which requires GNU Radio >= 3.9.2). I don't have immediate plans to switch to gr-soapy, but perhaps in Gqrx 2.17.

vladisslav2011:
> which requires GNU Radio >= 3.9.2

https://github.com/dicta/gr-soapy/

GNU Radio 3.7.7+

It will be another commit to revert to me as I personally do not like SoapySDR. If I'll have to choose between SoapySDR runtime "Dll hell" and gr-osmosdr compile time "dependency hell", I'll prefer compile time - gr-osmosdr without SoapySDR support.

argilo:
I suppose it may be possible to support the external gr-soapy, but only if its API is the same as the built-in gr-soapy in GNU Radio >= 3.9.2. (I'm not sure whether that's the case.)

vladisslav2011:
On the other hand, bumping minimum required GNU Radio version to 3.9.2 does not look problematic to me as I can't use upstream Gqrx on my computers (including a tiny laptop running Ubuntu 16.04 32-bit) since this PR. :-)

PR #1224, opened 2023-04-19
Fix high-DPI rendering of frequency control and S-meter

argilo:
The frequency control and S-meter look pixelated on high-DPI screens where the device pixel ratio is fractional. This can be observed by running, for instance, `QT_SCALE_FACTOR=1.5 gqrx`:

Screenshot from 2023-04-19 18-15-20

After storing the device pixel ratio in a `qreal` instead of `int`, all is well:

Screenshot from 2023-04-19 18-16-10


argilo:
To support Qt 5 it was also necessary to replace `devicePixelRatio` with `devicePixelRatioF`, which I've now done.

argilo:
Tested with Qt 5 and all is well.

PR #1225, opened 2023-04-21
Don't explicitly set button sizes

argilo:
Three of the docks (Audio, Receiver Options, FFT Settings) explicitly increase minimum button sizes, then selectively decrease them on Linux with conditionally-compiled code. This complicates development. I think it would be best to leave full responsibility for determining appropriate button sizes to Qt.

The FFT Settings panel is already addressed as part of #1218, so I'm only touching the other two panels here.

### DIFF
--- a/src/qtgui/dockaudio.cpp
+++ b/src/qtgui/dockaudio.cpp
@@ -38,15 +38,6 @@ DockAudio::DockAudio(QWidget *parent) :
 {
     ui->setupUi(this);
 
-#ifdef Q_OS_LINUX
-    // buttons can be smaller than 50x32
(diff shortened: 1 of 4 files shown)


argilo:
The buttons look the same to me before & after this change on both Linux and macOS. So this would seem to be a safe change.

willcode (approved):
This fixes the squashed buttons on those panels. Nice. We could also tighten those panels up vertically later on to match the fft dock.

argilo:
It looks like this had the side effect of making the minimum width of the Audio panel larger, which forces other panels to be wider as well. I'll see what can be done about that.

argilo:
Proposed fix in #1226.

PR #1226, opened 2023-04-22
Allow dock buttons to be smaller

argilo:
In #1225 I removed code that explicitly changed the minimum size of buttons in the "Audio" and "Receiver Options" panels. Because Qt has a fairly wide default minimum width for buttons, this made the panels noticeably wider, reducing usability on small screens.

Here I've added back a minimum width (of 40, a bit narrower than before). I left out the Linux-specific tweaks, which I hope are unnecessary.

Here's how the panels now look (on Ubuntu 22.04) at their minimum widths:

Screenshot from 2023-04-22 12-23-54

Screenshot from 2023-04-22 12-23-59

### DIFF
--- a/src/qtgui/dockaudio.ui
+++ b/src/qtgui/dockaudio.ui
@@ -163,6 +163,12 @@
           <verstretch>0</verstretch>
          </sizepolicy>
         </property>
+        <property name="minimumSize">
+         <size>
+          <width>40</width>
+          <height>0</height>
+         </size>
+        </property>
         <property name="toolTip">
          <string>Mute audio output</string>
         </property>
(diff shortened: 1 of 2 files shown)


argilo:
Minimum widths on macOS:

Screen Shot 2023-04-22 at 1 34 03 PM

Screen Shot 2023-04-22 at 1 34 00 PM

vladisslav2011 (approved):

willcode:
I think the secret is to add a stylesheet to buttons `padding: 0px 10px 0px 0px`  (top, right, bottom, left). Using that, and some hlayout stretch policies, the dockfft is pretty small now. I wonder if we can remove the scroll.

PR #1236, opened 2023-05-07
mainwindow: round instead of int cast for freq and rate

willcode:
Frequencies could show up wrong due to cast/precision. This showed up in file names recorded by the IQ tool.

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1098,8 +1098,8 @@ void MainWindow::setIgnoreLimits(bool ignore_limits)
 {
     updateHWFrequencyRange(ignore_limits);
 
-    auto filter_offset = (qint64)rx->get_filter_offset();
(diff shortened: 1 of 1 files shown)


argilo:
This seems reasonable, but I'd like to understand what's going on here. Doubles can exactly represent all integer values up to 2^53, so I wonder where the imprecision is creeping in.

argilo:
Looks like it's gr-osmosdr; `get_center_freq` doesn't necessarily return the same frequency that was set in `set_center_freq`.

argilo:
And gr-osmosdr just gets the value from the underlying hardware driver. For USRP, I expect imprecision comes from this calculation: https://github.com/EttusResearch/uhd/blob/8a6709588b2dd8fee5bcd33a851406e64ef6d609/host/lib/usrp/multi_usrp.cpp#L269

This convinces me that the imprecision is not due to a bug in Gqrx, so I think rounding is the right approach.

PR #1171, opened 2022-11-10
IQ tool: disable unused controls when recording/playback is in progress.

vladisslav2011:
Changing IO devices and loading/saving settings does not look like good thing to do while recording/playing an IQ file.
Disable seek slider during recording as it has no function in this state.
Disable file list and directory selector during recording/playback.

Demonstration: https://www.youtube.com/watch?v=fo0e3f7QuUQ

PR #1237, opened 2023-05-07
IQ Tool: QFileSystemWatcher, tolerate directory changes

willcode:
- Use QFileSystemWatcher instead of polling
- Sort by time instead of name
- Rec time is not dependent on dir contents
- Gray out more widgets while recording

This is similar to #1164 and part of #1171 together, with some added tolerance for files changing during recording or being named as if they were recorded in the future.


vladisslav2011:
I suggest adding a context menu to the file list to make file sorting order/property configurable by the user.
Sometimes it may be better to sort by creation/modification time, but as file name contains it's recording time, sorting by name effectively sorts by creation time.

willcode:
Good idea. There's still a problem to deal with, which is why I made it by creation time. The tool specifically looks at the last line in the list to determine which file is being recorded. It does this because rx creates the file and the tool doesn't know the name otherwise. Since the tool looks at an "*.raw" file (we could make it more specific) it is possibly for other files to appear last in the list. It's also possibly to get a file from a computer that has a different clock setting, and has what looks like a future time. When I tested these cases, things didn't work right.

willcode:
Maybe we should just "do it right" and have the rx signal that it's recording/playing a certain file. Then the IQ tool doesn't need to infer the file name.

PR #1238, opened 2023-05-09
Avoid implicit promotion to double in AGC

### LINKED ISSUE #1232: Get rid of unnecessary promotions to double
By using the `-Wdouble-promotion` compiler flag, we can find places where `float` values are promoted to `double`. The extra precision is usually unintentional, and can have a performance impact. I wouldn't expect much impact on x86 where most floating point calculations are 80-bit, but on other platforms the difference could be larger.

As an experiment, I removed the double promotions from `CAgc::ProcessData` (https://github.com/gqrx-sdr/gqrx/compare/master...argilo:gqrx:agc-speedup), and that reduced the CPU utilization of the AGC block by about 6% on a Raspberry Pi. (I didn't observe any improvement on x86_64.)

The experiment shows that there are performance benefits available on some hot paths, so it's probably worth turning on the `-Wdouble-promotion` flag and fixing the rest of the warnings. I'll do this once #1218 is merged.

argilo:
Partially fixes #1232.

As noted there, removing implicit promotions to double from `CAgc::ProcessData` reduces the AGC block's CPU utilization by about 6% on Raspberry Pi. I've removed implicit promotions from `CAgc::SetParameters` as well, so that compiling with `-Wdouble-promotion` produces no warnings.

### DIFF
--- a/src/dsp/agc_impl.h
+++ b/src/dsp/agc_impl.h
@@ -34,7 +34,7 @@ class CAgc
 public:
     CAgc();
     virtual ~CAgc();
-    void SetParameters(bool AgcOn, bool UseHang, int Threshold, int ManualGain, int Slope, int Decay, double SampleRate);
+    void SetParameters(bool AgcOn, bool UseHang, int Threshold, int ManualGain, int Slope, int Decay, float SampleRate);
     void ProcessData(int Length, const TYPECPX * pInData, TYPECPX * pOutData);
 
 private:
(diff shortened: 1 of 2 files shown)


src/dsp/agc_impl.h
```diff
@@ -34,7 +34,7 @@ class CAgc
 public:
     CAgc();
     virtual ~CAgc();
-    void SetParameters(bool AgcOn, bool UseHang, int Threshold, int ManualGain, int Slope, int Decay, double SampleRate);
+    void SetParameters(bool AgcOn, bool UseHang, int Threshold, int ManualGain, int Slope, int Decay, float SampleRate);
```
argilo, line 37: `SampleRate` is assigned to the `m_SampleRate` member variable, which is a `float`. So there's not much point receiving it as a `double`.

PR #1242, opened 2023-05-09
plotter: limit plot update rate

willcode:
Helps performance, and faster looks worse on the screen.

### DIFF
--- a/src/qtgui/dockfft.ui
+++ b/src/qtgui/dockfft.ui
@@ -653,6 +653,11 @@
               </item>
               <item>
                <property name="text">
+                <string>333 fps</string>
+               </property>
+              </item>
+              <item>
+               <property name="text">
                 <string>250 fps</string>
                </property>
               </item>
--- a/src/qtgui/dockfft.ui
+++ b/src/qtgui/dockfft.ui
@@ -155,7 +155,7 @@
               </property>
               <property name="toolTip">
                <string>Amount of smoothing for plot
-(histogram is not affected)</string>
+(persistence for histogram)</string>
               </property>
               <property name="maximum">
                <number>100</number>
--- a/src/qtgui/dockfft.ui
+++ b/src/qtgui/dockfft.ui
@@ -648,6 +648,11 @@
               </property>
               <item>
                <property name="text">
+                <string>500 fps</string>
+               </property>
+              </item>
+              <item>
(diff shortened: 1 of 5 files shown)


vladisslav2011:
I've almost written a suggestion to use a define/new configuration option instead of hardcoded value, but you are reading my mind. LOL.

src/qtgui/dockfft.ui
```diff
@@ -646,6 +646,11 @@
               <property name="currentIndex">
                <number>0</number>
               </property>
+              <item>
+               <property name="text">
+                <string>500 fps</string>
+               </property>
+              </item>
```
argilo, line 653: It would probably make sense to add 333 as well, to fill in the large gap between 250 and 500.

src/qtgui/plotter.cpp
```diff
@@ -1252,7 +1253,8 @@ void CPlotter::draw(bool newData)
     const double frameTime = 1.0 / (double)fft_rate;
 
     // Redraw the plot if it is visible.
-    const bool doPlotter = !m_2DPixmap.isNull();
+    const bool doPlotter = (!m_2DPixmap.isNull()
+        && tnow_ms >= tlast_plot_drawn_ms + PLOTTER_UPDATE_LIMIT_MS);
```
argilo, line 1257: One quirk of this calculation method is that 30 fps gives a refresh rate of 30 Hz, whereas the next step (40 fps) gives 20 Hz, which is noticeably slower. I'll have to think about whether there's a better way.

argilo: There's also enough jitter in `tnow_ms` that some frames are often skipped even at 30 fps.

willcode: We may want to think about what "frame rate" means for the app. Some stuff can be completely automatic and some stuff matters to the user.

argilo: Agreed. I wonder whether it might be possible to drive things with vsync, so that drawing occurs (at most) once per frame. Animation might be smoother that way.

vladisslav2011:
Current plotter implementation derives update rate from the system clock. It may be interesting to add a setting to choose between the system clock (Qt timer) and frontend clock (sample counter) as a reference clock for the update rate.

Issue #1241, opened 2023-05-09
Settings stored as QComboBox indexes


argilo:
Some settings are stored as integers representing the index of a `QComboBox`. This is problematic, because the choices cannot easily be changed without breaking backwards compatibility. (See #930 for an example of the complexity involved in making such changes.)

In version 2.16, the following settings are stored this way:

* `fft_window`
* `waterfall_span`

#1218 added some more:

* `plot_x_unit`
* `plot_y_unit`
* `plot_mode`
* `waterfall_mode`

Hopefully we can at least fix the latter before releasing 2.17.

PR #1243, opened 2023-05-11
dockfft: use strings for config file settings

willcode:
- plot_y_unit
- plot_x_unit
- plot_mode
- waterfall_mode
- fft_window

Partial fix for #1241

### DIFF
--- a/src/qtgui/dockfft.cpp
+++ b/src/qtgui/dockfft.cpp
@@ -41,6 +41,11 @@
 #define DEFAULT_FFT_AVG         25
 #define DEFAULT_COLORMAP        "gqrx"
 
+static const QStringList window_strs = {
+    "hamming", "hann", "blackman", "rectangular", "kaiser",
+    "blackmanharris", "bartlett", "flattop"
+};
+
 DockFft::DockFft(QWidget *parent) :
     QDockWidget(parent),
     ui(new Ui::DockFft)
@@ -180,6 +185,7 @@ int DockFft::fftSize()
 void DockFft::saveSettings(QSettings *settings)
 {
     int  intval;
+    QString strval;
 
     if (!settings)
         return;
@@ -199,10 +205,10 @@ void DockFft::saveSettings(QSettings *settings)
         settings->remove("fft_rate");
 
     intval = ui->fftWinComboBox->currentIndex();
-    if (intval != DEFAULT_FFT_WINDOW)
-        settings->setValue("fft_window", intval);
-    else
-        settings->remove("fft_window");
+    if ((unsigned int)intval > window_strs.size())
+        intval = DEFAULT_FFT_WINDOW;
+    strval = window_strs[intval];
+    settings->setValue("fft_window", strval);
 
     intval = ui->wfSpanComboBox->currentIndex();
     if (intval != DEFAULT_WATERFALL_SPAN)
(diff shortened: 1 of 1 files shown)


willcode:
FFT window setting has enough options to justify a little nicer code. We could go back later and add a global generic function that makes it easier to do table lookup, and then do some of the other settings this way.

Loading of the FFT rate setting could use some attention. It will call `setFftRate()` with whatever is found in the file, and that function tries to do an exact match search.

Waterfall span would be similar when converted to deal with values.

src/qtgui/dockfft.cpp
```diff
@@ -365,24 +376,32 @@ void DockFft::readSettings(QSettings *settings)
...
+    else if (strval == "fill") intval = 2;
+    else if (strval == "hist") intval = 3;
+    else                       intval = 0;  // "max", default
+    ui->plotModeBox->setCurrentIndex(intval);
+
+    strval = settings->value("waterfall_mode", "max").toString();
+    if      (strval == "avg")  intval = 1;
+    else if (strval == "sync") intval = 2;
+    else                       intval = 0;  // "max", default
+    ui->wfModeBox->setCurrentIndex(intval);
```
argilo, line 404: There's a duplicate `waterfall_mode` decoder down on lines 464-468 that still reads an integer. I presume that was a copy & paste error, and should be deleted.

src/qtgui/dockfft.cpp
```diff
@@ -41,6 +41,11 @@
 #define DEFAULT_FFT_AVG         25
 #define DEFAULT_COLORMAP        "gqrx"
 
+static const std::vector<const char*>window_strs({
+    "hamming", "hann", "blackman", "rectangular", "kaiser",
+    "blackmanharris", "bartlett", "flattop"
+});
```
argilo, line 47: This could be simplified:
```suggestion
static const QStringList window_strs = {
    "hamming", "hann", "blackman", "rectangular", "kaiser",
    "blackmanharris", "bartlett", "flattop"
};
```

src/qtgui/dockfft.cpp
```diff
@@ -352,9 +359,13 @@ void DockFft::readSettings(QSettings *settings)
         setFftSize(intval);
     emit fftSizeChanged(fftSize());
 
-    intval = settings->value("fft_window", DEFAULT_FFT_WINDOW).toInt(&conv_ok);
-    if (conv_ok)
-        ui->fftWinComboBox->setCurrentIndex(intval);
+    strval = settings->value("fft_window", "hann").toString();
+    auto it = std::find(window_strs.begin(), window_strs.end(), strval);
+    if (it == window_strs.end())
+        intval = DEFAULT_FFT_WINDOW;
```
argilo, line 365: Do you suppose we should fall back to integer decoding for backwards compatibility with earlier versions of Gqrx?

argilo:
I notice you've dispensed with removing the setting when it's the default. I think that's fine, since it doesn't seem terribly useful to me.

src/qtgui/dockfft.cpp
```diff
@@ -216,28 +222,28 @@ void DockFft::saveSettings(QSettings *settings)
...
+    else if (intval == 3) strval = "hist";
+    else                  strval = "max";  // 0, default
+    settings->setValue("plot_mode", strval);
 
     intval = ui->wfModeBox->currentIndex();
-    if (intval != 0)
-        settings->setValue("waterfall_mode", intval);
-    else
-        settings->remove("waterfall_mode");
+    if      (intval ==1)  strval = "avg";
```
argilo, line 243: Nitpick:
```suggestion
    if      (intval == 1)  strval = "avg";
```

src/qtgui/dockfft.cpp
argilo: Even better, increment `configversion` to 4 and read the setting like so:

https://github.com/gqrx-sdr/gqrx/blob/282d8559332db7b0d424ae6acbb5edc1af9b22fa/src/qtgui/dockrxopt.cpp#L474-L480

willcode: I was thinking about options for this, and whether it was worth the code to save the user a click. I'm on the road for a few and don't want to rush a solution ... but feel free to add something if you want to get a release out.

argilo: I added a fallback in 91a42eea657f53567f90499f35bfbe2d4b73f9f1.

PR #1247, opened 2023-05-20
plotter: more responsive overlay redraws

### LINKED ISSUE #1244: Plotter does not redraw if "Z" pressed while DSP is stopped
As noted here: https://github.com/gqrx-sdr/gqrx/pull/1200#issuecomment-1546798281

@willcode

willcode:
Do not wait for the next fft frame to redraw overlay elements

Fixes #1244 and also makes the overlay more responsive to changes in general.


argilo:
This seems to work well.

PR #1248, opened 2023-05-20
dockfft: zoom slider uses log scale with max 100000x

### LINKED ISSUE #1235: Freq zoom slider behaves erratically
Sometimes the "Freq zoom" slider refuses to decrease below a certain value, or becomes completely stuck.

Scenario 1:

1. Click "Reset" in the "FFT Settings" panel.
2. Hover over the frequency axis close to the upper frequency and zoom in to ~30x using the mouse wheel.
3. Try to drag the "Freq zoom" slider back down to 1x. It will stop well before, refusing to go any lower.

Scenario 2:

1. Click "Reset" in the "FFT Settings" panel.
2. Hover over the frequency axis close to the upper frequency and zoom in to ~30x using the mouse wheel.
3. Drag the frequency axis far enough to move beyond the end of the FFT, so that part of the display is empty.
4. Try to drag the "Freq zoom" slider back down to 1x. It will jump up to a larger value and become completely stuck.

It remains possible to zoom out by other means, just not using the "Freq zoom" slider.

willcode:
A log scale is more sensible for the zoom slider, and it was difficult to make it work at all with a linear scale.

Fixes #1235 

Also:
- The slider is now log scaled so zoom per tick is consistent over the entire range
- Center frequency stays the same when zooming using the slider - it previously was centered on the demod freq which led to strange behavior when both mouse-based and slider-based zoom were used

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -910,53 +910,48 @@ void CPlotter::zoomStepX(float step, int x)
     }
 
     // calculate new range shown on FFT
-    double new_range = qBound(10.0, m_Span * (double)step, (double)m_SampleFreq * 10.0);
+    float new_span = std::min(
+        (float)m_Span * (float)step, (float)m_SampleFreq);
 
-    // Frequency where event occurred is kept fixed under mouse
-    double ratio = (double)x / (qreal)m_Size.width() / m_DPR;
-    qint64 fixed_hz = freqFromX(x);
-    double f_max = (double)fixed_hz + (1.0 - ratio) * new_range;
-    double f_min = f_max - new_range;
+    // Keep frequency under pointer the same and calculated the offset to the
+    // plot center.
+    float offset = (float)(freqFromX(x) - m_CenterFreq - m_FftCenter);
+    float new_FftCenter = (float)m_FftCenter + offset * (1.0f - step);
 
-    // ensure we don't go beyond the rangelimits
-    bool limit_hit = false;
-    double lolim = (double)(m_CenterFreq) - (double)m_SampleFreq / 2.0;
-    double hilim = (double)m_CenterFreq + (double)m_SampleFreq / 2.0;
-    if (f_min < lolim)
+    // Keep edges of plot in valid frequency range. The plot may need to be
+    // panned.
+    const float max_limit = (float)m_SampleFreq / 2.0f;
+    const float min_limit = - (float)m_SampleFreq / 2.0f;
+    float f_max = new_FftCenter + new_span / 2.0f;
+    float f_min = new_FftCenter - new_span / 2.0f;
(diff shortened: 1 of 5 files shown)


argilo:
After this change, it's no longer possible to zoom all the way out using the slider, because its minimum value is 1, which corresponds to a zoom level of 1.12x:

https://github.com/gqrx-sdr/gqrx/blob/996737cd790c140b35a2b8e35b6cc6e60c7b3ca3/src/qtgui/dockfft.ui#L300-L302

Changing that to 0 seems to solve the problem.

argilo:
> Fixes #1235

This PR implements the suggestion I made in https://github.com/gqrx-sdr/gqrx/issues/1235#issuecomment-1546901455, but it doesn't solve the bugs described in #1235; I verified that they're still present.

willcode:
I'll take a look.

argilo:
This is mostly looking good, but there are still a couple problems that I think should be addressed:

1. When zooming out using the mouse wheel while hovering over the frequency axis, it's now possible to zoom out beyond 1x, leaving black on both sides of the plot, and the Freq Zoom indicator showing 0x:
Screenshot from 2023-06-21 16-54-50
1. When zooming out using the mouse wheel or the Freq Zoom slider, it's now much easier to end up with empty black space on the left or right of the plot & waterfall. Perhaps we should disallow this completely.

src/qtgui/plotter.cpp
```diff
@@ -1525,6 +1502,7 @@ void CPlotter::draw(bool newData)
     {
         m_2DPixmap.fill(PLOTTER_BGD_COLOR);
         QPainter painter2(&m_2DPixmap);
+        painter2.translate(QPointF(0.5, 0.5));
```
argilo, line 1505: I think this change (and the corresponding change in `drawOverlay` deserves to be in its own PR, since it's unrelated to zooming.

willcode:
PRs split. I'll take a look at zoom/pan limits.

willcode:
Zoom and pan are now limited to avoid blank space.

willcode:
There are still some problems with this ... still working ...

willcode:
OK, ready for you to try out.

argilo:
> if you're fully zoomed out

The same behaviour also happens if you're zoomed in and are close to the upper or lower frequency limit.

argilo:
No, it's not. It should always be possible to zoom in around the mouse cursor, regardless of its position, because the frequency limits are moving towards the cursor frequency.

The problem here is that `setFftCenterFreq` doesn't know that there's going to be a new, decreased span. So it thinks that it clamps the center frequency based on the old span, not the new.

willcode:
OK, I was looking at a different effect ... will fix.

argilo:
Your update looks like it should fix it. Taking it for a spin now.

PR #1249, opened 2023-05-22
nbrx: make extra room for history in rx filter

### LINKED ISSUE #1233: Setting the filter width too narrow causes DSP to hang
Fedora 38 with gqrx-2.15.9-7.fc38.x86_64. (However, I was able to reproduce the same issue with latest Gqrx master.)

DSP hangs after setting the filter width too narrow using Ctrl + mouse wheel. See the following screencast:

[filter_width_hang.webm](https://user-images.githubusercontent.com/2339930/235373018-cec9cc1c-27b9-4a74-8df3-0649488a964e.webm)

This gets printed on stdout:
```
block_executor :error: sched: <block fir_filter_blk<IN_T,OUT_T,TAP_T> (14)> is requesting more input data  than we can provide.  ninput_items_required = 11570  max_possible_items_available = 8191  If this is a filter, consider reducing the number of taps.
```

willcode:
The default output buffer allocation for the nb (noise blanker) block in nbrx could end up being smaller than the history requirement of the rx filter when filter params are adjusted at runtime. Allocate space for 64k items, which should be enough for a very tight filter.

Fixes #1233

### DIFF
--- a/src/receivers/nbrx.cpp
+++ b/src/receivers/nbrx.cpp
@@ -43,6 +43,7 @@ nbrx::nbrx(float quad_rate, float audio_rate)
     iq_resamp = make_resampler_cc(PREF_QUAD_RATE/d_quad_rate);
 
     nb = make_rx_nb_cc(PREF_QUAD_RATE, 3.3, 2.5);
+    nb->set_min_output_buffer(64 * 1024);
     filter = make_rx_filter(PREF_QUAD_RATE, -5000.0, 5000.0, 1000.0);
     agc = make_rx_agc_cc(PREF_QUAD_RATE, true, -100, 0, 0, 500, false);
     sql = gr::analog::simple_squelch_cc::make(-150.0, 0.001);


vladisslav2011:
I've tried this and it didn't work.

willcode:
You ran this patch, or you've tried something like this before?

vladisslav2011:
1. I've tried this before.
2. I've checked out the commit before my fix (direct implementation of the FIR filter) and added this one line.

It does not work at least on my system.

willcode:
Since the quad rate is fixed at 96k and the filter widths are going to be the same, I'm not sure how we're seeing different results. What do you see if you add this to master? How much space did the error message say it was looking for?

vladisslav2011:
Master does not compile, so I see nothing.

willcode:
What version of GR?

vladisslav2011:
```
linux; GNU C++ version 7.3.0; Boost_106501; UHD_003.010.003.000-0-unknown

qt5ct: using qt5ct plugin
gr-osmosdr 0.1.5 (0.1.5) gnuradio 3.7.11
```

willcode:
No, it's going to be more complicated than this. The GR buffer allocator is not doing what I thought it would do.

vladisslav2011:
Thing are getting more and more interesting here.
Gqrx segfaults when built with current GNU Radio main, so I can't test this at all...

willcode:
Was it a segfault once after build, then worked after that? I've seen that occasionally and don't know what that is about. If it's a constant segfault, I haven't seen it.

vladisslav2011:
The segfault is not related to this change.
I've made it work. That was HackRF and wrong gr-osmosdr branch without fixes.

PR #1250, opened 2023-05-23
Zoom the plotter faster with Ctrl

LevitatingBusinessMan:
Holding control when scrolling will allow you to zoom faster.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -3,6 +3,7 @@
 
        NEW: Delete key clears the waterfall.
        NEW: I/Q tool can save recordings in SigMF format.
+       NEW: Holding Ctrl speeds up mouse wheel zoom.
   IMPROVED: Reduced CPU utilization of waterfall display.
    CHANGED: DMG release requires macOS 12.7 or later.
 


argilo:
Thanks for the pull request, and sorry for the long delay in looking at this.

Some changes were made to zooming in #1270, so this code would need to be updated to adjust the 0.9 base value in the new `pow` calculation.

That said, I'm not sure whether this is a feature that would be used much, particularly since keyboard modifiers to mouse movements are not particularly discoverable. What was your motivation for making this change?

LevitatingBusinessMan:
I got tired of having to scroll really aggressively to quickly identify a signal. The zoom speed is fine most of the time but a bit slow when you want to go from seeing the entire band to a tiny signal. At the time I was using some other piece of software that allowed to speed up scrolling by holding a modifier key. It might've been some other SDR software.

Using modifiers with the mouse isn't at all uncommon. Think of 2d and 3d creative software. And even on the webpage you're viewing this on, you can zoom the window using Ctrl and the scrollwheel.

I'll update the patch so that you can try it.

As for the discoverability, you could opt to include it in the little info box at the bottom that shows other modifier keys when hovering over the plotter.

argilo:
I use modifier keys infrequently enough that I had forgotten that Ctrl+wheel adjusts the filter bandwidth when hovering over the plot. I suppose it wouldn't hurt to add Ctrl+wheel functionality to zooming. Do you think the Y axis should be included for consistency?

argilo:
I refactored the code slightly, and included both axes. @LevitatingBusinessMan Let me know what you think.

PR #1251, opened 2023-05-31
plotter: change "and" to "&&"

willcode:
As pointed out by @gvanem, I was thinking Python while writing C++.


vladisslav2011 (approved):

gvanem:
That was fast!

argilo:
It would be great if we could add Windows to CI so we could avoid MSVC issues.

Issue #1115, opened 2022-04-10
Wav recording volume too high for NFM demodulator


AsciiWolf:
Fedora 35, Gqrx 2.15.8-2 (x86_64), SDRplay RSPdx.

When trying to record any NFM signal, the recorded file sounds distorted because the volume is obviously too high for some reason (although the actual reception in Gqrx is fine). Setting the Gain slider to a lower volume in Gqrx did not help.

[Here](https://github.com/gqrx-sdr/gqrx/files/8458926/test_recording.zip) ia an example recording. It seems to happen for all NFM signals, but WFM signals sound fine. AM or SSB signals also sound fine. I have tried recording the same NFM signals in another SDR software and everything was fine there.

The issue did not happen in older versions of Gqrx, it started happening recently (not more than a few months ago).

vladisslav2011:
Hello.
I have experienced this issue too.
I my case even some WFM recordings have been distorted die to high nonstandard deviation, that Gqrx can't handle well.

> Setting the Gain slider to a lower volume in Gqrx did not help.

The "gain" slider applies to audio sink only. It has nothing to do with streaming/recording.

I have fixed similar issue by replacing old audio AGC implementation with new one and putting the new AGC block into audio path regardless of demodulator selected.

Here are PR series:
https://github.com/gqrx-sdr/gqrx/pull/1056
https://github.com/gqrx-sdr/gqrx/pull/1060
https://github.com/gqrx-sdr/gqrx/pull/1061
https://github.com/gqrx-sdr/gqrx/pull/1096

You may try the latest one (as it includes 1-3 and have most bugs fixed) and give some feedback.

AsciiWolf:
Thanks for the info and links to the PRs! I will try testing it and providing feedback when I have some time.

vladisslav2011:
Hmmm.
Have you tried setting wider NFM deviation from a popup dialog:
![dialog](https://user-images.githubusercontent.com/17007837/164987080-ed52c235-5d73-4755-84d2-30f2224bb606.png)
![deviation](https://user-images.githubusercontent.com/17007837/164986959-266df14c-a21e-4560-a364-304332c0446a.png)
This may be enough if you don't have mixed wide/narrow transmissions at the same channel or do not want audio normalization
.

AsciiWolf:
Thanks! Setting the max deviation to 17 kHz did work around the issue (but made weaker NFM transmissions quieter/harder to understand).

[...]

PR #1255, opened 2023-06-17
Add 6 dB attentuation to WAV file recordings

argilo:
Partially addresses #1115.

I think it makes sense to include some attenuation in the WAV file recording path, to add some headroom to account for things like tuning offset when recording FM signals. The audio output defaults to -6 dB, so I think that's a sensible choice for WAV file output.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -22,6 +22,7 @@
   IMPROVED: Plot can be zoomed & resized while DSP is stopped.
   IMPROVED: FFT window setting is stored as a string.
   IMPROVED: Waterfall span setting is stored in milliseconds.
+  IMPROVED: Added 6 dB attenuation to WAV file recordings to prevent clipping.
    CHANGED: Default narrow FM deviation increased to 5 kHz.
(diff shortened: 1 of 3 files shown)


vladisslav2011:
It looks like hiding a problem instead of solving it.
Time to open a new issue: "WAV recordings are too quiet"?
Isn't it better to add a global AGC, like I did and connect WAV recorder and audio sink to it's output?

argilo:
AGC is not a perfect solution either, because it distorts the input signal. I think adding a 6 dB of headroom is a reasonable solution for now.

vladisslav2011:
I agree, that AGC is not perfect, but distortion can be managed by increasing rise/fall times and limiting maximum gain. At zero gain it will just attenuate peaks and not touch the signal within -1.0...1.0 range...
And, of course, the AGC may be switched off completely and it's gain may be controlled manually if somebody knows, that the signal would never reach some level and wants to avoid distortion.

argilo:
I agree that post-demod AGC could be useful, and may consider it eventually.

PR #1259, opened 2023-06-21
plotter: draw on pixel center for cleaner picture

willcode:
In the Qt coordinate system, a shift of 0.5 is required to draw on pixel centers. Without this shift, lines come out more than one pixel wide.

In the main drawing routine, shift by (0.5, 0.5). In the overlay drawing routine, shift by (-0.5, -0.5) since the result is copied by the main drawing routine (the shifts cancel out).

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1525,6 +1525,7 @@ void CPlotter::draw(bool newData)
     {
         m_2DPixmap.fill(PLOTTER_BGD_COLOR);
         QPainter painter2(&m_2DPixmap);
+        painter2.translate(QPointF(0.5, 0.5));
 
         // Update histogram IIR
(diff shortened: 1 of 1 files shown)


argilo:
This definitely improves rendering of the plot line.

I notice some remaining off-by-0.5 issues: the center line and red tuning line are, at certain zoom levels, rendered one pixel to the right of a dotted frequency line, even when the two frequencies are equal. Is this a separate issue, or should it be fixed here?

argilo:
Here's a screenshot showing things off by a pixel in both cases:

Screenshot from 2023-06-21 20-26-17

argilo:
I checked, and that issue is unrelated to the changes here, so I guess it can be treated separately.

The change here seem to work properly, so I'll merge.

argilo:
It appears the root cause of the misalignment is that xFromFreq is not high-DPI aware, but the grid lines are.

PR #1260, opened 2023-06-22
Fix some more double promotion warnings

argilo:
This fixes all remaining double promotion warnings, except for the ones in `CPlotter::draw`, which I think warrant a more careful analysis.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -348,9 +348,9 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
         {
             setCursor(QCursor(Qt::ClosedHandCursor));
             // move Y scale up/down
-            qreal delta_px = m_Yzero - py;
-            qreal delta_db = delta_px * fabs(m_PandMindB - m_PandMaxdB) /
-                             (qreal)h;
+            float delta_px = m_Yzero - py;
+            float delta_db = delta_px * fabsf(m_PandMindB - m_PandMaxdB) /
+                             (float)h;
             m_PandMindB -= delta_db;
             m_PandMaxdB -= delta_db;
             if (out_of_range(m_PandMindB, m_PandMaxdB))
@@ -965,7 +965,7 @@ void CPlotter::zoomStepX(float step, int x)
 // Zoom on X axis (absolute level)
 void CPlotter::zoomOnXAxis(float level)
 {
-    double current_level = (double)m_SampleFreq / (double)m_Span;
+    float current_level = (float)m_SampleFreq / (float)m_Span;
     zoomStepX(current_level / level, xFromFreq(m_DemodCenterFreq));
     updateOverlay();
 }
@@ -1017,9 +1017,9 @@ void CPlotter::wheelEvent(QWheelEvent * event)
         // Vertical zoom. Wheel down: zoom out, wheel up: zoom in
         // During zoom we try to keep the point (dB or kHz) under the cursor fixed
         float zoom_fac = delta < 0 ? 1.1 : 0.9;
-        float ratio = (qreal)py / (qreal)(h);
+        float ratio = (float) py / (float) h;
         float db_range = m_PandMaxdB - m_PandMindB;
-        auto y_range = (qreal)(h);
+        float y_range = (float) h;
         float db_per_pix = db_range / y_range;
         float fixed_db = m_PandMaxdB - py * db_per_pix;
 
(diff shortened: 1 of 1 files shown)


src/qtgui/plotter.cpp
```diff
@@ -348,9 +348,9 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
         {
             setCursor(QCursor(Qt::ClosedHandCursor));
             // move Y scale up/down
-            qreal delta_px = m_Yzero - py;
-            qreal delta_db = delta_px * fabs(m_PandMindB - m_PandMaxdB) /
-                             (qreal)h;
+            float delta_px = m_Yzero - py;
+            float delta_db = delta_px * fabsf(m_PandMindB - m_PandMaxdB) /
+                             (float)h;
```
argilo, line 353: `m_PandMindB` and `m_PandMaxdB` are floats, so I think it makes sense to calculate the deltas as floats.

src/qtgui/plotter.cpp
```diff
@@ -965,7 +965,7 @@ void CPlotter::zoomStepX(float step, int x)
 // Zoom on X axis (absolute level)
 void CPlotter::zoomOnXAxis(float level)
 {
-    double current_level = (double)m_SampleFreq / (double)m_Span;
+    float current_level = (float)m_SampleFreq / (float)m_Span;
```
argilo, line 968: Zoom stuff generally uses floats, so I think it makes sense to do so here.

src/qtgui/plotter.cpp
```diff
@@ -1017,9 +1017,9 @@ void CPlotter::wheelEvent(QWheelEvent * event)
         // Vertical zoom. Wheel down: zoom out, wheel up: zoom in
         // During zoom we try to keep the point (dB or kHz) under the cursor fixed
         float zoom_fac = delta < 0 ? 1.1 : 0.9;
-        float ratio = (qreal)py / (qreal)(h);
+        float ratio = (float) py / (float) h;
         float db_range = m_PandMaxdB - m_PandMindB;
-        auto y_range = (qreal)(h);
+        float y_range = (float) h;
```
argilo, line 1022: Float seems OK for consistency here.

src/qtgui/plotter.cpp
```diff
@@ -1936,8 +1936,8 @@ void CPlotter::setNewFftData(const float *fftData, int size)
     if (needIIR) {
         for (int i = 0; i < size; ++i)
         {
-            const double v = m_fftData[i];
-            const double iir = m_fftIIR[i];
+            const float v = m_fftData[i];
+            const float iir = m_fftIIR[i];
```
argilo, line 1940: `m_fftData` and `m_fftIIR` are float anyway.

src/qtgui/plotter.cpp
```diff
@@ -2184,8 +2184,8 @@ void CPlotter::drawOverlay()
                 qMin(w / (metrics.boundingRect(label).width() + metrics.boundingRect("O").width()),
                      (qreal)HORZ_DIVS_MAX),
                 m_StartFreqAdj, m_FreqPerDiv, m_HorDivs);
-    pixperdiv = w * (float) m_FreqPerDiv / (float) m_Span;
-    adjoffset = pixperdiv * float (m_StartFreqAdj - StartFreq) / (float) m_FreqPerDiv;
+    pixperdiv = w * (qreal) m_FreqPerDiv / (qreal) m_Span;
+    adjoffset = pixperdiv * (qreal) (m_StartFreqAdj - StartFreq) / (qreal) m_FreqPerDiv;
```
argilo, line 2188: Explicitly promoting to `qreal` since we're dealing with coordinates.

src/qtgui/plotter.cpp
```diff
@@ -2227,8 +2227,8 @@ void CPlotter::drawOverlay()
     dbstepsize = (float) dbDivSize;
     mindbadj = mindBAdj64;
 
-    pixperdiv = h * (float)dbstepsize / (m_PandMaxdB - m_PandMindB);
-    adjoffset = h * (mindbadj - m_PandMindB) / (m_PandMaxdB - m_PandMindB);
+    pixperdiv = h * (qreal) dbstepsize / (qreal) (m_PandMaxdB - m_PandMindB);
+    adjoffset = h * (mindbadj - (qreal) m_PandMindB) / (qreal) (m_PandMaxdB - m_PandMindB);
```
argilo, line 2231: Explicitly promoting to `qreal` since we're dealing with coordinates.

src/qtgui/plotter.cpp
willcode: I'd leave this one, since v can be much smaller than iir (this is linear scale). I think I did this after seeing some sort of clipping effect.

willcode (approved):
One that shouldn't change, all of the others look safe.

src/qtgui/plotter.cpp
argilo: Strange. I wouldn't expect precision issues to come up when multiplying or dividing floats with different magnitudes. Anyway, I'll leave this change out for now and we can come back to it.

PR #1262, opened 2023-06-22
plotter: use volk kernels for some calculations

willcode:
VOLK may give a little performance boost in this calculation. It's not much on my machine, so let's see how this works on other kinds of machines.


argilo:
This saves about 3% CPU on Raspberry Pi when the FFT size is 1M and rate is 2 fps.

argilo:
I rebased to resolve a merge conflict, and will merge once CI passes.

PR #1263, opened 2023-06-23
Switch to float in CPlotter::draw

argilo:
After #1260 and #1261, the only remaining double promotion warnings were in `CPlotter::draw`. To fix them, I converted all calculations in this method from `double` to `float`. There were a few instances in other methods which had to be changed as well.

These changes noticeably increase performance of the histogram mode on my Raspberry Pi; CPU utilization drops by more than 10%. Performance on x86 is not likely to increase much, since double arithmetic is no slower than float.

Since all double promotion warnings are now fixed, I've turned on these warnings in CMake.

I'll leave this as a draft until I've had a chance to double check the changes to make sure none of the calculations require double precision.


src/qtgui/plotter.cpp
```diff
@@ -1218,34 +1218,34 @@ void CPlotter::draw(bool newData)
...
-    const double sampleFreq = m_SampleFreq;
-    const double fftCenter = (double)m_FftCenter;
-    const double span = (double)m_Span;
-    const double startFreq = fftCenter - span / 2.0;
-    const double binsPerHz = fftSize / sampleFreq;
+    const float fftSize = m_fftDataSize;
+    const float sampleFreq = m_SampleFreq;
+    const float fftCenter = (float)m_FftCenter;
+    const float span = (float)m_Span;
+    const float startFreq = fftCenter - span / 2.0f;
```
argilo, line 1229: Frequencies are one place where we may well need double precision, so I'll see about putting this back.

argilo: Or maybe not. These frequencies are relative to the FFT center, so they might not have the same precision requirement that an absolute frequency would.

argilo: I did end up switching frequency calculations back to double, because quantization was beginning to become visible far from the center frequency rate when using high sample rates. (For example, frequencies were quantized to 2 Hz when more than 16 MHz above or below the center frequency.)

willcode:
`m_wfbuf` is potentially affected by `float` in avg mode. I don't see effects in a quick test, but the effective resolution could be lowered with extreme waterfall interval and fft rate settings.

Overall, I don't see any effects. It's probably a good idea to go all `float` then find the places `double` is needed again, since I can't remember the exact places that caused quantization/clamping, and the code may have changed enough that those places don't exist any more.

PR #1265, opened 2023-07-21
nbrx: set minimum buffer size for input to rx_filter

### LINKED ISSUE #1233: Setting the filter width too narrow causes DSP to hang

willcode:
Width of rx_filter can be adjusted at run time, so the input buffer (the output buffer of nb) needs to be large enough for the longest history required by the filter (Narrow/Sharp). This setting may not be reliable for GR prior to v3.10.7.0.

Fixes #1233

### DIFF
--- a/src/receivers/nbrx.cpp
+++ b/src/receivers/nbrx.cpp
@@ -57,7 +57,7 @@ nbrx::nbrx(float quad_rate, float audio_rate)
     // output buffer of nb) needs to be large enough for the longest history
     // required by the filter (Narrow/Sharp). This setting may not be reliable
     // for GR prior to v3.10.7.0.
-    nb->set_min_output_buffer(64 * 1024 * sizeof(gr_complex));
+    nb->set_min_output_buffer(32768);
 
     audio_rr0.reset();
     audio_rr1.reset();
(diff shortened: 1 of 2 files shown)


src/receivers/nbrx.cpp
```diff
@@ -53,6 +53,12 @@ nbrx::nbrx(float quad_rate, float audio_rate)
     demod_am = make_rx_demod_am(PREF_QUAD_RATE, true);
     demod_amsync = make_rx_demod_amsync(PREF_QUAD_RATE, true, 0.001);
 
+    // Width of rx_filter can be adjusted at run time, so the input buffer (the
+    // output buffer of nb) needs to be large enough for the longest history
+    // required by the filter (Narrow/Sharp). This setting may not be reliable
+    // for GR prior to v3.10.7.0.
+    nb->set_min_output_buffer(32768);
```
argilo, line 60: @willcode I removed the `sizeof(gr_complex)` here, because `set_min_output_buffer` takes the number of items. I also decreased the value so that we don't end up with higher latency than required. The largest number of taps I could generate (upper sideband, sharp filter, Ctrl+scroll wheel to narrow to 100 Hz) was 23127, so I chose the next power of 2.

PR #1270, opened 2023-08-11
Properly handle smooth scrolling

sultanqasim:
On computers with smooth scrolling that issue frequent scroll events with small deltas (such as Mac laptops), the current scrolling logic behaves very poorly, as each event zooms the same amount regardless of its size (delta), resulting in overly fast and jittery zoom, and the cumulative effects of non-zooming small scroll events are lost due to rounding and integer discretization. This new logic handles smooth scrolling trackpads/mice properly for zooming, filter adjustment, and frequency adjustment, while also handling old fashioned mice with infrequent large scroll events correctly too.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1025,6 +1025,9 @@ void CPlotter::wheelEvent(QWheelEvent * event)
             m_PandMaxdB = FFT_MAX_DB;
 
         m_PandMindB = m_PandMaxdB - db_range;
+        if (m_PandMindB < FFT_MIN_DB)
+            m_PandMindB = FFT_MIN_DB;
+
         m_MaxHoldValid = false;
         m_MinHoldValid = false;
         m_histIIRValid = false;
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -142,6 +142,7 @@ CPlotter::CPlotter(QWidget *parent) : QFrame(parent)
     m_PandMaxdB = m_WfMaxdB = 0.f;
     m_PandMindB = m_WfMindB = FFT_MAX_DB;
 
+    m_CumWheelDelta = 0;
     m_FreqUnits = 1000000;
     m_CursorCaptured = NOCAP;
     m_Running = false;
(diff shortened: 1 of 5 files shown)


vladisslav2011 (commented):
I'm sometimes using this feature (zooming past FFT_MIN_DB) to examine audio filter/resampler stopband attenuation and ripple. So it may be better to leave it as is or make FFT_MIN_DB configurable at runtime...

vladisslav2011:
I've tested this PR. It works well on my laptop. 
I think, CFreqCtrl needs similar fix as it's really hard to change the digit exactly one step with trackpoint. It either does not change or starts rolling too fast.

sultanqasim:
> I'm sometimes using this feature (zooming past FFT_MIN_DB) to examine audio filter/resampler stopband attenuation and ripple. So it may be better to leave it as is or make FFT_MIN_DB configurable at runtime...

Thanks for testing.

Zooming/scrolling below FFT_MIN_DB causes the plotter's scroll logic to become glitchy, so if you need to go lower, a better solution would be reducing FFT_MIN_DB or making it configurable. As is, being able zoom below FFT_MIN_DB is a bug.

sultanqasim:
> I've tested this PR. It works well on my laptop. I think, CFreqCtrl needs similar fix as it's really hard to change the digit exactly one step with trackpoint. It either does not change or starts rolling too fast.

Great point on CFreqCtrl, I actually forgot it supported frequency change with the scroll wheel till you mentioned it. I'll add an extra commit to this PR to fix that shortly.

sultanqasim:
I now implemented a similar accumulator of scroll events for CFreqCtrl, so it's fixed there too.

src/qtgui/plotter.cpp
```diff
@@ -2238,22 +2248,23 @@ void CPlotter::drawOverlay()
     }
 
     // draw amplitude values (y axis)
-    for (int i = 0; i < m_VerDivs; i++)
+    for (int i = 0; i <= m_VerDivs; i++)
     {
         qreal y = h - ((double)i * pixperdiv + adjoffset);
         qreal th = metrics.height();
+        qreal th_2 = th / 2;
```
willcode, line 2255: Minor: adding another variable doesn't really make things clearer or improve performance.

sultanqasim: I was reusing the same value three times, so I felt having it as a variable was cleaner, but that’s a matter of opinion.

sultanqasim: I can remove the extra variable and just to th/2 in the added comparison

sultanqasim: Changed it back to th / 2, I decided the change is cleaner without introducing an extra variable.

argilo:
Thanks for this, and sorry for the long delay in reviewing it.

So far it looks like touchpad scrolling is working much better on Mac. I'll do a bit more testing and code review before getting this merged in.

argilo:
Touchpad scrolling also appears to work well on Ubuntu.

PR #1274, opened 2023-08-27
Save FFT plotter center frequency in settings

sn4kebite:
Store the selected center frequency in the plotter as `fft/fft_center`. This will make the plotter center consistent on startup when zoomed in.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -12,6 +12,7 @@
        NEW: "Avg" waterfall mode, which displays average of FFT bins.
        NEW: "Sync" waterfall mode, which mirrors averaged data from plot.
        NEW: Support for RTL-SDR Blog V4 in AppImage and DMG releases.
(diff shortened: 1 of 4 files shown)


argilo:
Thanks!

This seems to work well, and it fixes what was probably an unintended side effect of #1248 (namely, that the demod frequency would no longer be in view upon startup, because the zoom slider no longer used the demod frequency as its center).

argilo:
The following undefined behaviour occurs on startup:

`/home/argilo/prefix_310/src/gqrx/src/qtgui/plotter.cpp:918:56: runtime error: signed integer overflow: -9223372036854775808 - 1 cannot be represented in type 'long long int'`

This looks to be due to an existing bug, where `CPlotter::freqFromX` performs a division by zero due to `m_Size.width()` and `m_DPR` being zero before the UI is rendered. I'll add a fix here.

PR #1280, opened 2023-09-20
Add support for RTL-SDR Blog V4

### LINKED ISSUE #1272: Add support for the RTL-SDR Blog V4 to AppImage and dmg packages
Hi, I've just released the RTL-SDR Blog V4 (https://www.rtl-sdr.com/rtl-sdr-blog-v4-dongle-initial-release/) which is based on the R828D tuner chip. But the hardware is slightly different as it uses a different LO frequency for the R828D compared to what is hardcoded in the drivers, and the RF inputs need to be switched in a different way.

I've created a fork of the Osmocom drivers here https://github.com/rtlsdrblog/rtl-sdr-blog-minimal, and a more heavily modified one here with other unrelated changes https://github.com/rtlsdrblog/rtl-sdr-blog. These repo's were modified to add support for the V4, and the -minimal fork has the minimal amount of changed to make it work. I've submitted a patch to Osmocom to try and upstream the main changes for the V4, but the Osmocom rtl-sdr project might be dead as I see that patch requests have been unanswered for a few years. There are no changes that break compatibility with other RTL-SDR dongles.

If it's not upstreamed is there some other way to support the RTL-SDR Blog V4?

On Windows it's easy as I can just tell users to replace the .dll file, and if installed natively on Linux users just need to update the system RTL-SDR driver. But for solutions that prepackage the rtlsdr driver together with the software like AppImage and dmg for Mac's there is a problem. 

I see at https://github.com/gqrx-sdr/gqrx/blob/master/.github/workflows/build.yml#L119 you install librtlsdr from apt and brew. But perhaps you could install and compile from the https://github.com/rtlsdrblog/rtl-sdr-blog-minimal fork instead? Or the osmocom GitHub if it is indeed ever upstreamed?

argilo:
Closes #1272.

Here I've added support for RTL-SDR Blog V4 dongles. On Linux, the `master` branch of rtl-sdr is built from source, and on macOS the `HEAD` version is built using Homebrew. Because @rtlsdrblog's gr-osmosdr patch has not yet been accepted upstream, I've temporarily forked gr-osmosdr and applied the patch.

Pre-release binaries (AppImage and DMG) can be downloaded in the "Artifacts" section here: https://github.com/gqrx-sdr/gqrx/actions/runs/6244124081

@rtlsdrblog I'd appreciate your feedback on the pre-release versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,12 +39,22 @@ jobs:
             libairspyhf-dev \
             libfreesrp-dev \
             libhackrf-dev \
-            librtlsdr-dev \
+            libusb-1.0-0-dev \
             libsoapysdr-dev \
             soapysdr-module-remote \
             libuhd-dev \
             liborc-0.4-dev \
             libhidapi-dev
+
+          cd /tmp
+          git clone https://gitea.osmocom.org/sdr/rtl-sdr.git
+          cd rtl-sdr
+          mkdir build
+          cd build
+          cmake -DCMAKE_INSTALL_PREFIX=/usr -DDETACH_KERNEL_DRIVER=ON ..
+          make -j4
+          make install
+          ldconfig
 
           cd /tmp
           git clone https://github.com/Nuand/bladeRF.git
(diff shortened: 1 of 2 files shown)


argilo:
Thanks for testing! The changes are also working well here, so I'll merge them in.

Your upconverter patch should automatically be picked up by the next build once it is accepted upstream.

PR #1281, opened 2023-09-27
Fix inconsistent 1-2-5 scaling of amplitude axis

argilo:
If the amplitude axis is zoomed to a level that is close to a 1-2-5 threshold (for instance, between 2 dB steps and 5 dB steps), then dragging the axis vertically will result in the axis flickering randomly between 2 dB and 5 dB steps. This happens because the floating-point `m_PandMindB` and `m_PandMaxdB` are converted independently to integers (by rounding down) when passed to the `calcDivSize` function. As the axis is dragged, `m_PandMindB` and `m_PandMaxdB` cross integer thresholds at different times, so the difference varies by 1 dB.

To fix this problem, I've changed the code to convert the difference `m_PandMaxdB - m_PandMindB` to an integer, then add that to the integer version of `m_PandMindB`. That way, the difference between the endpoints to `calcDivSize` remains consistent.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -36,6 +36,7 @@
      FIXED: Crash in AFSK1200 decoder.
      FIXED: Y axis in saved waterfall is too narrow.
      FIXED: Hang when setting a very narrow filter width.
+     FIXED: Inconsistent 1-2-5 scaling of amplitude axis.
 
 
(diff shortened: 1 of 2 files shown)


argilo:
This seems to be working well. Merging.

PR #1285, opened 2023-09-29
Remove unnecessary clamping checks

argilo:
Clamping was removed from `xFromFreq` in #1248. There are a few places in the plotter that explicitly checked for clamping to avoid drawing elements against the edge of the window when `xFromFreq` clamped their X coordinate. These checks are no longer necessary, since the clamping is no longer present.

In cases where the drawn elements are completely outside the visible area, Qt will simply not draw them.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -2129,11 +2129,8 @@ void CPlotter::drawOverlay()
     if (m_CenterLineEnabled)
     {
         x = xFromFreq(m_CenterFreq);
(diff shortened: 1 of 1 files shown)


argilo:
A nice side effect of this change is that A/B label heads remain visible even when the lines are slightly off-screen.

argilo:
The diff is easier to review with whitespace changes omitted: https://github.com/gqrx-sdr/gqrx/pull/1285/files?w=1

willcode (approved):

Issue #1282, opened 2023-09-27
Minor issues in the new plotter 


argilo:
The new plotter (introduced in #1218, and soon to be released in version 2.17) has a few minor issues:

- [x] There is a row of black pixels on the left side of the waterfall (except in the most recent FFT).
- [x] At some zoom levels, there is an empty pixel at the right hand side of the plot and waterfall.
- [x] Dragging the mouse over the plotter or waterfall to change the demod frequency causes the peak/max/min display to reset.
- [x] When the DSP is paused, zooming in or out on the frequency axis causes the peak/max/min display to be one "frame" behind.
- [x] Parts of the overlay (especially the center line and demod line) can obscure thin peaks in the FFT plot.
- [x] In the band plan, labels for wide bands are often displayed off-screen.
- [ ] X coordinates are sometimes slightly misaligned due to remnants of integer-based pixel calculations, for instance in `xFromFreq`.
- [x] Min hold sometimes doesn't work when the DSP is first started, because `rx_fft_c::get_fft_data` returns zeroes if no samples have arrived yet from the SDR hardware.

Since the issues are minor, they don't necessarily need to be fixed prior to the 2.17 release.

argilo:
The 2.17 release is out, and most of these issues have been fixed. I'll split the coordinate issue off.

PR #1287, opened 2023-09-30
Add transparency to plotter overlay elements

argilo:
Partially addresses #1282.

Since the overlay is now drawn on top of the FFT plot, some transparency is needed so that thin peaks are not obscured. I got rid of some existing `setOpacity` calls (which become unwieldy when there are a lot of them) and moved transparency into the colour values. Because the `QColor` constructor ignores the alpha value, it's necessary to use the `QColor::fromRgba` helper instead.

Before:
Screenshot from 2023-09-30 08-52-14

After:
Screenshot from 2023-09-30 09-20-05

### DIFF
--- a/src/qtgui/meter.cpp
+++ b/src/qtgui/meter.cpp
@@ -176,7 +176,6 @@ void CMeter::draw()
         pen.setJoinStyle(Qt::MiterJoin);
         painter.setPen(pen);
         painter.setBrush(QBrush(color));
-        painter.setOpacity(1.0);
 
         painter.drawRect(QRectF(marg, ht + 2, x - marg, 4));
     }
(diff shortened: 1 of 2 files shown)


argilo:
The grid lines were a bit brighter than before, so I decreased their alpha value from 0x40 to 0x28.

argilo:
The filter box was also a bit bright brighter than before, so I decreased its alpha as well.

willcode (approved):
Looks good. The grid doesn't show well over the histogram, but I think that's ok.

argilo:
> The grid doesn't show well over the histogram

A darker colour with a higher alpha might fix that. I'll try it out.

argilo:
@willcode This looks decent. What do you think?

```cpp
#define PLOTTER_GRID_COLOR          0x80606060
```

willcode:
That looks good. There's always a band in some color scheme where it won't show, so don't try too hard. The solution with the numbers and peaks was to put in a drop shadown, but we don't really want that with the fine grid.

PR #1288, opened 2023-09-30
Don't run FFTs until enough samples have arrived

argilo:
Partially addresses #1282.

The new "Min hold" feature sometimes fails if it's enabled when DSP is started for the first time. This happens because the `rx_fft` block is initially filled with zeroes, and will return them through `get_fft_data` if the plot is updated before any samples have arrived from the hardware.

To fix this, I've added sample counting to the `rx_fft` block. If not enough samples have arrived yet, the `get_fft_data` method returns -1, and the plot is not updated.

After the change, the "Min hold" feature works properly on startup, even with RTL-SDR.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -29,6 +29,7 @@
   IMPROVED: Reduced CPU utilization of demodulators.
   IMPROVED: Properly handle smooth scrolling.
   IMPROVED: Shorten band names to fit in narrow spaces.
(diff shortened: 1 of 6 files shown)


willcode (approved):
Tried a few different SDRs and they all worked as expected, including a bit of a delay for low sample rates and high FFT. Using a 4M FFT on a sub-1MS/s SDR causes quite an overlap anyway, so some non-real-time behavior should be expected and this shouldn't shock anyone.

PR #1290, opened 2023-09-30
Avoid resetting histogram, min & max unnecessarily

argilo:
Partially addresses #1282.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -364,8 +364,6 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
             {
                 emit pandapterRangeChanged(m_PandMindB, m_PandMaxdB);
 
-                m_MaxHoldValid = false;
-                m_MinHoldValid = false;
                 m_histIIRValid = false;
 
                 m_Yzero = py;
@@ -482,9 +480,6 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
                                               m_ClickResolution );
                 emit newDemodFreq(m_DemodCenterFreq,
                                   m_DemodCenterFreq - m_CenterFreq);
-                m_MaxHoldValid = false;
-                m_MinHoldValid = false;
-                m_histIIRValid = false;
                 updateOverlay();
             }
             else
@@ -955,8 +950,6 @@ void CPlotter::wheelEvent(QWheelEvent * event)
         if (m_PandMindB < FFT_MIN_DB)
             m_PandMindB = FFT_MIN_DB;
 
-        m_MaxHoldValid = false;
-        m_MinHoldValid = false;
         m_histIIRValid = false;
 
         emit pandapterRangeChanged(m_PandMindB, m_PandMaxdB);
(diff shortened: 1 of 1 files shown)


src/qtgui/plotter.cpp
```diff
@@ -364,8 +364,6 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
             {
                 emit pandapterRangeChanged(m_PandMindB, m_PandMaxdB);
 
-                m_MaxHoldValid = false;
-                m_MinHoldValid = false;
```
argilo, line 368: No need to reset max/min hold data when dragging the Y axis.

src/qtgui/plotter.cpp
```diff
@@ -482,9 +480,6 @@ void CPlotter::mouseMoveEvent(QMouseEvent* event)
                                               m_ClickResolution );
                 emit newDemodFreq(m_DemodCenterFreq,
                                   m_DemodCenterFreq - m_CenterFreq);
-                m_MaxHoldValid = false;
-                m_MinHoldValid = false;
-                m_histIIRValid = false;
```
argilo, line 487: No need to reset anything when adjusting the demod frequency.

src/qtgui/plotter.cpp
```diff
@@ -955,8 +950,6 @@ void CPlotter::wheelEvent(QWheelEvent * event)
         if (m_PandMindB < FFT_MIN_DB)
             m_PandMindB = FFT_MIN_DB;
 
-        m_MaxHoldValid = false;
-        m_MinHoldValid = false;
```
argilo, line 959: No need to reset max/min when zooming the Y axis.

src/qtgui/plotter.cpp
```diff
@@ -1187,7 +1180,7 @@ void CPlotter::draw(bool newData)
         && tnow_ms >= tlast_plot_drawn_ms + PLOTTER_UPDATE_LIMIT_MS);
 
     // Do not waste time with histogram calculations unless in this mode.
-    const bool doHistogram = (plotterVisible && m_PlotMode == PLOT_MODE_HISTOGRAM);
+    const bool doHistogram = (plotterVisible && m_PlotMode == PLOT_MODE_HISTOGRAM && newData);
```
argilo, line 1183: Processing the histogram after an overlay update causes the IIR data to be blown away.

src/qtgui/plotter.cpp
```diff
@@ -1900,8 +1893,6 @@ void CPlotter::setPandapterRange(float min, float max)
 
     m_PandMindB = min;
     m_PandMaxdB = max;
-    m_MaxHoldValid = false;
-    m_MinHoldValid = false;
```
argilo, line 1904: No need to reset max/min when dragging the Plot dB / WF dB sliders.

src/qtgui/plotter.cpp
willcode: At low frame rates, this make the histogram chase the avg line while panning. I think that's ok, since it also cuts down on processing, since the histogram isn't constantly redrawn during panning. Either way is fine with me.

willcode (approved):
Don't see any bad effects from removing those invalidation.

src/qtgui/plotter.cpp
argilo: Oops, that was unintended. I was trying to prevent the IIR from being reset when the demod frequency was changed. I'll remove this change and have a look to see whether there's a better way to accomplish that.

argilo: Actually, it's not that the IIR is reset, just that one IIR averaging step executes each time `updateOverlay` is called, so the IIR rapidly decays even though no new data has arrived.

argilo: There are some other events that also trigger `updateOverlay` and cause IIR decay:

* filter cutoff adjustment
* changing demod type
* setting A/B markers

argilo: Changing the condition to `(!m_histIIRValid || newData)` seems to do the trick.

argilo: Fixed.

PR #1291, opened 2023-10-01
Fix drawing of black areas in waterfall

argilo:
Partially addresses #1282.

An extra black pixel is drawn along the left edge of the waterfall because the code that draws the black areas to the left and right of the waterfall uses `drawRect` which draws the outline of a rectangle, which extends down and to the right by one pixel beyond the rectangle's dimensions.

I've changed the code to use `drawLine` instead, which draws a single line of pixels.


willcode (approved):
Gets rid of the funny pixel, and I don't see any other side effects.

PR #1292, opened 2023-10-01
Fix the pixels/bin calculation

argilo:
Partially addresses #1282.

At many zoom levels, there's an empty pixel at the right hand side of the plot & waterfall displays. As far as I can tell, this happens because the `xScale` variable, which represents the number of pixels per bin, uses the wrong figure (`w-1`) for the width of the display. (The display is actually `w` pixels wide.) As a result, the subsequent `xmin` and `xmax` calculations try to map the bins onto a `w-1` pixel display, and the rightmost pixel is often left empty (except when rounding error happens to push it one pixel larger). Changing the number of pixels to `w` fixes the problem.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1154,9 +1154,8 @@ void CPlotter::draw(bool newData)
     const double startFreq = fftCenter - span / 2.0;
     const double binsPerHz = fftSize / sampleFreq;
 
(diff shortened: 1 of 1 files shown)


willcode (approved):
The comment above that line (second sentance) can go too, then.

argilo:
Done.

willcode:
I think I'd compensated for +/- 1 in a bunch of places. Hopefully, you've simplified it all now.

PR #1296, opened 2023-10-02
Remove clamping from Y-coordinate calculations

argilo:
The plotter clamps the Y coordinate of many things:

* max line
* average line
* max hold line
* min hold line
* peaks

This is not visible at the bottom of the plot, but at the top of the plot, lines and peaks are drawn across the top of the window:

Screenshot from 2023-10-02 16-26-18

This seems undesirable and wastes CPU time. I think the clamping can be safely removed. After:

Screenshot from 2023-10-02 16-28-09

The top-bin highlighting in the histogram plot suffers from a similar problem, but it's not so easy to fix so I'll leave that for later.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1543,12 +1543,8 @@ void CPlotter::draw(bool newData)
         {
             const int ix = i + xmin;
             const qreal ixPlot = (qreal)ix;
-            const qreal yMaxD = (qreal)std::max(std::min(
-                panddBGainFactor * (m_PandMaxdB - 10.0f * log10f(m_fftMaxBuf[ix])),
-                (float)plotHeight), 0.0f);
-            const qreal yAvgD = (qreal)std::max(std::min(
-                panddBGainFactor * (m_PandMaxdB - 10.0f * log10f(m_fftAvgBuf[ix])),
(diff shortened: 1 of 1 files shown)


argilo:
@willcode Since you've done a lot of work on the plotter, I expect you might know if there is anything that depends on this clamping. So far I haven't noticed any negative effects.

willcode:
I don't think it will affect anything else. It was mainly to show that the signal was above scale. But then it should probably do the same thing at the bottom.

willcode:
The bottom clipping probably has a one-off somewhere.

argilo:
Yeah, I expect it would need to be `plotHeight - 1` to be visible.

Without clamping, it's still possible to tell that the lines have gone out of scale (since they disappear). I suppose with this change you can't tell there's an off-screen peak circle, but I don't think that's a big deal.

I'll let this sit for a bit to see if anyone else has an opinion on clamping vs not.

willcode:
Some indication of off-scale is nice. We could do it a different way. Nothing great comes to mind at the moment.

argilo:
I suppose an off-scale indication does become useful when the entire signal is off-screen.

argilo:
> The bottom clipping probably has a one-off somewhere.

I opened #1299 as an alternative, which fixes that bug.

PR #1298, opened 2023-10-03
plotter: clear waterfall using Delete key

### LINKED ISSUE #1297: Waterfall: add Clear button for waterfall
Thanks for the new release!

In the past, if the '**WF Span**' time was changed, the waterfall would be **cleared**.

As of v2.17, I see that the waterfall no longer clears, so you have a partial display of two different rates until the 'old' value scrolls off the bottom of the screen.

Is there any method remaining that results in a clean waterfall?   I've toggled several of the settings but not found one.

Thanks!

-Scott,  K4KDR

willcode:
The waterfall is not automatically cleared after changes to the many settings that could cause it to be distorted or invalid. An explicit clear allows users who do not want to see outdated data to get rid of it.

I did not add a button, as the FFT dock is already pretty full. We can add one if the shortcut is not sufficient.

Resolves #1297

### DIFF
--- a/resources/kbd-shortcuts.txt
+++ b/resources/kbd-shortcuts.txt
@@ -12,6 +12,7 @@ Main window actions:
  F11            Toggle full screen mode
  F              Set focus to the frequency controller
  Z              Zero frequency offset
+ Delete         Clear waterfall
  Ctrl+Q         Quit the program
 
 Receiver modes:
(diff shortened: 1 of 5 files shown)


0penBrain:
This PR did break the ability to delete a bookmark with Del key. Was it intentional?

willcode:
No, I didn't know about that feature. We can probably turn off the accelerator for the bookmarks widget.

PR #1299, opened 2023-10-03
Make out-of-range values visible at the bottom of the plot

argilo:
This is an alternative to #1296.

Instead of removing clamping, we could correct the clamping calculation so that out-of-range values are visible at the lower edge of the plot the same way as they are at the top.

Before:
Screenshot from 2023-10-03 09-31-20

After:
Screenshot from 2023-10-03 09-30-05


willcode:
This did had no effect for me, but replacing `1.0f` with `2.0f` did work. There may be some combination of `m_DPR` and rounding affecting this. I did not find the magic constant in a few tries.

argilo:
Oops, I guess I'll have to do a deeper dive into the calculations to see what's going on.

PR #1300, opened 2023-10-03
Refactor the signal meter

argilo:
The signal meter draws into an off-screen `QPixmap`, which has become unwieldy now that high-DPI screens must be supported. I've changed the code do directly draw the widget. Also, I removed a bunch of dead code and removed a lot of state variables which made the code unnecessarily complicated.

I made one visual change: there was code to draw shorter tick marks in between the numbers, but the length was set to the same value for both types of ticks. I've now made the minor ticks slightly shorter:

Screenshot from 2023-10-03 16-10-55

### DIFF
--- a/src/qtgui/meter.cpp
+++ b/src/qtgui/meter.cpp
@@ -35,11 +35,11 @@
 #include "meter.h"
 
 // ratio to total control width or height
-#define CTRL_MARGIN 0.07		// left/right margin
-#define CTRL_MAJOR_START 0.3	// top of major tic line
-#define CTRL_MINOR_START 0.3	// top of minor tic line
-#define CTRL_XAXIS_HEGHT 0.4	// vertical position of horizontal axis
-#define CTRL_NEEDLE_TOP 0.4		// vertical position of top of needle triangle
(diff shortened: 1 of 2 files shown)


src/qtgui/meter.cpp
```diff
@@ -83,167 +69,94 @@ QSize CMeter::sizeHint() const
...
 {
-    if (dbfs < MIN_DB)
-        dbfs = MIN_DB;
-    else if (dbfs > MAX_DB)
-        dbfs = MAX_DB;
-
-    float level = m_dBFS;
-    float alpha  = dbfs < level ? ALPHA_DECAY : ALPHA_RISE;
-    m_dBFS -= alpha * (level - dbfs);
-    m_Siglevel = (qreal)(level - MIN_DB) * m_pixperdb;
```
argilo, line 122: The meter should also be slightly more responsive now that it's drawing the current signal level instead of the previous one from 100ms ago. :smile:

willcode (approved):
Looks good with and without QT_SCALE_FACTOR.

argilo:
The numbers on the scale should also be better aligned now, since I fixed the calculation.

PR #1302, opened 2023-10-05
Refactor the frequency control

argilo:
Here I've made the following changes:

* Paint the frequency control directly, instead of painting into an off-screen QPixmap and copying that to the widget. (This avoids the need for device pixel ratio calculations.)
* Remove unused and redundant member variables.
* When processing mouse move events, only repaint the widget if the active digit changes.

These changes are not intended to have any change on the appearance or functionality of the frequency control.


willcode (approved):
Tested, works.

PR #1304, opened 2023-10-05
Use a QImage for the waterfall

argilo:
`QImage` is similar to `QPixmap` but allows direct access to pixel data. This is ideal for the waterfall, which consists entirely of pixels, allowing the `QPainter` to be removed.

This change results in a significant performance increase. CPU utilization for a full-screen waterfall-only display at 333 fps and 2048-point FFT drops from about 92% to 70% on my machine. I'm also able to achieve 500 fps at full-screen.

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -149,7 +149,7 @@ CPlotter::CPlotter(QWidget *parent) : QFrame(parent)
     m_DrawOverlay = true;
     m_2DPixmap = QPixmap();
     m_OverlayPixmap = QPixmap();
-    m_WaterfallPixmap = QPixmap();
+    m_WaterfallImage = QImage();
     m_Size = QSize(0,0);
     m_GrabPosition = 0;
     m_Percent2DScreen = 35;	//percent of screen used for 2D display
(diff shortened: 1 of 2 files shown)


src/qtgui/plotter.cpp
```diff
@@ -1168,7 +1159,7 @@ void CPlotter::draw(bool newData)
     const qint32 maxbin = std::min(endBin + 1, m_fftDataSize - 1);
 
     const qint32 xmin = qRound((double)(minbin - startBin) * xScale);
-    const qint32 xmax = qRound((double)(maxbin - startBin) * xScale);
+    const qint32 xmax = std::min(qRound((double)(maxbin - startBin) * xScale), qRound(w));
```
argilo, line 1162: We have to be more careful here to avoid drawing outside the bounds of the waterfall image.

willcode:
I also see a frame rate increase, all other things equal. Funny thing is that I avoided QImage because I thought it was supposed to be slower.

vladisslav2011:
I've tried switching to `QImage::Format_Indexed8` QImage for the waterfall to improve IQ file rendering performance and it was a bit slower than QPixmap. But it looks like `QImage::Format_RGB32` is a lot faster.

argilo:
I tried Indexed8 also, but I think then a lot of time is wasted looking up colors when painting the entire waterfall to the plotter widget.

PR #1182, opened 2023-01-16
Allow recording IQ samples in the SigMF format

Niautanor:
This implements part 1 of #988. Users can choose between the old format (raw) and SigMF with a chooser in the iq tool window.

Since SigMF keeps metadata and signal data separate, the actual recording remains unchanged regardless of which format is selected (except for the file name). For this reason, `.sigmf-data` files are made available for playback in the iq tool as well (this does not work in the general case (yet) but `.raw` isn't super well defined either).

One potential improvement would be to do some refactoring to separate formats into different classes that can be interacted with during the recording. This would 1. more cleanly separate the code paths for raw and sigmf recordings and 2. allow adding more metadata during the recording (e.g. we could start a new `capture` object when we tune the radio to a different frequency or change the gain).

PR #1305, opened 2023-10-06
Add core:hw to SigMF metadata files

argilo:
This is a follow-up to #1182.

As noted in https://github.com/gqrx-sdr/gqrx/pull/1182#issuecomment-1749787086, it would be useful to store input device metadata in the `core:hw` field, which the SigMF spec defines as:

> A text description of the hardware used to make the Recording.

Unfortunately, the gr-osmosdr API does not appear to provide any means to fetch metadata about the device associated with a source block. Metadata about all scanned devices can be retrieved with `osmosdr::device::find()`, but there is no guarantee that the device selected by the device string matches one of the scanned devices. So I think the best we can do for now is write out the current device string, e.g. `"core:hw": "airspyhf=0"`.

/cc @Niautanor

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -1656,6 +1656,7 @@ void MainWindow::startIqRecording(const QString& recdir, const QString& format)
                 {"core:sample_rate", sr/dec},
                 {"core:version", "1.0.0"},
                 {"core:recorder", "Gqrx " VERSION},
+                {"core:hw", QString("OsmoSDR: ") + m_settings->value("input/device", "").toString()},
             }}, {"captures", QJsonArray {
                 QJsonObject {
                     {"core:sample_start", 0},


willcode:
Too bad there's no way of declaring something purely informational. I can see someone parsing this string, since it means something to osmo. Maybe add a prefix? Something like
`osmo://airspyhf=0`
Not sure I really like that exact syntax though because we could potentially use a URI for something like ZMQ later. So,
`osmo: airspyhf=0` would be a less formal way of saying the same thing. Just some ideas.

argilo:
The spec sounds like it's meant to be human-readable. How about "gr-osmosdr device specification: airspyhf=0"?

willcode:
A little less verbose would be nice, in case it's used in a title bar or something, but yah, something that doesn't look parseable.

argilo:
> A little less verbose would be nice

OK, I changed the format to `OsmoSDR: airspyhf=0`.

willcode (approved):

PR #1306, opened 2023-10-06
Add feature to update existing bookmark

0penBrain:
This PR mainly adds a feature that an existing can be updated using contextual menu:
image

When updating, frequency + bandwidth + modulation are updated.
Bookmarks are re-sorted and selected bookmark is kept selected at its new location.

Also this PR changes that "Delete bookmark" (and "Update bookmark" as well) is only shown if contextual menu is called when hovering a bookmark. It's not more displayed when right-clicking on an empty area of the QTableView. Only "Add new bookmark" is available in this case.

### DIFF
--- a/src/qtgui/bookmarks.h
+++ b/src/qtgui/bookmarks.h
@@ -84,6 +84,15 @@ struct BookmarkInfo
     {
         return frequency < other.frequency;
     }
+
+    bool operator==(const BookmarkInfo &other) const
+    {
+        return frequency == other.frequency
+            and name == other.name
+            and modulation == other.modulation
+            and bandwidth == other.bandwidth
+            and tags == other.tags;
+    }
 /*
     void setTags(QString tagString);
     QString getTagString();
@@ -104,6 +113,7 @@ public:
     static Bookmarks& Get();
 
     void add(BookmarkInfo& info);
+    int update(int index, BookmarkInfo& info);
     void remove(int index);
     bool load();
     bool save();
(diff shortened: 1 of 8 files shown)


src/qtgui/bookmarks.h
```diff
@@ -84,6 +84,15 @@ struct BookmarkInfo
     {
         return frequency < other.frequency;
     }
+
+    bool operator==(const BookmarkInfo &other) const
```
willcode, line 88: Where is this used? Sort?

argilo:
I don't understand why this is needed. It is already possible to update a bookmark by clicking on the frequency, name, modulation, or bandwidth.

src/qtgui/bookmarks.h
0penBrain: As you edited. :smile: This is needed for `indexOf`.

0penBrain:
> I don't understand why this is needed. It is already possible to update a bookmark by clicking on the frequency, name, modulation, or bandwidth.

Actually it's possible to edit existing bookmarks, i.e. hand type new values in the fields.
Here the concept of "update" is that frequency+bandwidth+modulation are updated with the current GUI settings. :wink:

0penBrain:
I have checked that contextual `QMenu` is correctly deleted by Qt when hidden after being popped up.
Now consistent with `BookmarksTagList`.
Pushed as a second commit to ease review but can squash on demand.

argilo:
> Here the concept of "update" is that frequency+bandwidth+modulation are updated with the current GUI settings.

Ah, OK. What do you typically use this feature for?

0penBrain:
> Ah, OK. What do you typically use this feature for?

When exploring a frequency range with sparse emission. I can display the range at full scale and place coarse bookmarks where the waterfall is showing emissions. Then I can individually zoom in the frequency band and fine-tune each bookmark to more precise freq and bandwidth (and eventually fix the modulation if it was incorrectly set).

argilo:
Thanks for the explanation. It does sound useful to be able to update a bookmark that way.

vladisslav2011:
I have implemented something similar to this functionality in my fork: trying to create a bookmark on top of existing bookmark updates demodulator, filter bandwidth and other settings (most of demodulator/audio processing settings can be stored to a bookmark) instead of creating a new one.
Ah, It's part of this PR https://github.com/gqrx-sdr/gqrx/pull/1106

PR #1309, opened 2023-10-08
Fix double call to receiver::set_rf_freq, may fix #1129

### LINKED ISSUE #1129: Cannot switch between bookmarked channels.
Fedora 35
Repo installed Gqrx 2.15.9
Bookmark issue

I have multiple bookmarks, I cannot switch between them. If I'm on one bookmark, and click another to go to it, it looks like I tuned to it, however the waterfall stays the same, and I'm still on the previous station, it doesn't actually tune. However, if I click from one station to another, and then quit and restart Gqrx, hit play, I'm on the station I switched to, but I cannot switch to bookmarks I have by clicking them.

1. Listening to one bookmarked channel
2. Click another bookmarked channel
3. Interface shows I switched
4. Still on old channel
5. Exit software
6. Relaunch
7. Play
8. I'm on the station I tried to switch to before closing.

If I'm not on a bookmarked station, I can type and change the channel I'm on to another and not have to restart, the issue only exists if I'm on a bookmarked station and try to change.

0penBrain:
Also commented at conf loading because same thing arises there. I can't see any regression doing so but kept commented in case I missed something.


src/applications/gqrx/mainwindow.cpp
```diff
@@ -673,7 +673,7 @@ bool MainWindow::loadConfig(const QString& cfgfile, bool check_crash,
         }
 
         ui->freqCtrl->setFrequency(int64_val);
-        setNewFrequency(ui->freqCtrl->getFrequency()); // ensure all GUI and RF is updated
+        //setNewFrequency(ui->freqCtrl->getFrequency()); // ensure all GUI and RF is updated
```
argilo, line 676: Please don't comment out code; just delete it. Revision control already keeps track of changes.

0penBrain: Sorry, was unsure how you handle this. :confused:

argilo:
> may fix #1129

Before merging a change intended to fix a bug, I'd want to understand the root cause of the bug in question, and have some confirmation that the bug is actually fixed.

I have my doubts that the duplicate `setFrequency` calls are actually causing #1129, because it wouldn't explain why it only happens when tuned to a bookmarked frequency. And even if duplicate `setFrequency` calls were triggering the issue, removing duplicate calls here wouldn't fix the root cause (which presumably would be in in rtl-sdr or gr-osmosdr). There are other ways of triggering rapid `setFrequency` calls, like dragging the mouse across the plotter.

0penBrain:
> Before merging a change intended to fix a bug, I'd want to understand the root cause of the bug in question, and have some confirmation that the bug is actually fixed.

I'm sorry I can't help because I'm not reproducing here. But at least guys reproducing can get this branch and test. :wink: 

Also notice that even if it doesn't fix the bug, it still is more proper implementation. :smile:

> I have my doubts that the duplicate `setFrequency` calls are actually causing #1129, because it wouldn't explain why it only happens when tuned to a bookmarked frequency.

I share this but at some point you have to try something. :smile: Also notice in the bug discussion, no one formally said "it happens only when I'm on a bookmark and activate another bookmark, but doesn't happen if I'm on an arbitrary frequency and activate a bookmark". :wink: 

> And even if duplicate `setFrequency` calls were triggering the issue, removing duplicate calls here wouldn't fix the root cause (which presumably would be in in rtl-sdr or gr-osmosdr). There are other ways of triggering rapid `setFrequency` calls, like dragging the mouse across the plotter.

The time frame of the dup call wouldn't be exactly the same, as when dragging the mouse it would happen on (at least) 2 different runs of Qt event loop. Here it happen in the same run (guess so, a bit unsure about Qt signal/slot).

argilo:
> it still is more proper implementation

Sure, if there are reasons other than #1129 for making this change, we can make it. I'll need to review to make sure there aren't any unintended side effects.

> you have to try something

Yeah, experiments to get to the bottom of #1129 are a good idea. Another thing to try might be intentionally stressing the RTL-SDR with very frequent `setFrequency` calls.

willcode:
Blocking the signal in `freqctrl::setFrequency()` to avoid the loop would be better, architecturally. There are probably multiple places where we depend on a signal being routed through a GUI element, but it's not terribly modular that way.

0penBrain:
> Blocking the signal in `freqctrl::setFrequency()` to avoid the loop would be better, architecturally.

Yes. Thought to that but this is far more changing and testing, because the signal is actually connected to 3 or 4 slots in MainApplication. Would deserve an architectural decision first.

The main problem IMHO is that 2 functions are calling for a (hardware) frequency change in the code (1 in main application and 1 in the frequency controller). There should be only one (I have no clue which one is better to keep). 😉

PR #1310, opened 2023-10-09
Allow the Qt version to be chosen explicitly

argilo:
Gentoo has requested that it be possible to choose Qt6 or Qt5 explicitly. Here I've added cmake options (`-DFORCE_QT6=ON`, `-DFORCE_QT5=ON`) to do this. If neither is specified, then the newest available version of Qt will be selected.

/cc @ZeroChaos-

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,10 +122,20 @@ else()
 endif()
 
 # 3rd Party Dependency Stuff
-find_package(Qt6 QUIET COMPONENTS Core Network Widgets Svg SvgWidgets)
-if(NOT Qt6_FOUND)
+option(FORCE_QT6 "Force Qt6 to be used" OFF)
+option(FORCE_QT5 "Force Qt5 to be used" OFF)
+
+if(FORCE_QT6)
+    find_package(Qt6 QUIET COMPONENTS Core Network Widgets Svg SvgWidgets)
+elseif(FORCE_QT5)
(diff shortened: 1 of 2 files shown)


ZeroChaos-:
the code seems to work perfectly.  I tested setting the flags individually (expected results), setting no flags (default automagic behavior), and setting both (forces qt6).

This all works for me, thanks so much for poking me to update, and then adding my request so quickly!

argilo:
FYI @martymac 

I hope this will make this patch unnecessary: https://cgit.freebsd.org/ports/tree/comms/gqrx/files/patch-CMakeLists.txt

martymac:
Thanks @argilo! I've just updated the FreeBSD port. It is now built against Qt6 as everything seems to work fine with it by now!

argilo:
@martymac I'm glad to hear that. If any other problems come up on FreeBSD, feel free to open a GitHub issue.

martymac:
@argilo Thanks!

argilo:
One small fix: 5de29c178bf67af8fa6ac311990cac9745c6546a

I forgot to change `QUIET` to `REQUIRED` in the `FORCE_QT6` case, so CMake does not fail immediately if Qt6 is missing. It does fail later on, although not with a user-friendly error message.

This doesn't seem important enough to cut a new release, so it can wait until the next scheduled release.

kgarrels:
The QT5 QT6 chooser also works very nicely on my setup:
- macos 12.7
- gnuradio from conda
- QT and all the rest from macports.

Thanks!

PR #1308, opened 2023-10-08
Consistent bookmark tags edition with single click

0penBrain:
This PR aims at bringing consistent edition of tags attached to bookmarks, using a single click on the "Tag" field of a selected bookmark.

Commit 1 implements this with a QItemDelegate and drops code made obsolete by the change.

Commits 2 & 3 improves behavior in specific situations that wasn't introduced by this PR.

Next commits are some code fixes and cleanup.

PR #1312, opened 2023-10-12
bookmarks: restore Delete key

willcode:
Delete is used as a global shortcut to clear the waterfall. It is also used to delete a bookmark when the bookmarks table has focus. Allow the bookmarks table to override the global Delete shortcut.

Delete is used by other widgets, e.g., text fields, and is handled there in a similar way by the Qt library.

Note - the bookmarks table holds focus until something else acquires it, so a second Delete will offer to delete the next bookmark. There is a confirmation step.

Fixes issue mentioned in #1298

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -1,4 +1,9 @@
 
+    2.17.3: In progress...
+
+     FIXED: Delete key shortcut for bookmark removal.
+
+
     2.17.2: Released October 9, 2023
 
        NEW: FORCE_QT6 and FORCE_QT5 CMake options to force Qt version.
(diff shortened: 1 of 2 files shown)


0penBrain:
Should work (didn't test actually). A simpler solution may be to change the context of the global shortcut so it is limited to the plotter, but maybe this isn't what you want (eg. in this case it won't work when the focus is on the settings).

Not for myself : if this PR is merged, last commit of #1308 shall be removed (and probably some conflicts be managed).

willcode:
I think the Delete should work wherever it's not needed for something else. There could be a better way of doing this, but I couldn't think of one.

0penBrain:
> I think the Delete should work wherever it's not needed for something else. There could be a better way of doing this, but I couldn't think of one.

If this is what you're wanting, I think this is the right way.

willcode:
Another option is to use CTRL-Delete or some other key to clear the waterfall. Delete seemed more obvious than some random key.

0penBrain:
'C' as Clear can work but maybe already in use.

willcode:
That CW/L or CW/H with SHIFT. After a while you end up with so many shortcuts that nobody can remember them all.

argilo (approved):
The code changes look fine and work correctly. I'll merge once CI is happy.

PR #1319, opened 2023-11-01
cmake: add support for PipeWire audio backend

liambeguin:
add PipeWire support to build from source on Fedora F37+


argilo:
I don't understand this change. Gqrx does not have a PipeWire backend, and modifying the CMake won't change that.

argilo:
What problem are you trying to solve? Are you having trouble building Gqrx on Fedora? If so, what's going wrong?

liambeguin:
Hi @argilo,

My apologies for the lack of context.
I was trying to build `gqrx` on Fedora which was failing with the following error:
```
CMake Error at /usr/lib64/cmake/gnuradio/gnuradio-audioTargets.cmake:65 (set_target_properties):
  The link interface of target "gnuradio::gnuradio-audio" contains:

    JACK::JACK

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/lib64/cmake/gnuradio/gnuradio-audioConfig.cmake:14 (include)
  /usr/lib64/cmake/gnuradio/GnuradioConfig.cmake:105 (include)
  CMakeLists.txt:143 (find_package)

```

It turns out that Fedora uses PipeWire which has its own replacement for JACK::JACK. This sets the proper flags for PipeWire using `pkg_check_module()` . 

Liam

argilo:
Do you have the `jack-audio-connection-kit-devel` package installed?

liambeguin:
> Do you have the jack-audio-connection-kit-devel package installed?

I tried installing it, but it conflicts with `pipewire-jack-audio-connection-kit-devel`

argilo:
OK.

I suspect you've done something else wrong then. The Fedora project itself builds Gqrx packages without making any modifications to the CMake files: https://src.fedoraproject.org/rpms/gqrx/tree/rawhide

I'll close this off for now, since the changes here shouldn't be necessary.

argilo:
It might be helpful to have a look over the `BuildRequires` lines here to make sure you have all the build dependencies installed:

https://src.fedoraproject.org/rpms/gqrx/blob/rawhide/f/gqrx.spec

argilo:
If you're still not able to build without modifications, feel free to open an issue and we can try to find an appropriate solution.

liambeguin:
> It might be helpful to have a look over the `BuildRequires` lines

I was just looking at that, I'm trying to see where my `pipewire-jack-audio-connection-kit-devel` package comes from. 

I'll try removing that and will reach out if I still have issues.

Thanks for your time!

kukabu:
For building with pipewire-jack-audio-connection-kit-devel package you must run `cmake -DJACK_LIBRARY=/usr/lib64/pipewire-0.3/jack/libjack.so`
and I don't know why :(

PR #1321, opened 2023-11-09
Rename appdata file to metainfo

AsciiWolf:
Not a big deal, but the metainfo.xml suffix is preferred nowadays. See the [official AppStream documentation](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location).


AsciiWolf:
The macOS CI fail is unrelated.

AsciiWolf:
@argilo Feel free to ask if you have any questions. :-)

argilo:
I have considered this, but I didn't make the change because it would require many distribution maintainers to update their packaging. Do you think it's worth it?

AsciiWolf:
That's a very good point and you are probably right. We already install the file in a non-deprecated location and the `.metainfo.xml` suffix is more of a cosmetic thing since most AppStream parsers do not care much about the file suffix if it gets installed in a correct location. Thanks for your comment, I am closing this PR. :-)

argilo:
It may still be worth slipping this in the next time there is a major change that requires packaging updates anyway (perhaps the same time as a switch from gr-osmosdr to gr-soapy, if and when that happens).

PR #1323, opened 2023-11-16
Cmake : Fix build in MSYS2 MINGW64 environment

cozycactus:


argilo (approved):
These changes look reasonable, and don't seem to have any negative effects on the Linux and Mac builds.

PR #1325, opened 2023-11-26
Github Actions : Add Windows CI

cozycactus:


.github/workflows/ci.yml
```diff
@@ -93,3 +93,70 @@ jobs:
...
+          spdlog:p
+          volk:p
+          libsndfile:p
+          soapysdr:p
+          fftw:p
+          rtl-sdr:p
+    - name: Clone and Build GNU Radio
+      shell: msys2 {0}
+      run: |
+        git clone --recursive https://github.com/cozycactus/gnuradio.git
```
argilo, line 127: I think I'll hold off merging this PR until https://github.com/gnuradio/gnuradio/pull/6971 is merged. That will let us use the upstream GNU Radio.

.github/workflows/ci.yml
```diff
@@ -93,3 +93,70 @@ jobs:
...
+        cd gnuradio
+        git checkout fixmingw64
+        mkdir build && cd build
+        cmake -DCMAKE_INSTALL_PREFIX=C:/gqrx -G Ninja ..
+        ninja
+        ninja install
+    - name: Clone and Build gr-osmosdr
+      shell: msys2 {0}
+      run: |
+        git clone https://github.com/osmocom/gr-osmosdr.git
```
argilo, line 137: Better to use the official source instead of the GitHub mirror, as we do elsewhere in CI.
```suggestion
        git clone https://gitea.osmocom.org/sdr/gr-osmosdr.git
```

.github/workflows/ci.yml
```diff
@@ -93,3 +93,70 @@ jobs:
...
+          fftw:p
+          rtl-sdr:p
+    - name: Clone and Build GNU Radio
+      shell: msys2 {0}
+      run: |
+        git clone --recursive https://github.com/cozycactus/gnuradio.git
+        cd gnuradio
+        git checkout fixmingw64
+        mkdir build && cd build
+        cmake -DCMAKE_INSTALL_PREFIX=C:/gqrx -G Ninja ..
```
argilo, line 131: We could perhaps speed this up by only building the components that Gqrx needs:
```
            -DENABLE_DEFAULT=False \
            -DENABLE_GNURADIO_RUNTIME=True \
            -DENABLE_GR_ANALOG=True \
            -DENABLE_GR_AUDIO=True \
            -DENABLE_GR_BLOCKS=True \
            -DENABLE_GR_DIGITAL=True \
            -DENABLE_GR_FFT=True \
            -DENABLE_GR_FILTER=True \
            -DENABLE_GR_NETWORK=True \
```

.github/workflows/ci.yml
argilo: We should be able to switch this to upstream now.

I think `--recursive` can be removed as well, since GNU Radio doesn't have submodules anymore.

argilo:
Thanks for this. I made a few improvements and will merge once CI is happy.

PR #1326, opened 2023-12-01
Add Windows build

### LINKED ISSUE #1313: Why is there no Windows version?
Why is there no Windows version?  🤷‍♀️

argilo:
Fixes #1313.

Based on @cozycactus's work in #1325, I've added a Windows release to the build workflow. This is a prototype, and it has some limitations:

* The code is unsigned, so Microsoft Defender SmartScreen will likely block execution. To bypass it, click More info → Run anyway.
* Only RTL-SDR is supported so far. I don't expect it will be difficult to add other drivers, though.
* Even though the RTL-SDR driver is [available as an MSYS2 package](https://packages.msys2.org/base/mingw-w64-rtl-sdr), I had to build it from source because the packaged version crashes.
* ~~RTL-SDR Spectrum Server (rtl_tcp) is not supported because the gr-osmosdr module fails to build. It will need to be patched.~~

The build can be downloaded from the Artifacts section here: https://github.com/gqrx-sdr/gqrx/actions/runs/7055766990

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -269,3 +269,148 @@ jobs:
         with:
           name: gqrx-macos-${{ github.run_id }}
           path: Gqrx-*.dmg
+  windows:
+    runs-on: windows-latest
+    name: Windows
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - name: Setup MSYS2
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        update: true
+        install: >-
+          git
+          make
+        pacboy: >-
+          toolchain:p
+          cmake:p
+          ninja:p
+          libusb:p
+          boost:p
+          qt6-base:p
+          qt6-svg:p
+          spdlog:p
+          volk:p
+          libsndfile:p
+          soapysdr:p
+          fftw:p
+    - name: Clone and build GNU Radio
+      working-directory: ${{ runner.temp }}
+      run: |
+        git clone --branch v3.10.9.2 --depth 1 https://github.com/gnuradio/gnuradio.git
+        cd gnuradio
+        cmake -S . -B build \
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -268,3 +268,142 @@ jobs:
...
+          libSoapySDR.dll \
+          libspdlog.dll \
+          libstdc++-6.dll \
+          libusb-1.0.dll \
+          libvolk.dll \
+          libvorbis-0.dll \
+          libvorbisenc-2.dll \
+          libwinpthread-1.dll \
+          libzstd.dll \
+          zlib1.dll \
```
argilo, line 403: I had to put together this list of library dependencies manually. I wonder if there's a tool that could do this automatically (like dylibbundler on macOS).

cozycactus: can it be helpful? https://github.com/eyllanesc/69108420/blob/main/.github/workflows/test.yml

argilo: > can it be helpful?

What specifically were you looking at? It doesn't look like that project has any shared library dependencies like Gqrx does.

willcode:
OK, I'll do that now.

.github/workflows/build.yml
ryanvolz: Maybe this? https://github.com/mpreisler/mingw-bundledlls

cozycactus:
it works with rlt-sdr stick on my windows 11

kukabu:
I think the '--depth=1' should be added to 'git clone' for speed up.

argilo:
> I think the '--depth=1' should be added to 'git clone' for speed up.

Where? I've already done that for gnuradio, which is the only big repository involved.

kukabu:
> > I think the '--depth=1' should be added to 'git clone' for speed up.
> 
> Where? I've already done that for gnuradio, which is the only big repository involved.

rtl-sdr and gr-osmosdr

PR #1330, opened 2023-12-20
Discord Activity Support

imyoric:
Add Discord Activity Plz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,9 @@ set(${PROJECT_NAME}_MINOR "15")
 set(${PROJECT_NAME}_PATCH "10")
 set(IS_RELEASE FALSE)
 
+# Is Yoricya build?
+set(IS_YRCYA TRUE)
+
 if(IS_RELEASE)
     if(${PROJECT_NAME}_PATCH EQUAL 0)
         set(VERSION "${${PROJECT_NAME}_MAJOR}.${${PROJECT_NAME}_MINOR}")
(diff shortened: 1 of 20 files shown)


willcode:
Uh, probably not. But do you at least want to explain this in more detail?

imyoric:
> Uh, probably not. But do you at least want to explain this in more detail?

More information about Discord RPC? Or what do you mean?)

argilo:
> what do you mean?

Why do you want to add this feature? What is your use case?

imyoric:
Well, to be honest, this function is useless, but sometimes it’s nice to see that Discord displays information about an open program and even details

argilo:
OK, so I guess you did this just for fun?

In any case, thanks for sharing your work. But I don't think the code can be merged in as it is, because it breaks a bunch of things, includes libraries in binary form, and looks like it only works on macOS.

imyoric:
Oh, yep, for fun)
Can I try to improve it?

argilo:
Perhaps if you could share a video of this feature in action, it might help explain to others what it can be used for, and then we could see whether anybody else is interested.

imyoric:
Oh, okay I'll try to make a video. But here are some screenshots:

This is an example of Discord Rich Presence from the Internet and in English.
74b796acf11523f5f5a69ed00ae15f03355f727e

This is a finished prototype for Gqrx
Снимок экрана 2023-12-20 в 22 39 38
Translate: 
PLAYING A GAME
RDS: Empty (The signal is not stabl...
103.500MHz

imyoric:
I removed or changed some functions (For example, a lot of things were removed from the macos_bundle.sh file) so that the application could run in build mode. But this does not have to be included in the main repository

Mr-Precise:
I think the default should be English 
or add translation support...
And move "non-free components" into a separate git submodule.
It would be good if it were an independent fork (just my opinion).

imyoric:
> I think the default should be English or add translation support... And move "non-free components" into a separate git submodule. It would be good if it were an independent fork (just my opinion).

yep) the default language is English, and I have already finished working on the translation. But what do you mean by non-free components?) Discord Game SDK?
image

However, about the independent fork, I only partially agree with you, the fact is that I am new to Github and do not know how to resolve branch conflicts) that is why Gqrx will not be updated in my branch, and that is why I decided to make a pull request

I already asked a friend to make a more detailed video and PDF file with this feature

AsciiWolf:
@imyoric You can use TCP remote control to fetch these data instead (using hamlib) and write a simple proxy to send these data to Discord Rich Presence. Having this directly in Gqrx seems like a bad idea to me.

Also, the `discord_game_sdk.dylib` binary file should not be a part of Gqrx code.

argilo:
Thanks again for sharing your experiment. I'll close this pull request off for now, since the code is not in a state that could be merged.

PR #1333, opened 2024-01-30
Added ccache speedup compilation

GermanAizek:


argilo:
Thanks. This seems like a useful addition.

PR #1335, opened 2024-02-06
Add get_powerstat/set_powerstat to remote control.

kukabu:


argilo:
Did you have a look at #1334?

Responding to `\get_powerstat` with `0` seems to be problematic for WSJT-X, so the proposal was to always respond with `1`.

Is there another program that you're trying to integrate with here?

kukabu:
Why should we add a workaround for a bug in WSJT-X?
When you run WSJT-X with poweroff radio (not GQRX) you must understand that this won't work.

It's old patch and I forgot why I wrote it. :(

argilo:
> Why should we add a workaround for a bug in WSJT-X?

I'm not sure it's a bug. It sounds like WSJT-X just has an annoying error dialog that appears when the rig is off.

It's also not entirely clear to me whether it makes sense to map rigctl's "power" to Gqrx's "DSP on/off". "Power" could just as well mean "Gqrx is running".

kukabu:
"Gqrx is running" is standby mode, "DSP on" is power on, imho.

argilo:
Thanks for the contribution.

I'd prefer not to tie hamlib's "power" function to Gqrx's "DSP on/off", so I'm closing this off in favour of #1334.

Issue #1173, opened 2022-11-12
[FEATURE REQUEST] Dumping all visible signal levels via remote control


grappas:
Hi. 
I'm modifying gqrx-scanner and one thing missing for me is depending on sequence set frequency (via telnet) -> get signal level (via telnet) - > compare it to squelch level (inside program). It would be much faster if I could have access to whole visible spectrum of frequencies I could get info for. It could be csv, json, or whatever format.

neural75:
I am the author of gqrx-scanner and I would like to know if this feature has been considered: the remote control would only need the levels plotted in the main graph window (not the ones from the FFT panadapter). I've read some of the sources but I am not sure where to start. Where is the displayed data stored? Perhaps the command answer can send a small header with the current number of bins and frequencies boundaries and then the last raw levels data acquired. Would it be possible?

argilo:
It would certainly be possible, but some details would need to be considered. For instance:

* There would potentially be a lot of data transferred, so for efficiency reasons it might be necessary to use a binary format.
* For a scanner I presume you'd want a more-or-less fixed resolution bandwidth, independent of the sample rate and user-configured FFT size. It might make sense to perform a separate FFT for this purpose. How wide would you want each FFT bin to be, approximately?

willcode:
When I finish https://github.com/gqrx-sdr/gqrx/pull/1199 (output data via zmq), the user will be able to turn out output of various data. In the current PR, that includes raw data or fft data at either the current frame rate or a user-selectable rate. Since the output frames are tagged with a type and metadata, additional data channels can be added. I'm also considering a control channel over zmq. That function would overlap with the remote control function, so it needs some more thought.

grappas:
> There would potentially be a lot of data transferred, so for efficiency reasons it might be necessary to use a binary format.
That's not a problem honestly.

neural75:
>  How wide would you want each FFT bin to be, approximately?

I am not sure, my first guess was to have the entire spectrum bw from the dsp in a snapshot. It would be also fine to have a variable bin width (depending on the current settings) as long as it is possible to retrieve/calculate the resolution and tune to a frequency. The scanner could use that as an acceleration, then it can pinpoint the exact frequency carrier tuning and searching like it does already (to maximize the carrier level). Something like a hint there is something on that range to scan.

However, I would say a 10Khz bin should be fine in the most cases ( this is my default to scan step to get the rough level value). So, on a 3Mhz bw it would only need 300 samples.
[...]

PR #1341, opened 2024-03-12
Remote snr map

grappas:
According to feature request: https://github.com/gqrx-sdr/gqrx/issues/1173

Feature is not done yet. Do not merge it yet.
I need to figure out how to populate `snr_map`. Here's the question - in which function have you implemented SNR diagram? Tried to figure it out on my own, but no success to this moment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,7 +210,7 @@ jobs:
           MATRIX_ARCH: ${{ matrix.arch }}
         run: |
           GQRX_VERSION=$(<build/version.txt)
-          mv Gqrx.dmg Gqrx-$MATRIX_ARCH-$GQRX_VERSION.dmg
+          mv Gqrx.dmg Gqrx-$GQRX_VERSION-$MATRIX_ARCH.dmg
       - name: Save artifact
         uses: actions/upload-artifact@v4
         with:


grappas:
Thx. I'll see into it.

grappas:
@argilo can you check it?

grappas:
?

src/applications/gqrx/remote_control.cpp
```diff
@@ -902,12 +908,46 @@ QString RemoteControl::cmd_lnb_lo(QStringList cmdlist)
...
+    return answer;
+}
+
+void RemoteControl::populate_map(const std::vector<float> map) {
+    std::vector<float> cache;
+    cache.reserve(map.size());
+
+    for (size_t i = 0 ; i < map.size(); i++) {
+        cache.push_back(20.f * log10f(map[i] / 200.f));
+    }
```
argilo, line 932: This loop reduces Gqrx's performance substantially when the FFT size and/or rate are high, even if the remote control is not used.

I think it will be necessary to defer the calculations until the time that `\dump_map` is requested.

grappas: Or ditch it and send over raw fft.

argilo: That would be an improvement, but I suspect the memory copy would still be a drag.

grappas: how about now?

grappas:
anytime now...

willcode:
Agree that the data should only be copied when requested.

grappas:
It's not copied anymore. It's referenced.

grappas:
I'm looking forward to merge it.

PR #1343, opened 2024-03-19
Allow remote control commands to be pipelined

### LINKED ISSUE #1342: Remote Control delayed responses
Problem and reproduction:
------------------------
When using the rigctl compatible remote control feature of gqrx 2.17.4 (NixOS) from a custom C client, responses start to be delayed by some integer number of *commands*. This delay increased with the number of commands sent.

I have reproduced the issue with netcat, looked at the traffic with tcpdump, and believe I can explain why this happens.

This is trivially reproducible by opening up `nc localhost 7356` and _copying and pasting_ the following text:
```
m
f
F 146520000
f
m
M USB
m
```
(Note that copying and pasting is critical - if you type it out, you won't see the issue. One set of input, two behaviors.)

You will notice that you only get one or maybe two response(s). Hitting enter will cause the commands that did not get a response to be processed. For the test string above, if you get two responses, you will have to send five more lines to process all the commands. Any new commands will also need five more commands to be processed, as shown here (I have manually added '>' characters to show gqrx replies for readability):
```
$ nc localhost 7356
m
f
F 146520000
f
m
M USB
m
>USB
>2700
>146520000

>RPRT 0

>146520000

>USB
>2700

>RPRT 0

>USB
>2700

m





>USB
>2700
```

The number of commands that are incorrectly 'buffered' will increase every time you send multiple commands in a short window. Note that this is not related to Nagle's, or whether commands arrive in single or multiple TCP packets, and indeed neither of those could cause this behavior. 

I don't know a thing about socket programming as it relates to QT specifically, but I believe it's related to a mismatch between the Signal API and GQRX's usage of it for line-oriented commands.

Here's what is happening:
**`RemoteControl::startRead()` is called each time readyRead fires.**

**It then reads _one line_ off the buffer.**

So what happens when readyRead is fired once, and the resulting read (if not line-oriented) *would have* contained multiple commands? 

This is my informed guess as to what's happening. Since we are only reading one line off the buffer, not all commands that have arrived with this readyRead signal are processed, only the first one. This means if we send 6 commands, one will be processed. When we get another line later, readyRead is fired again ... and we get the next of the lines that were remaining from the first batch. At no point do we process all incoming bytes, we are making the assumption that one readyRead means one command, and that doesn't necessarily follow.

I note that the [readyRead documentation](https://doc.qt.io/qt-6/qiodevice.html#readyRead) says:
> (do not emit it [readyRead] only because there's data still to be read in your buffers)

So we know that we shouldn't expect another signal to process more bytes if our handling of the first signal didn't consume them.

The fix might make good use of [`canReadLine()`](https://doc.qt.io/qt-6/qiodevice.html#canReadLine) together with readLine, or use [`bytesAvailable()`](https://doc.qt.io/qt-6/qiodevice.html#bytesAvailable) as a hint.

From looking at the user group, I found a couple more people who seem to have encountered this. They were not able to articulate it well at the time, but the symptoms they describe match mine.

I've struggled with getting the dependencies all together, sorry for not including a patch with the bug report!

argilo:
Fixes #1342.

Gqrx currently expects remote control clients to send a single command at a time, and process the response before sending another command. This behaviour could be confusing. I don't see a reason to disallow pipelining, so I've enabled it here.

@tarxvftech Let me know if this change solves the problem you were seeing.

### DIFF
--- a/resources/news.txt
+++ b/resources/news.txt
@@ -3,6 +3,7 @@
 
        NEW: PlutoSDR and LimeSDR support in AppImage release, via SoapySDR.
   IMPROVED: Updated GNU Radio, SDR driver, and Qt versions in AppImage release.
+     FIXED: Respond correctly to pipelined remote control commands.
    REMOVED: FreeSRP support in AppImage release.
 
 
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -193,80 +193,82 @@ void RemoteControl::acceptConnection()
  */
 void RemoteControl::startRead()
 {
-    char    buffer[1024] = {0};
-    int     bytes_read;
-    QString answer = "";
+    while (rc_socket->canReadLine()) {
+        char    buffer[1024] = {0};
+        int     bytes_read;
+        QString answer = "";
 
-    bytes_read = rc_socket->readLine(buffer, 1024);
-    if (bytes_read < 2)  // command + '\n'
-        return;
+        bytes_read = rc_socket->readLine(buffer, 1024);
+        if (bytes_read < 2)  // command + '\n'
+            continue;
 
 #if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
-    QStringList cmdlist = QString(buffer).trimmed().split(" ", QString::SkipEmptyParts);
+        QStringList cmdlist = QString(buffer).trimmed().split(" ", QString::SkipEmptyParts);
 #else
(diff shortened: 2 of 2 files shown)


argilo:
This code change may be easier to review with whitespace changes excluded: https://github.com/gqrx-sdr/gqrx/pull/1343/files?w=1

willcode (approved):

tarxvftech:
That sure seems to do the trick - I can't replicate the problem using nc or my custom client. Awesome! :grin:

argilo:
Perfect, thanks for confirming!

PR #1361, opened 2024-07-03
RDS: fix flags decoding

vladisslav2011:
Change d0...d3 order to d3...d0 to match the standard.
Change decoding of Mono/Stereo flag to match the standard.
Change decoding of stPTY flag to match the standard.
rds0
rds1


argilo:
Thanks for the fix, and sorry for the delay in reviewing it.

`parser_impl.cc` is copied from https://github.com/bastibl/gr-rds, so it would make sense to apply the fix there as well. (In case you don't have time to open a pull request there, let me know and I can do it.)

The mono/stereo issue was fixed in gr-rds here: https://github.com/bastibl/gr-rds/pull/66, but it looks like the `segment_address` handling is wrong.

src/dsp/rds/parser_impl.cc
```diff
@@ -103,16 +103,16 @@ void parser_impl::decode_type0(unsigned int *group, bool B) {
 	/* see page 41, table 9 of the standard */
 	switch (segment_address) {
 		case 0:
-			mono_stereo=decoder_control_bit;
+			static_pty=!decoder_control_bit;
```
argilo, line 106: I think it may make sense to leave this bit as is, and change the display (in dockrds.cpp) instead. That way there won't be an API change in the PMT message format, which could potentially break applications that depend on gr-rds.

```suggestion
			static_pty=decoder_control_bit;
```

vladisslav2011: Hmmm.
I've tried to preserve compatibility by keeping PMT message characters meaning unchanged.
This bit means "dynamic PTY" when set, but the corresponding PMT message character means "static PTY". That's why it is inverted here.

argilo: > the corresponding PMT message character means "static PTY"

I suppose that may have been the intention, since the variable is named `static_pty`. But it could also be that the choice of variable name was a mistake, and it should have been named `dynamic_pty`.

I think the less confusing choice would be to copy the DI bits into the PMT message verbatim (as they are currently) and rename the variable.

In any case, it's gr-rds's API so we should probably see what @bastibl would prefer to do upstream.

argilo:
The changes look good.

I have proposed the fix upstream in https://github.com/bastibl/gr-rds/pull/78 and will give @bastibl a bit of time to look them over before merging this.

argilo:
It's been a while and I haven't heard back on the upstream PR, so I'll merge this and we can re-synchronize with gr-rds if needed.

argilo:
The upstream PR was accepted. Thanks for having a look @bastibl!

PR #1373, opened 2024-09-16
Add remote API for RDS STATION and RADIOTEXT

Sasszem:
> Note: This code is based on changes by our club members HA7CSK and HA5DNC done in 2020. I recently found them in the files, updated to the latest version and reworked the API a bit.

This PR adds 2 new queries to the "get parameter" command `p`:
- ~~`RDS_STATION`~~ `RDS_PS_NAME` reports program service name (usually station text)
- `RDS_RADIOTEXT` reports the radio text (usually name of the played song)

Example:
```
F 105300000
U RDS 1
p RDS_PI
p RDS_PS_NAME
p RDS_RADIOTEXT
```

### DIFF
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -344,12 +344,15 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     connect(remote, SIGNAL(gainChanged(QString, double)), uiDockInputCtl, SLOT(setGain(QString,double)));
     connect(remote, SIGNAL(dspChanged(bool)), this, SLOT(on_actionDSP_triggered(bool)));
     connect(uiDockRDS, SIGNAL(rdsPI(QString)), remote, SLOT(rdsPI(QString)));
+    connect(uiDockRDS, SIGNAL(stationChanged(QString)), remote, SLOT(setRdsStation(QString)));
+    connect(uiDockRDS, SIGNAL(radiotextChanged(QString)), remote, SLOT(setRdsRadiotext(QString)));
 
     rds_timer = new QTimer(this);
     connect(rds_timer, SIGNAL(timeout()), this, SLOT(rdsTimeout()));
 
     // enable frequency tooltips on FFT plot
     ui->plotter->setTooltipsEnabled(true);
+
 
     // Create list of input devices. This must be done before the configuration is
     // restored because device probing might change the device configuration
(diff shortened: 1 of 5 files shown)


argilo:
Thanks. This looks like it could be useful. Out of curiosity, what have you used it for? An RDS scanner?

One issue I would anticipate: Gqrx's RDS decoder (a copy of gr-rds) reports partially-received program service name & radiotext. Thus there's a good chance that querying these values through the API will return incomplete data. Here in North America many broadcasters dynamically change the program service name to display scrolling text, and Gqrx will happily return a mixture of old and new data during changes.

Do you think it would be useful to modify gr-rds so that it only reports complete program service name & radiotext?

argilo:
> `RDS_STATION`

In the standard, this field is called "Program Service name" or "PS name". Maybe something like `RDS_PS_NAME` would be clearer.

vladisslav2011:
I've tested this PR and it works very well combined with my improved RDS decoder.
The only thing, that I have changed is the default value when no RDS is received - changed it to an empty string.

Sasszem:
> Thanks. This looks like it could be useful. Out of curiosity, what have you used it for? An RDS scanner?

We used to have a discord bot that can stream FM stations using GQRX, and it was used to report played song.

> One issue I would anticipate: Gqrx's RDS decoder (a copy of gr-rds) reports partially-received program service name & radiotext. Thus there's a good chance that querying these values through the API will return incomplete data. Here in North America many broadcasters dynamically change the program service name to display scrolling text, and Gqrx will happily return a mixture of old and new data during changes.

I found that also, some stations do it as well. I'm not sure how could it be fixed tough. This was the reason I wrote "usually name of the played song".

> Do you think it would be useful to modify gr-rds so that it only reports complete program service name & radiotext?

Again, I'm not sure how could it be done to work well, but worth a try, bud possibly in another PR.

Sasszem:
> In the standard, this field is called "Program Service name" or "PS name". Maybe something like `RDS_PS_NAME` would be clearer.

Renamed to that.

argilo:
I'm working on some improvements to gr-rds that should make this work better: https://github.com/bastibl/gr-rds/pull/86

argilo:
I've pulled the gr-rds improvements into Gqrx here: https://github.com/gqrx-sdr/gqrx/pull/1394

argilo:
I did some cleanup (d9727bd557ea308d0b65e90b45a6b54fc715f766) and manually merged this.

As vladisslav2011 had done, I changed the default value for PS name and RadioText to the empty string. I also removed the string trimming, since I expect the improvements in #1394 make it unnecessary. Downstream applications can apply their own string transformations if they so desire.

Please let me know the names of the authors of this code so I can add them to the readme.

PR #1377, opened 2024-09-25
Add muting to remote control

Sasszem:
Adds "MUTE" to both L and l commands. Setting it to any value other than 0 mutes the audio, querying it results in 1 if audio is muted, else 0.


kukabu:
Hello
The "MUTE" must be in U/u commands.
see https://hamlib.sourceforge.net/html/rigctl.1.html

Sasszem:
I'm not 100% convinced that's the case, but will change it so when I get
around to it.

Michael Tatarinov ***@***.***> ezt írta (időpont: 2024.
szept. 26., Cs 4:52):

Sasszem:
> Hello The "MUTE" must be in U/u commands. see https://hamlib.sourceforge.net/html/rigctl.1.html

Did this change.
Also found a problem with the uninitialized bool variable.

argilo:
This needed a rebase and a bit of cleanup, so I did that and manually merged. The changes are now on the master branch.

Issue #1104, opened 2022-02-09
Apple Silicon / M1 builds?


drwetter:
Are there any plans to supply builds for the new architecture? Even hombrew fails (_Bad CPU type in executable_).

Are there any obstacles since qt6 works on aarch64?

argilo:
I would like the macOS builds to have M1 support. I just don't have the expertise (or hardware) to work on this at the moment.

One obstacle is that the Gqrx build currently uses gnuradio from Homebrew, which does not support M1:

https://github.com/gqrx-sdr/gqrx/blob/f108a292e400ada937cad27e80748eca23da448f/.github/workflows/build.yml#L89

https://formulae.brew.sh/formula/gnuradio

One possibility would be to build Gqrx's dependencies from source, instead of installing them from Homebrew. This would slow down builds, but perhaps they could be sped up again using ccache.

If anyone has expertise in building software with M1 support, I'd appreciate help.

drwetter:
FYI (but probably doesn't help much): I tried yesterday to compile gnuradio via homebrew. It stopped when compiling volk as dependency (error when compiling volk_machine_neonv8.c. It did only state the clang command but not the actual error).

vielmetti:
The "volk doesn't build on M1" issue is https://github.com/gnuradio/volk/issues/428 upstream, which is said to be fixed by https://github.com/google/cpu_features/commit/69d39934e83be9a0133d2850f15f5debd5eba06f reported in https://github.com/google/cpu_features/issues/179 with a PR in https://github.com/google/cpu_features/pull/209 - which as much to say that it might be worth revisiting this issue if indeed the problems are resolved.

vielmetti:
I was just now able to install "volk" with Home-brew, getting this:

==> Pouring volk--2.5.1_1.arm64_monterey.bottle.tar.gz

and no errors.

vielmetti:
Yup, and GNU Radio now installs OK

==> Pouring gnuradio--3.9.3.0_4.arm64_monterey.bottle.tar.gz

dsegel:
I just installed v2.15.9 (I don't understand the version numbering) via homebrew (brew install gqrx) and it runs fine on my M1 MacBook Air.

`==> Downloading https://github.com/gqrx-sdr/gqrx/releases/download/v2.15.9/Gqrx-2.15.9.dmg
[...]

PR #1384, opened 2024-10-16
Bump macOS CI versions

yuzawa-san:
per https://github.com/orgs/Homebrew/discussions/5603

macOS 12 is no longer supported by homebrew. it will run but the fast installation from "bottles" (precompiled blobs) will no longer work. this causes the dependencies to be compiled from source which takes a very long time, hence the timeouts.

for this reason i'm moving to the later versions which are still supported and have the fast bottles.

~NOTE: the sdk baseline is not changing, so versions 12 and earlier should still work.~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ jobs:
           path: Gqrx-*.AppImage
   macos:
     name: MacOS
-    runs-on: macos-12
+    runs-on: macos-13
     steps:
       - name: Check for Secret availability
         id: secret-check
@@ -79,6 +79,8 @@ jobs:
           fi
       - name: Install dependencies
         run: |
+          # for https://github.com/actions/runner-images/issues/9272
+          sudo chown -R runner:admin /usr/local/
           brew update
           brew install --HEAD librtlsdr
           brew install airspy airspyhf boost dylibbundler gnuradio hackrf libbladerf libserialport portaudio pybind11 six uhd qt@6 || true
@@ -111,7 +113,7 @@ jobs:
           cd build
           cmake ..
           make -j4
-          make install
+          sudo make install
 
           cd /tmp
           cp /Library/Frameworks/iio.framework/iio /usr/local/lib/libiio.dylib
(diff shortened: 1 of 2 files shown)


yuzawa-san:
@argilo I added your changes and also made some permissions fixes to address macOS 14 runner quirks https://github.com/orgs/Homebrew/discussions/5603

both the CI and Build workflows appear to have passed:

https://github.com/yuzawa-san/gqrx/actions/runs/11365851271
https://github.com/yuzawa-san/gqrx/actions/runs/11365851267

BUT: i was incorrect about that baseline changing. i'm going to put this into draft until i can dig into that further.

argilo:
> i was incorrect about that baseline changing.

I think it's unavoidable as long as the build uses packages from Homebrew.

argilo:
I was holding off on updating to macOS 13 as long as I could to avoid increasing the baseline, but at this point there is no choice.

Perhaps getting packages from conda or building from source would solve the problem, but I don't know enough about building software on macOS to do that.

argilo:
Thanks for the changes.

It looks like I accidentally checked in an unrelated change on the `fix-win-build` branch. I squashed commits here and remove that extraneous change. I'll merge once CI passes.

yuzawa-san:
i am going to continue to dig into the baseline.

i'll also take a look at ARM builds as well since I do have experience on mac builds.

argilo:
Thanks.

It would be great to have a proper ARM build, as there is plenty of interest: https://github.com/gqrx-sdr/gqrx/issues/1104

I believe the macos-14 runner is ARM, so that could be used if needed.

Unfortunately I don't have any hardware myself to use for testing.

argilo:
I pushed one more fix to get code signing working. (For some reason base64's `--output` argument was not working on macOS 13.) CI & builds should be working on the master branch now.

argilo:
> Perhaps getting packages from conda

By the way, Gqrx's Linux AppImage build is based on conda packages. It appears that conda packages target macOS 10.13+, so they may be a good choice.

PR #1391, opened 2024-10-24
Optimize Plotter Fills

### LINKED ISSUE #1386: bf40e76 disrupts remote control on Mac build
Understand this was mostly a graphics update that doesn't really touch anything to do with remote control. Nevertheless, something is crushing remote control output. GQRX is acting on external commands but isn't responding reliably when queried. Correct operation is restored when bf40e76 is reverted via git.

yuzawa-san:
fixes #1386

something about the polygon fills is not super efficient. i would imagine there is some sort of flood filling happening which is overkill since we know each pixels state. for this reason, i used drawPixel. i also go left to right row by row since that has the best performance. i also made the fill below a fillRect which is efficient for bulk regions.

image

UPDATE:
used vertical lines instead

### DIFF
--- a/src/qtgui/plotter.cpp
+++ b/src/qtgui/plotter.cpp
@@ -1597,7 +1597,7 @@ void CPlotter::draw(bool newData)
             {
                 const QPointF point = m_PlotMode == PLOT_MODE_MAX ? m_maxLineBuf[i] : m_avgLineBuf[i];
                 const qreal yFill = point.y();
-                painter2.fillRect(QRectF(point.x(), yFill, 1.0, yFillMax - yFill), m_FftFillCol);
+                painter2.fillRect(QRectF(point.x() - 1.0, yFill, 1.0, yFillMax - yFill), m_FftFillCol);
             }
             painter2.fillRect(QRectF(xmin, yFillMax, npts, plotHeight - yFillMax), m_FftFillCol);
         }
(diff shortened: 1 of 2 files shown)


vladisslav2011:
On my machine this strategy is a lot slower, than drawing the fill with `fillRect` per point: ~20 ms per spectrogram @64k FFT vs ~9ms per spectrogram with `fillRect`. Unfilled spectrogram draws in ~7ms. I'm still using the old plotter (slightly patched to improve performance and add features) on linux though.
I'm getting ~100FPS without fill/75FPS with fill @65k FFT, 4k using `fillRect` strategy and 60FPS with fill using `drawLines` strategy.

yuzawa-san:
i played around with fillRect and it appears to work nicely, so i'm putting that back. the only optimization i left was to do that bulk fill for the region below the lowest part of the polyline.

the bad performance before i believe was related to QT_THREAD_PARALLEL_FILLS being used, but something about the solid colors (instead of the alpha blending) seems to be avoiding that code path and making things faster. https://github.com/qt/qtbase/blob/b2e0ff4bca5d81205c8425cdc528e5c2a2988172/src/gui/painting/qdrawhelper.cpp#L3967

cc @markjfine

if this does not work, i'll just revert fully to whatever it was before.

argilo:
The fill appears to be drawn one pixel too far to the right:

Screenshot from 2024-10-28 15-38-03

yuzawa-san:
@argilo i also checked 2.17.5 and found that it was actually also off by a pixel there too, but i have fixed this. once again points in polygons can be in the middle of pixels, but the rectangles cannot it appears.

2.17.5:
Screenshot 2024-10-28 at 5 31 04 PM

this branch:
Screenshot 2024-10-28 at 6 14 34 PM

argilo:
The changes look reasonable, and work well on my machine.

PR #1396, opened 2024-11-09
Support tag renaming

yuzawa-san:
based on the comment "The problem is that after the tag name is changed in GUI you can not find the right TagInfo because you dont know the old tag name." i found that if we used a dialog like we did in https://github.com/gqrx-sdr/gqrx/pull/1395 rather than the row/cell edit mechanism, then we would not have that problem and could do something similar to what was being done in BookmarksTagList::changeColor. i also needed to emit the TagListChanged from the bookmarks since i found that editing the color in all 3 places (tags list, new bookmark dialog, bookmarks tags dialog) would not consistently update the color in the other places.

this likely supersedes https://github.com/gqrx-sdr/gqrx/pull/1165


vladisslav2011:
It is still possible to edit the tag inline with inconsistent results...

My attempt to implement tag renaming is here https://github.com/gqrx-sdr/gqrx/pull/1398
I think, it still has some minor issues (it may be possible to create multiple "*new*" tags, that will collapse when assigned to a bookmark for example), so it is possible, that I'll update it soon.

yuzawa-san:
could you please elaborate? is the solution to drop the `Qt::ItemIsEditable`? all i see if that there is still some oddness maybe around creating a new tag and not attaching it to any bookmarks, but that predates this pr.  the solution would be to write the tags to the config regardless of their attachment to any bookmark.

the `*new*` tag behavior should be fixed in https://github.com/gqrx-sdr/gqrx/pull/1395

vladisslav2011:
> is the solution to drop the `Qt::ItemIsEditable`?

I have not tested it, but I think, it will prevent inline editing.

vladisslav2011:
Hmmm. It may be possible to get rid of modal dialogs in the bookmarks dock completely by adding a mode switcher (a combo box or a check box) to switch tag list mode between "Filter by tags"  and "Edit selected bookmark's tags".

PR #1398, opened 2024-11-13
Tags fixes

vladisslav2011:
Make it possible to edit tag name inline both in the "Bookmarks" dock and "Edit tags" popup dialog window.
Update tag list in the "Bookmarks" dock when tags are changed (added, deleted, renamed, changed color) in the "Edit tags" popup dialog window.
Refuse to create duplicate tags by renaming.
Refuse to edit "Untagged" tag name.
Fix small memory leaks, spotted while implementing this fixes.


argilo:
Sorry for the delay in looking at this.

In general, the user interface here looks good. A couple observations:

* The "Rename" function is also accessible from the "New bookmark" dialog, but doesn't seem to work properly there. It creates a new tag, and does not rename the tag for existing bookmarks.
* In the "Change Bookmark Tag" dialog (and the "New bookmark" dialog), double clicking on a tag to rename it also toggles the checkbox, which is a bit confusing. Perhaps the double-click action could be removed? (Unless there's some better way to separate the two actions.)

argilo:
After renaming a tag from within the Bookmarks dock, the following is printed at the console:

```
QAbstractItemView::closeEditor called with an editor that does not belong to this view
```

After creating a new tag from within the Bookmarks dock, the following is printed:

```
QAbstractItemView::closeEditor called with an editor that does not belong to this view
QAbstractItemView::commitData called with an editor that does not belong to this view
QAbstractItemView::closeEditor called with an editor that does not belong to this view
```

Is that indicative of a bug? I don't think I've seen these messages appear before.

vladisslav2011:
Thanks for a review.

The "rename" function in   a "New bookmark" dialog needs more work. I'll either fix it or remove it from a context menu and update the PR.

> * double clicking on a tag to rename

Interesting. I didn't expect double clicking to start tag renaming, so disabling this trigger (or just making it not change the checkbox state) may be a good choice. I'll check this and update the PR.

> ```QAbstractItemView::closeEditor called with an editor that does not belong to this view```

This warning means that the method is called on a detached editor. It should not harm anything.

> I don't think I've seen these messages appear before.

Maybe you were using Qt < 6.4.

I'll try to find a way to silence these warnings.

vladisslav2011:
I've fixed the "New bookmark" dialog issues.
The ```QAbstractItemView::closeEditor called with an editor that does not belong to this view``` warning in not reproducible on my machine, so I'll install ubuntu 24.04 and make some tests later. I think, it may be fixed by converting ```void DockBookmarks::updateTags()``` to a slot and calling it via ```Qt::QueuedConnection``` instead of default ```Qt::DirectConnection```...

argilo:
> Maybe you were using Qt < 6.4.

I've been using Qt 6.4 for quite a while. What I meant is that I haven't seen this error message prior to this PR. I'll see how things look with your updates.

vladisslav2011:
I think, the warning (and possible use after free) is fixed now. No warning under Ubuntu 24.04, but it was not tested under Windows/MacOS.
Should I squash commits into a single commit?
Or is it, maybe better, squash it into two commits: memory leak fixes and tag renaming feature?

PR #1404, opened 2025-01-19
have playback stop at eof instead of waterfall smear

mhassell:
I added logic to the IQ playback tool to stop playback when we hit eof to stop the smeared waterfall.

This is my first work on GQRX so I started simple.   Please let me know if there is more in the PR review process.

### DIFF
--- a/src/qtgui/iq_tool.cpp
+++ b/src/qtgui/iq_tool.cpp
@@ -303,6 +303,16 @@ void CIqTool::timeoutFunction(void)
             ui->slider->blockSignals(false);
             refreshTimeWidgets();
         }
+        else
+        {
+            emit stopPlayback();
+            ui->listWidget->setEnabled(true);
+            ui->recButton->setEnabled(true);
+            ui->slider->setValue(0);
+            ui->playButton->setChecked(false);
+            is_playing = false;
+            refreshTimeWidgets();
+        }
     }
     if (is_recording)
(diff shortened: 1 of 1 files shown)


argilo:
Thanks for the contribution.

I agree it's annoying that DSP processing doesn't stop when the end of the file is reached. One thing I wonder about though: the UI slider is free-running and not synchronized to the actual file playback. Is it possible that the slider might reach its maximum value before EOF is reached, which would cause playback to stop too early?

mhassell:
@argilo I did test the build with these changes and did confirm with IQ files of various lengths (6s - 5min) that playback stops at the end of the file.  I also tried sliding the slider back and forth while playing back the files and it did not break anything.  So I think that is good.

It looks like the failure in the pipeline was for the windows build on files I didn't change.  Is that a pre-existing problem?

argilo:
Yes, the problem with the Windows build was fixed on the master branch after you created your PR.

mhassell:
Okay. I tried merging master into my branch but it didn't catch anything. (shrug)

argilo:
Perhaps you didn't pull the master branch before you did that?

I usually do a rebase instead of merge to keep the commit history cleaner.

mhassell:
I'm an idiot and learned the minutiae of branches this morning.  Merged master a moment ago and pushed.

PR #1406, opened 2025-01-25
Add developer tag to AppStream metadata

AsciiWolf:
To make the metadata compatible with future AppStream standard version that uses the [developer tag](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) instead of developer_name, but keep the original one for compatibility purposes.

@argilo Feel free to review and merge. :-)


argilo:
Thanks!

PR #1408, opened 2025-02-07
cmake: Fix exporting gr-rds symbols

ryanvolz:
CMake adds a definition for `gnuradio_RDS_EXPORTS`, but the code was recently changed to use `gnuradio_rds_EXPORTS` (no capitalization of RDS) instead.

Fixes build with MSVC, where linking fails with missing symbols when `gnuradio_rds_EXPORTS` is not defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ if(MSVC)
     add_definitions(-DBOOST_ALL_DYN_LINK)
 
     #export gr-rds symbols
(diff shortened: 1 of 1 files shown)


argilo:
Thanks! I'll merge this once CI is happy.

argilo:
@ryanvolz FYI, this made it into the 2.17.7 release, so you should be able to remove this workaround: https://github.com/conda-forge/gqrx-feedstock/blob/main/recipe/0003-cmake-Fix-exporting-gr-rds-symbols.patch

PR #1412, opened 2025-03-04
Update appimage workflow conda dependencies

ryanvolz:


argilo:
Thanks. This fixes the build failure observed in https://github.com/gqrx-sdr/gqrx/actions/runs/13639309688/job/38125316791, and I verified that the resulting AppImage still runs on an older Linux distribution (Ubuntu 20.04).

PR #1415, opened 2025-03-11
Upd build script

AndreiCherniaev:
1) the main point: stop to use make because since CMake 3.15 (Ubuntu 20.04), Generator can be set in env CMAKE_GENERATOR and sometimes it is not the Unix Makefiles.
2) let's don't install software which is under development to system's /usr/local/share/applications So I install to $HOME's folder
3) Tested in Ubuntu 24

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ To compile gqrx from source you need the following dependencies:
     - Network
     - Widgets
     - Svg (runtime-only)
-- cmake version >= 3.2.0
+- cmake version >= 3.13.0
 
 Gqrx can be compiled from within Qt Creator or in a terminal:
 
@@ -117,12 +117,18 @@ For command line builds:
 <pre>
 $ git clone https://github.com/gqrx-sdr/gqrx.git gqrx.git
 $ cd gqrx.git
-$ mkdir build
-$ cd build
-$ cmake ..
-$ make
+gqrx.git$ mkdir build
+gqrx.git$ cmake -S . -B build
+gqrx.git$ cmake --build build --parallel
+gqrx.git$ cmake --install build --prefix artifacts/
+gqrx.git$ rm -rf build/
 </pre>
-On some systems, the default cmake release builds are "over-optimized" and
+1) Last step is optional, use if you need to free disk space.
+2) Try to run `artifacts/bin/gqrx`. If you get error "artifacts/bin/gqrx: error while loading shared libraries: libQt6SvgWidgets.so.6: cannot open shared object file: No such file or directory" try to set path to folder with Qt like:
+```
+gqrx.git$ export LD_LIBRARY_PATH="$HOME/Qt/6.8.0/gcc_64/lib" && artifacts/bin/gqrx
+```
+3) On some systems, the default cmake release builds are "over-optimized" and
 perform poorly. In that case try forcing -O2 using
 <pre>
 export CXXFLAGS=-O2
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -109,20 +109,26 @@ To compile gqrx from source you need the following dependencies:
     - Network
     - Widgets
     - Svg (runtime-only)
-- cmake version >= 3.2.0
+- cmake version >= 3.13.0
```
argilo, line 112: The main CMakeLists file only requires cmake >= 3.5. Have you increased this due to an issue in the CMakeLists files, or because of the build commands you've suggested below?

I do see that `--parallel` was only added in cmake 3.12.

README.md
```diff
@@ -109,20 +109,26 @@ To compile gqrx from source you need the following dependencies:
...
 $ git clone https://github.com/gqrx-sdr/gqrx.git gqrx.git
 $ cd gqrx.git
-$ mkdir build
-$ cd build
-$ cmake ..
-$ make
+gqrx.git$ mkdir build
+gqrx.git$ cmake -S . -B build
+gqrx.git$ cmake --build build --parallel
+gqrx.git$ cmake --install build --prefix artifacts/
```
argilo, line 123: Why install to `artifacts`? We should probably recommend installing to the system, adding `sudo`.

README.md
AndreiCherniaev: In according with [link](https://cmake.org/cmake/help/latest/manual/cmake.1.html),  "-B <path-to-build>" added in version 3.13.
P.S. And CMake 3.13 is very old software. For example [Qt project](https://github.com/qt/qtbase/blob/a2e60ebee3737548d1be14fdbb39b08c515ae602/CMakeLists.txt#L8) for now use 3.16

README.md
AndreiCherniaev: There are a lot of discussions why "don't use make install", for [example](https://www.reddit.com/r/linux/comments/16iizeh/why_sudo_make_install/). The main point: usually gqrx is working under some distro like Ubuntu, not on embedded systems. For embedded systems `sudo cmake --install build` is ok but for systems with packet manager (like apt) such code will break system packages...
As a compromise we can add to readme.md `sudo cmake --install build` way as optional and dangerous. Do you agree?

README.md
```diff
@@ -109,20 +109,26 @@ To compile gqrx from source you need the following dependencies:
...
+gqrx.git$ rm -rf build/
 </pre>
-On some systems, the default cmake release builds are "over-optimized" and
+1) Last step is optional, use if you need to free disk space.
+2) Try to run `artifacts/bin/gqrx`. If you get error "artifacts/bin/gqrx: error while loading shared libraries: libQt6SvgWidgets.so.6: cannot open shared object file: No such file or directory" try to set path to folder with Qt like:
+```
+gqrx.git$ export LD_LIBRARY_PATH="$HOME/Qt/6.8.0/gcc_64/lib" && artifacts/bin/gqrx
+```
```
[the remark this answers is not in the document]
AndreiCherniaev: For example there is Qt Creator and Qt from [official source](https://www.qt.io/download-open-source) but sometimes users want to have CLI way to do the same like using Qt Creator but without Qt Creator. Qt is build as shared library so Qt Creator make trickery manipulations with LD_LIBRARY_PATH to successfully start your application. I know that not not all users of gqrx are professional in shared library and linking so I decide to add such comment...

How LD_LIBRARY_PATH in Qt Creator works

argilo:
Thanks for the proposal, but I'll leave the instructions as they are for now. I did update the required CMake version in the readme to match the `cmake_minimum_required` in CMakeLists.txt.

PR #1417, opened 2025-03-23
Allow multiple connections in remote control

### LINKED ISSUE #429: Remote control: only the last connection working

Sasszem:
Fixes #429.

Modified the remote control server so multiple clients can remote control the program at the same time.


argilo:
This seems like a useful improvement, and it worked well when I tested it.

PR #1424, opened 2025-05-25
Move macOS builds to Conda

### LINKED ISSUE #1414: Recent macOS builds fail to run
Automated macOS builds currently fail to run, due to a signature validation error when loading libfftw3f.3.dylib. This library comes from Homebrew: https://formulae.brew.sh/formula/fftw

The problem began when the fftw bottle was updated to 3.3.10_2. `otool -l` shows that the library's SDK version is set to "n/a", but it needs to be set to 10.9 or later for the corresponding signature to be accepted.

Homebrew is tracking the issue here: https://github.com/Homebrew/homebrew-core/issues/209981

For now, I guess we'll have to wait until the issue is fixed and an updated bottle is published.

argilo:
Fixes #1414.

Recent Gqrx builds fail to start on macOS, due to a problem with Homebrew's fftw package. Building Gqrx using Homebrew packages has been troublesome, so I'd like to switch over to using Conda packages instead. A wide variety of SDR packages are available, which will eliminate the need to build dependencies from source. Gqrx's AppImage releases are already built using Conda packages, so we'll also have one fewer package source to deal with.

Many thanks to @kgarrels, who did the heavy lifting and prototyped a Conda-based build. I cleaned up the prototype, and it appears to be running well. I would appreciate help with testing various SDR models on both Intel- and ARM-based Macs.

Disk images are available here: https://github.com/gqrx-sdr/gqrx/actions/runs/15242711733

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,71 +85,28 @@ jobs:
           else
             echo "available=false" >> $GITHUB_OUTPUT;
           fi
-      - name: Install dependencies
-        run: |
-          uname -a
-          # for https://github.com/actions/runner-images/issues/9272
-          sudo chown -R runner:admin /usr/local/
-          brew update
-          brew install --HEAD librtlsdr
-          brew install airspy airspyhf boost dylibbundler gnuradio hackrf libbladerf libserialport portaudio pybind11 six soapyremote uhd qt@6 || true
-
-          cd /tmp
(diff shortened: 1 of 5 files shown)


argilo:
I tested this with HackRF, BladeRF, Airspy R2, PlutoSDR, and RTL-SDR. All are working well. I have also received positive feedback from a couple testers, so I'll merge this.

PR #1426, opened 2025-07-03
Add new remote command to adjust center frequency

JackZielke:
I sometimes use gqrx as a panadapter.  I want to sync the frequency of my ham radio with the center frequency of gqrx.  The F remote command to change frequency will just adjust the tuning line if the jump is small.  I do not want to wait until I get near the edge of the bandwidth before the center frequency changes.  I want it to stay centered.  I made a quick new command "C" that changes the center frequency every time.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -5,6 +5,8 @@ Supported commands:
     Get frequency [Hz]
  F <frequency>
     Set frequency [Hz]
+ C <frequency>
+    Set center frequency [Hz]
  m
     Get demodulator mode and passband
  M <mode> [passband]
--- a/src/applications/gqrx/remote_control.cpp
+++ b/src/applications/gqrx/remote_control.cpp
@@ -224,6 +224,8 @@ void RemoteControl::startRead()
             answer = cmd_get_freq();
         else if (cmd == "F")
             answer = cmd_set_freq(cmdlist);
+        else if (cmd == "C")
+            answer = cmd_set_center_freq(cmdlist);
         else if (cmd == "m")
             answer = cmd_get_mode();
         else if (cmd == "M")
@@ -357,6 +359,13 @@ void RemoteControl::setNewRemoteFreq(qint64 freq)
         emit newFrequency(freq);
     }
 
+    rc_freq = freq;
+}
+
+/*! \brief New remote center frequency received. */
+void RemoteControl::setNewRemoteCenterFreq(qint64 freq)
+{
+    emit newFrequency(freq);
     rc_freq = freq;
 }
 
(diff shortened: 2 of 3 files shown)


argilo:
Thank you for your contribution.

Unfortunately, `C` can't be used for this purpose, because Gqrx maintains compatibility with the rigctl protocol, and `C` is already assigned to "set CTCSS tone".

You may perhaps be interested in this related pull request: #1422.

JackZielke:
I appreciate your reply.  I did look at #1422, and it did not look like it would be suitable for my needs.  It looks like they are using an IF tap, so the hardware frequency does not match the receiver frequency.  Unless I am way off base, their setup will stay on the IF at all times, and they just want the display to change.  I am using a separate receiver that shares the antenna (DRV output on TS-590SG).  I need gqrx to change the receiving frequency to match the tranceiver's current frequency (which is working beautifully).

Looking at rigctl, it looks like the only letter left is K.  Would a symbol like ^ be fine?

argilo:
> Would a symbol like ^ be fine?

I would suggest using a long command name like `gqrx_set_hw_freq`.

#1422 adds such a command already, which I suspect has the same function as your proposed `^` command.

JackZielke:
I downloaded and compiled #1422.  I was able to read the current frequency with gqrx_get_hw_freq but swapping F or ^ for gqrx_set_hw_freq did not change the frequency in gqrx.  I get a nice RPRT 1 response.  I may look into what is going on at some point, but probably not.

All I needed to do to get the F command to work as I needed was to comment out all the code that determines if the F command should really change the frequency or not.

```C
/*! \brief New remote frequency received. */
void RemoteControl::setNewRemoteFreq(qint64 freq)
{
    qint64 delta = freq - rc_freq;
    qint64 bwh_eff = 0.8f * (float)bw_half;

    rc_filter_offset += delta;
    if ((rc_filter_offset > 0 && rc_filter_offset + rc_passband_hi < bwh_eff) ||
        (rc_filter_offset < 0 && rc_filter_offset + rc_passband_lo > -bwh_eff))
    {
        // move filter offset
        emit newFilterOffset(rc_filter_offset);
    }
    else
    {
        // moving filter offset would push it too close to or beyond the edge
        // move it close to the center and adjust hardware freq
        if (rc_filter_offset < 0)
            rc_filter_offset = -0.2f * bwh_eff;
        else
            rc_filter_offset = 0.2f * bwh_eff;
        emit newFilterOffset(rc_filter_offset);
        emit newFrequency(freq);
    }

    rc_freq = freq;
}
```
Replacing all of that with:
```C
/*! \brief New remote frequency received. */
void RemoteControl::setNewRemoteFreq(qint64 freq)
{
    emit newFrequency(freq);
    rc_freq = freq;
}
```
And boom, the frequency changes when it receives the change frequency command.

Not realizing the rigctl compatibility constraint, I chose what looked like an unused letter to prevent a collision and to leave the original functionality alone.  Since that fixes everything, honestly, my main interest in looking at #1422 was to see if clicking in the UI would set the frequency or the offset.  If it would cause a frequency change without an offset change, then my rig would as well.  Alas, it did not.  It changed frequency and added an offset that has to be manually cleared for the interface to match again.  That is not something I ever plan to add, but I would have jumped ship for.

PR #1438, opened 2025-09-26
Add ZMQ IQ Sink for sending IQ data externally

mhassell:
I have added code and GUI features to allow a ZMQ sink block from GNURadio to be easily set up.  This will stream raw IQ data from the device to other sinks, eg GRC or other processes with digital demodulation capabilities.

I have tested this by opening a ZMQ socket in GRC and plotting the waterfall from the IQ data fed out by GQRX.  I have also opened ZMQ sockets in C and confirmed that nontrivial data is coming through the socket.

### DIFF
--- a/src/applications/gqrx/receiver.h
+++ b/src/applications/gqrx/receiver.h
@@ -294,9 +294,6 @@ private:
     sniffer_f_sptr    sniffer;    /*!< Sample sniffer for data decoders. */
     resampler_ff_sptr sniffer_rr; /*!< Sniffer resampler. */
 
-    
-    
-
 #ifdef WITH_PULSEAUDIO
     pa_sink_sptr              audio_snk;  /*!< Pulse audio sink. */
 #elif WITH_PORTAUDIO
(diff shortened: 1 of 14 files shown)


src/applications/gqrx/receiver.h
```diff
@@ -29,6 +29,8 @@
 #include <gnuradio/blocks/wavfile_sink.h>
 #include <gnuradio/blocks/wavfile_source.h>
 #include <gnuradio/top_block.h>
+#include <gnuradio/zeromq/pub_sink.h>
+#include <gnuradio/network/tcp_sink.h>
```
alexf91, line 33: Including `tcp_sink.h` is not necessary anymore.

alexf91:
Works well, but I wouldn't hide the options and the streaming button in the audio settings.

The I/Q recording settings might be more suitable.

mhassell:
Will work on this over the next few weeks and update the MR

alexf91:
Maybe move it to the I/Q tool and add a tab for streaming.

PR #1443, opened 2025-12-04
Convert IQ recorder from dialog to dockable widget

gopher2:
The IQ recorder dialog was often hidden behind the main window, requiring clicking around to find it during recording sessions. Converting it to a dockable widget keeps it visible and accessible alongside other controls. For those who prefer the old behavior, it can simply be dragged out to float as a separate window.

Removed the IQ recorder from Tools menu since it's now accessible via View menu.

Changed default recording format to SigMF, seems more modern format with metadata support.

Also added warning when config file is not writable. Spent way too long trying to figure out why my changes would never save. Turns out permissions on the file were wrong.

Only tested on macOS. This is my first pull request, so apologies if anything isn't quite right.

Dave

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,13 @@ This fork contains several feature branches with enhancements not yet in upstrea
 
 ### feature/combined (Recommended)
 All features merged into one branch. Use this for the complete enhanced experience.
+- Includes fix: Bookmark category visibility now persists across sessions
 
 ### feature/enhanced-bandplan
 **Band Plan Manager** - A new dock widget for viewing and editing band plan entries.
 - Editable table with columns: Show, Name, Min/Max Freq, Modulation, Step, Color
 - Show/hide checkbox to toggle band visibility on FFT
+- Band visibility state persists across sessions
 - Vertical stacking for overlapping bands in FFT display
 - Auto-contrast text color (black on light backgrounds, white on dark)
 - Click band in FFT to highlight row in Band Plan Manager


src/applications/gqrx/mainwindow.cpp
```diff
@@ -161,6 +161,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent)
     uiDockFft->toggleViewAction()->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_F));
     uiDockAudio->toggleViewAction()->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_A));
     uiDockBookmarks->toggleViewAction()->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_B));
+    iq_tool->toggleViewAction()->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_I));
```
vladisslav2011: This is not required as the shortcut is already defined in the https://github.com/gqrx-sdr/gqrx/blob/79645d962a9c240bdd1e4f78b5780db86b2d069c/src/applications/gqrx/mainwindow.ui#L685
Assigning a shortcut for the second time result in non-working shortcut and Qt error "QAction::eventFilter: Ambiguous shortcut overload: Ctrl+I".

vladisslav2011:
>  But when the "IQ Tool" dock is floating and Gqrx is closed and relaunched, the "IQ Tool" dock is being created behind the main window. So, I'm getting the buggy behavior, mentioned in the commit description with this PR applied... I don't know if upstream is affected because upstream does not build on my machine anymore.

This is Qt issue. Gqrx built with a different Qt version behaves correctly.

PR #1444, opened 2026-01-03
metadata: Small fixes

ghost:
- Drop project group, not related to this project
- Add vcs-browser URL to point to the source code repository
- Move the homepage URL to the top


argilo:
Thanks for your contribution.

Unfortunately, validation with appstreamcli fails on Ubuntu versions older than 24.04, presumably because `vcs-browser` is a recent addition. Would it be advisable to have CI validate only on the most recent Ubuntu LTS?

PR #1446, opened 2026-02-21
New GQRX remote control commands for device management

dsward2:
Add new commands to GQRX Remote Control Protocol for device management:     
   \get_input_device_list
   \get_input_device
   \set_input_device
   \get_output_device_list
   \get_output_device
   \set_output_device

Edited a tooltip in the I/O Configuration window to add an example of the popular "rtl=0,direct_samp=2" setting.

Mitigated a crash that occured during debugging in plotter.h due to a negative value.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -66,6 +66,18 @@ Supported commands:
     Get VFO option status (only usable for hamlib compatibility)
  \dump_state
     Dump state (only usable for hamlib compatibility)
+ \get_input_device_list
+    Get the list of available input device names
+ \get_input_device
+    Get the current input device name
+ \set_input_device <device_name>
+    Set the current input device name
+ \get_output_device_list
+    Get the list of available output device names
+ \get_output_device
+    Get the current output device name
+ \set_output_device <device_name>
+    Set the current output device name
  \get_powerstat
     Get power status (only usable for hamlib compatibility)
  v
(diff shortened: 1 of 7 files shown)


dsward2:
Pushed `9c418d9` addressing two issues I hit consuming these commands from an external client:

1. **`\set_input_device` / `\set_output_device` rejected any name with a space.** They did `if (cmdlist.size() != 2) return RPRT 1; QString dev = cmdlist[1];`, but the command line is whitespace-split, so most Core Audio output names (`Built-in Output`, `BlackHole 2ch`, `Mac mini Speakers`) never got through. Now they re-join everything after the command word.

2. **The plain getters re-enumerated all SDR hardware.** `\get_input_device` / `\get_output_device` called `CIoConfig::getDeviceList()` just to then read `settings->value("input/device")` — that probe stalls for several seconds while a device is open, and the RC replies batch up behind it. They now read the setting directly.

Also dropped the throwaway `new CIoConfig(...)` in all four, added a null check on the `MainWindow` / `QSettings` pointer (was dereferenced unconditionally), and removed three unused `<osmosdr/*>` includes. `\get_input_device_list` is unchanged — it still needs `getDeviceList`.

Verified: `\get_input_device` / `\get_output_device` return immediately; `\set_output_device Mac mini Speakers` → `RPRT 0`, `\get_output_device` reads it back.

dsward2:
Rebased onto current `master` to clear the staleness — no conflicts. The branch now replays as two commits: the original device-management commit plus `0e3763e` (the getter/space-name fix described above). GitHub shows the PR as MERGEABLE. Ready for review.

Issue #1455, opened 2026-05-27
Remote protocol: separate controls for VFO and RX frequency


afiskon:
Hi,

I'm working on a Python script that tracks the beacon signal from QO-100 satellite and corrects LNB LO frequency accordingly (a.k.a. AFC / automatic frequency control feature, similar to the one in SDRConsole). In the process I discovered that when I change the frequency with `F` command Gqrx may choose to automatically shift the VFO (center) frequency. This is not what I want.

Having separate controls for VFO frequency and receiving frequency would be nice. Gqrx version is 2.17.7. I am using AppImage package and run it on Ubuntu Linux.

argilo:
This issue is discussed in #1442, and a proposed solution is in #1426. So I'll close this issue off as a duplicate, and I would suggest carrying on the discussion in one of those two places.

PR #1456, opened 2026-05-28
Remote protocol: separate controls for HW and RX frequencies

### LINKED ISSUE #1442: Setting frequency by the remote control protocol problem
Hello, everyone.  If I set the frequency with the mouse on the screen, then everything is fine. When I give the command F 14450000, the frequency is set, but with a offset relative to the center by 432 kHz. Why is it so and how to set the frequency without offset?
And why is the maximum sensitivity only in the 200kHz band? I use Sdrplay and in SDRUno for Windows, I see a band up to 10MHz
Image

afiskon:
The proposed patch adds low-level controls for VFO / hardware center frequency and receive (filter) frequencies. Unlike `f` / `F` commands the new commands don't have any smart logic. The client is free to implement what it needs. 

Example:

```
gqrx_get_hw_freq
10489750000
gqrx_get_rx_freq
0
gqrx_set_rx_freq 10000
RPRT 0
gqrx_get_rx_freq
10000
gqrx_get_hw_freq
10489750000
gqrx_set_hw_freq 10489755000
RPRT 0
gqrx_get_rx_freq
10000
```

I used long command names as was previously proposed here: https://github.com/gqrx-sdr/gqrx/pull/1426#issuecomment-3041442593 Tested manually on Ubuntu 24.04 x64.

Closes #1442 / #1455 and probably other duplicates. Replaces PR #1426 which to me looks abandoned.

If there is something I should change / can improve please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ come with a Simplified BSD license.
 
 The following people and organisations have contributed to gqrx:
 
+* Aleksander Alekseev, R2AUK
 * Alex Grinkov
 * Alexander Fasching
 * Andrea Merello
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -62,6 +62,14 @@ Supported commands:
  LNB_LO [frequency]
     If frequency [Hz] is specified set the LNB LO frequency used for
     display. Otherwise print the current LNB LO frequency [Hz].
+ gqrx_get_hw_freq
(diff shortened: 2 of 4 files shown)


afiskon:
I published a blog post that describes my particular use case in much more detail https://eax.me/2026/2026-06-08-qo-100-gqrx-afc.html It's in Russian but Google Translate should be able to help with that.

PR #1458, opened 2026-06-19
FFT remote protocol: on-demand spectrum data command (fixes #1173)

### LINKED ISSUE #1173: [FEATURE REQUEST] Dumping all visible signal levels via remote control

neural75:
## Summary

Adds a new `FFT` remote-control verb (with `Q`/`V` sub-commands) that returns IQ spectrum data from the receiver's panadapter in one (or multiple) request, enabling fast FFT-based scanning without iterating through individual frequencies.

Fixes #1173 — provides the remote-control FFT data access to Gqrx.

## Key features

- `FFT V` — returns pooled dBFS values for the full visible spectrum
- `FFT Q` — query-only (metadata without values, for parameter discovery)
- `S:<start_hz>` / `C:<n>` — sub-range slicing (start frequency, bin count)
- `W:<target_bw_hz>` — on-demand bin pooling. Adjacent native FFT bins are power-averaged to produce wider output bins. Defaults to the channel filter bandwidth so the output matches the audio-path resolution, preventing very large responses on wideband receivers. Specify `W:<native_bw>` for raw native resolution.
- `R:<rate>` / `Z:<fftsize>` — returned with every response so the client can compute the exact native bin width as `R / Z` without an extra query.

## Why

Traditional frequency-stepping scan is slow (one "F <freq>\n" per hop + settle time + "l\n"). Reading the full FFT in one request gives a scanner instant access to all signals within the receiver's bandwidth. On-demand pooling (`W:`) avoids forcing every client to understand native FFT parameters while keeping the wire protocol compact.

## Files changed

| File | Change |
|---|---|
| `src/applications/gqrx/remote_control.cpp` | `cmd_get_fft()` implementation with full FFT read, sub-range slicing, bin pooling, and default-to-filter-bandwidth logic |
| `src/applications/gqrx/remote_control.h` | `d_rx` member, `setReceiver()`, `cmd_get_fft()` declaration |
| `src/applications/gqrx/mainwindow.cpp` | Wire up receiver access: `remote->setReceiver(rx)` |
| `resources/remote-control.txt` | FFT command documentation added |

## Protocol reference

See `remote-control.txt` for full protocol spec.

## Protocol examples

```
FFT Q
RPRT 0 F:27155000 R:5000000 Z:131072 S:24659997.3 E:29647256.2 B:9994.5 N:500 C:500

FFT Q W:1000
RPRT 0 F:27155000 R:5000000 Z:131072 S:24655495.9 E:29654275.2 B:991.8 N:5041 C:5041

FFT V C:5
RPRT 0 F:27155000 R:5000000 Z:131072 S:24659997.3 E:24699975.3 B:9994.5 N:500 C:5 -91.3 -92.0 -91.9 -92.4 -91.5

FFT V C:5 W:1000
RPRT 0 F:27155000 R:5000000 Z:131072 S:24655495.9 E:24659463.2 B:991.8 N:5041 C:5 -103.8 -102.8 -102.7 -100.7 -101.2

FFT V S:27155000 C:3
RPRT 0 F:27155000 R:5000000 Z:131072 S:27158624.0 E:27178613.0 B:9994.5 N:500 C:3 -59.7 -60.5 -61.0

FFT V S:27155000 C:3 W:1000
RPRT 0 F:27155000 R:5000000 Z:131072 S:27154885.6 E:27156869.2 B:991.8 N:5041 C:3 -70.2 -70.2 -70.2
```

## Example client for testing

Please, clone and compile https://github.com/neural75/gqrx-scanner/tree/feature/fft-scan
and run it with:
`./bin/gqrx-scanner -m fft -v -l 1000:3000 --vox`

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -82,10 +82,12 @@ Supported commands:
     Query full FFT parameters (metadata only, no values).
     Optional W: specifies target bin width for pooled output pool.
  FFT V [S:<start_hz>] [C:<n>] [W:<target_bw_hz>]
-    Get FFT spectrum data. Response format:
-     RPRT 0 F:<freq> S:<start> E:<end> B:<bw> N:<total> C:<count> [<dbfs_1> ...]
+    Get     FFT spectrum data. Response format:
+     RPRT 0 F:<freq> R:<rate> Z:<fftsize> S:<start> E:<end> B:<bw> N:<total> C:<count> [<dbfs_1> ...]
    Metadata fields:
      F    Center frequency the receiver is tuned to [Hz]
+     R    I/Q sample rate [Hz] (native_bw = R / Z)
+     Z    FFT size (number of native bins)
      S    Start frequency of the first returned bin [Hz]
      E    End frequency of the last returned bin [Hz]
      B    Bin width [Hz]
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -78,7 +78,32 @@ Supported commands:
     Set 'Split' mode (only usable for hamlib compatibility)
  _
     Get version
-
+ FFT Q [W:<target_bw_hz>]
+    Query full FFT parameters (metadata only, no values).
+    Optional W: specifies target bin width for pooled output pool.
+ FFT V [S:<start_hz>] [C:<n>] [W:<target_bw_hz>]
+    Get FFT spectrum data. Response format:
+     RPRT 0 F:<freq> S:<start> E:<end> B:<bw> N:<total> C:<count> [<dbfs_1> ...]
+   Metadata fields:
+     F    Center frequency the receiver is tuned to [Hz]
+     S    Start frequency of the first returned bin [Hz]
+     E    End frequency of the last returned bin [Hz]
+     B    Bin width [Hz]
+     N    Total number of bins in the full FFT
+     C    Number of bins returned in this response
+   Arguments:
+      S:<start_hz>      Start frequency for the returned bins [Hz]
+      C:<n>             Number of bins to return
+      W:<target_bw_hz>  Target bin width [Hz]. Adjacent native FFT
+                        bins are power-averaged to produce wider
+                        output bins.  Defaults to the channel filter
+                        bandwidth so the output matches the audio-path
+                        resolution.  Specify W:<native_bw> (from FFT Q)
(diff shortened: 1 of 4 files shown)


neural75:
@argilo Hi, may you review it please when you can so I can adapt the implementation on gqrx-scanner in case you have something you want to change? I was testing it and it seems ok at runtime, it does not change the cpu usage in any sensible way (my pc is old and slow), but I think it would be appropriate to have a second opinion. The protocol allows to transfer the spectrum dividing it in smaller chunks. I didn't use binary format to be compliant with the rest of the remote text protocol commands and also because the current approximation is enough (one decimal digit in dBFS), in term of size it is not that different of floating point numbers. 

The default is to return the bins size ~= to the receiver filter bandwidth so the number of values is not at the full resolution (sampling rate / fft size) and also because the returned dBFS is aligned with the squelch power level by default, so that the user can choose what is the given precision/transfer size trade-off. 
For example: gqrx-scanner implements this in multiple passes: one coarse pass at the filter width and one fine-tuning pass with an increased precision around the peaks for each candidate, to properly center the signals with +- 100hz precision at 10khz filter bw.

Also note: this PR is unrelated to the ZeroMQ PR that streams I/Q samples out of the receiver. This one follows an on-demand pull model — the client requests FFT data only when needed over the existing TCP remote control protocol, so the external client never has to demodulate raw signals or receive a continuous stream of unsolicited data. Even if that is possible, I think demodulation belongs in Gqrx's radio receiver responsibility to avoid duplicating code and processing across separate programs. This one simply exposes a snapshot of the computed FFT, giving the remote client a way to "see" what the receiver is tuned to — useful for satellite Doppler tracking / scanners and the like. It is not meant to connect an external de-modulator using Gqrx as a UI; the ZMQ PR already serves that purpose with the gnuradio zmq sub source.
There is also another draft PR more specific to FFT over ZMQ, but that also streams FFT data at native resolution using a push model where the client cannot control what portion of the spectrum to read or at what resolution -- this is intrinsic to the pub/sub pattern unless one adds far more complexity in the implementation of a control channel.

Thank you.

PR #1463, opened 2026-09-10
remote-control: add read/write FILTER_SHAPE level

dsward2:
Adds a `FILTER_SHAPE` level to the remote-control interface so a client can query and set the channel-filter transition band (Soft / Normal / Sharp) that today is only reachable from the Receiver Options dock.

## Commands

```
l FILTER_SHAPE            -> 0 | 1 | 2          (0 = soft, 1 = normal, 2 = sharp)
L FILTER_SHAPE <shape>    -> RPRT 0 / RPRT 1    (<shape> = 0/1/2 or SOFT/NORMAL/SHARP, case-insensitive)
l ? / L ?                 -> now also list FILTER_SHAPE
```

Exposed as a **level** rather than an extra field on `m` / `M`, so hamlib/rigctld
clients (gpredict, WSJT-X, …) that parse `m` as exactly two lines are unaffected,
and so support is discoverable the usual way — look for `FILTER_SHAPE` in `l ?`.

## Behaviour

- The shape is receiver-wide and preserved across demodulator changes, exactly
  like the GUI selector.
- Setting it routes through `MainWindow::selectDemod()` — the same path the
  Receiver Options combo uses — so the running filter is updated immediately and
  remote / GUI state stay in sync both ways:
  - RC -> GUI: `RemoteControl::newFilterShape(int)` -> `MainWindow::setFilterShape(int)`
  - GUI -> RC: `MainWindow::selectDemod()` -> `RemoteControl::setFilterShape(int)`
- Out-of-range / unparseable values return `RPRT 1` and change nothing.

## Changes

| File | |
|---|---|
| `remote_control.h/.cpp` | `rc_filter_shape` (init NORMAL), `setFilterShape(int)` slot, `newFilterShape(int)` signal, `FILTER_SHAPE` in `cmd_get_level`/`cmd_set_level` + both `?` lists |
| `mainwindow.h/.cpp` | `setFilterShape(int)` slot (clamps, `setCurrentFilterShape`, re-runs `selectDemod`); one `connect()`; one sync line in `selectDemod()` |
| `resources/remote-control.txt` | documents the two commands |

No DSP/flowgraph changes — `receiver::set_filter()` already takes the shape.

## Testing

Builds clean on macOS (arm64, Qt 5.15.18, GNU Radio 3.8.5). Verify at runtime,
with **Tools ▸ Remote control** enabled:

```
$ printf 'l ?\nL FILTER_SHAPE 2\nl FILTER_SHAPE\nL FILTER_SHAPE normal\nL FILTER_SHAPE 5\n' | nc localhost 7356
# expect: list contains FILTER_SHAPE ; RPRT 0 ; 2 ; RPRT 0 ; RPRT 1
```

The Receiver Options "Shape" combo should follow each `L FILTER_SHAPE`, and a
manual combo change should be reflected by a later `l FILTER_SHAPE`.

Motivation: my AntennaHead / LocalRadio project drives gqrx over this interface
and wants filter shape alongside the frequency / mode / passband / gain controls
it already uses.

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -27,6 +27,12 @@ Supported commands:
     Get the value of the gain setting with the name <gain_name>
  L <gain_name>_GAIN <value>
     Set the value of the gain setting with the name <gain_name> to <value>
+ l FILTER_SHAPE
+    Get the channel filter shape (0 = soft, 1 = normal, 2 = sharp)
+ L FILTER_SHAPE <shape>
+    Set the channel filter shape. <shape> may be 0/1/2 or SOFT/NORMAL/SHARP
+    (case insensitive). The setting is receiver-wide and is kept across
+    demodulator changes, matching the Receiver Options filter shape selector.
  p RDS_PI
     Get the RDS PI code (in hexadecimal). Returns 0000 if not applicable.
  p RDS_PS_NAME
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -345,6 +345,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
     connect(remote, SIGNAL(stopIqRecorderEvent()), iq_tool, SLOT(stopIqRecorder()));
     connect(ui->plotter, SIGNAL(newFilterFreq(int, int)), remote, SLOT(setPassband(int, int)));
     connect(remote, SIGNAL(newPassband(int)), this, SLOT(setPassband(int)));
+    connect(remote, SIGNAL(newFilterShape(int)), this, SLOT(setFilterShape(int)));
     connect(remote, SIGNAL(gainChanged(QString, double)), uiDockInputCtl, SLOT(setGain(QString,double)));
     connect(remote, SIGNAL(dspChanged(bool)), this, SLOT(on_actionDSP_triggered(bool)));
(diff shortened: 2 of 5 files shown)


dsward2:
Runtime-tested this branch (merged onto current `master`) by driving the RC port directly. Build: macOS, Qt 5.15.18 / GNU Radio 3.8.5, run headless (`QT_QPA_PLATFORM=offscreen`, no SDR device — none of these commands need one).

```
_                      -> 'Gqrx v2.17.7-25-g6b578af-dirty'
l ?                    -> 'SQL STRENGTH AF FILTER_SHAPE'
l FILTER_SHAPE         -> '1'
L FILTER_SHAPE 2       -> 'RPRT 0'
l FILTER_SHAPE         -> '2'
L FILTER_SHAPE soft    -> 'RPRT 0'
l FILTER_SHAPE         -> '0'
L FILTER_SHAPE 5       -> 'RPRT 1'
L FILTER_SHAPE normal  -> 'RPRT 0'
l FILTER_SHAPE         -> '1'
m                      -> 'AM\n10000'
L ?                    -> 'SQL AF FILTER_SHAPE'
```

So: `FILTER_SHAPE` shows up in both `l ?` / `L ?`, ints and the `SOFT`/`NORMAL`/`SHARP` names both parse, out-of-range returns `RPRT 1`, and `m` is unchanged at two lines (no regression for hamlib clients).

dsward2:
Rebased onto current `master` — applies cleanly with no conflicts, GitHub shows the PR as MERGEABLE. Nothing in the diff changed; just confirming it's current and the runtime test above still holds. Ready for review whenever you have a chance.

PR #1464, opened 2026-09-10
remote-control: add bookmark download and recall commands

dsward2:
Lets a remote-control client read the bookmark list and jump to an entry —
useful for building an external UI on top of gqrx. (My AntennaHead / LocalRadio
project drives gqrx this way and already uses the frequency / mode / passband /
gain commands.)

## New commands

| Command | Reply |
|---|---|
| `\get_bookmarks` | count line `N`, then `N` lines `<freq_Hz>\|<name>\|<modulation>\|<bandwidth_Hz>\|<tag,tag,…>` |
| `\get_bookmarks_in_range <lo> <hi>` | same, limited to `lo <= frequency <= hi` [Hz]; `RPRT 1` on bad args |
| `\get_bookmark_tags` | count line, then one tag name per line |
| `\set_bookmark <index>` | `RPRT 0` — tune to the bookmark at 0-based `<index>` (in `\get_bookmarks` order) and apply its mode + bandwidth; `RPRT 1` if out of range |
| `\set_bookmark_freq <frequency>` | `RPRT 0` — activate the first bookmark whose frequency equals `<frequency>` [Hz]; order-independent; `RPRT 1` if no match |
| `\reload_bookmarks` | re-read `bookmarks.csv` from disk; `RPRT 0` / `RPRT 1` |

## Design notes

- **Framing.** `\get_bookmarks` uses a count line then one line per record, like
  `\dump_state`'s "lines" style. Fields are `|`-separated; `|`, CR and LF are
  replaced with a space in `name`, `modulation` and tag names so a record is
  always exactly one line (`bookmarks.csv` itself is unquoted `;`-separated, so a
  wire separator has to be chosen and sanitised rather than inherited).
- **Addressing.** `\set_bookmark <index>` is simple but can go stale if the list
  changes between calls, so `\set_bookmark_freq <Hz>` is provided as an
  order-independent alternative. Neither command takes a name argument — the
  command line is whitespace-split and bookmark names contain spaces.
- **Apply path.** `\set_bookmark*` emit a new `RemoteControl::newBookmarkActivated`
  signal wired to the existing `MainWindow::onBookmarkActivated` slot — the same
  slot the Bookmarks dock uses — so tuning, mode and bandwidth handling are
  identical to activating a bookmark in the GUI. No `MainWindow` dependency is
  added to `remote_control.cpp`; reads go through the `Bookmarks` singleton.
- **Read + apply only** — no `\add_bookmark` / `\remove_bookmark`.
- **Compatibility.** All additions are new `\`-verbs; no existing command's
  request or reply changes.

## Changes

| File | |
|---|---|
| `remote_control.h` | `newBookmarkActivated(qint64,QString,int)` signal; 6 `cmd_*` declarations |
| `remote_control.cpp` | `#include "qtgui/bookmarks.h"`; 6 dispatch entries; 6 handlers + two small static formatting helpers |
| `mainwindow.cpp` | one `connect()` next to the existing dock connection |
| `resources/remote-control.txt` | documents the six commands |

## Testing

Builds clean on macOS (arm64, Qt 5.15.18, GNU Radio 3.8.5). Verify at runtime
after saving a couple of bookmarks, with **Tools ▸ Remote control** enabled:

```
$ printf '\\get_bookmarks\n\\get_bookmark_tags\n\\set_bookmark 0\n\\set_bookmark_freq 95500000\n\\set_bookmark 99\n' | nc localhost 7356
# expect: count+rows ; count+tags ; RPRT 0 ; RPRT 0 ; RPRT 1   (receiver retunes on each RPRT 0)
```

### DIFF
--- a/resources/remote-control.txt
+++ b/resources/remote-control.txt
@@ -68,6 +68,25 @@ Supported commands:
     Dump state (only usable for hamlib compatibility)
  \get_powerstat
     Get power status (only usable for hamlib compatibility)
+ \get_bookmarks
+    Get all bookmarks. The first reply line is the bookmark count N, followed
+    by N lines, each formatted as:
+      <frequency_Hz>|<name>|<modulation>|<bandwidth_Hz>|<tags>
+    where <tags> is a comma separated list. The characters '|', CR and LF are
+    replaced with a space in <name>, <modulation> and tag names so every
+    bookmark is a single unambiguous line.
+ \get_bookmarks_in_range <lo> <hi>
+    Like \get_bookmarks but only bookmarks with lo <= frequency <= hi [Hz].
+ \get_bookmark_tags
+    Get all bookmark tag names: a count line followed by one name per line.
+ \set_bookmark <index>
+    Activate the bookmark at <index> (0-based, in \get_bookmarks order): tune
+    to it and apply its mode and bandwidth. Returns RPRT 1 if out of range.
+ \set_bookmark_freq <frequency>
+    Activate the first bookmark whose frequency equals <frequency> [Hz]. This
+    is independent of the list order. Returns RPRT 1 if there is no match.
+ \reload_bookmarks
+    Reload bookmarks.csv from disk, e.g. after editing it outside Gqrx.
  v
     Get 'VFO' (only usable for hamlib compatibility)
  V
--- a/src/applications/gqrx/mainwindow.cpp
+++ b/src/applications/gqrx/mainwindow.cpp
@@ -311,6 +311,7 @@ MainWindow::MainWindow(const QString& cfgfile, bool edit_conf, QWidget *parent) 
 
     // Bookmarks
     connect(uiDockBookmarks, SIGNAL(newBookmarkActivated(qint64, QString, int)), this, SLOT(onBookmarkActivated(qint64, QString, int)));
+    connect(remote, SIGNAL(newBookmarkActivated(qint64, QString, int)), this, SLOT(onBookmarkActivated(qint64, QString, int)));
     connect(uiDockBookmarks->actionAddBookmark, SIGNAL(triggered()), this, SLOT(on_actionAddBookmark_triggered()));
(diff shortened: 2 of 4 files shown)


dsward2:
Runtime-tested this branch (merged onto current `master`) by driving the RC port directly. Build: macOS, Qt 5.15.18 / GNU Radio 3.8.5, run headless (`QT_QPA_PLATFORM=offscreen`, no SDR device) with a seeded `bookmarks.csv` holding two entries (one tagged `Broadcast`, one `Untagged`; the second name deliberately contains a `|`).

```
\get_bookmarks                              -> '2\n95500000|KQED FM|WFM (stereo)|160000|Broadcast\n144390000|APRS   test|Narrow FM|12500|Untagged'
\get_bookmark_tags                          -> '2\nUntagged\nBroadcast'
\get_bookmarks_in_range 90000000 100000000  -> '1\n95500000|KQED FM|WFM (stereo)|160000|Broadcast'
\get_bookmarks_in_range 100 50              -> 'RPRT 1'
f                                           -> '14236000'
\set_bookmark 0                             -> 'RPRT 0'
f                                           -> '95500000'
\set_bookmark_freq 144390000                -> 'RPRT 0'
f                                           -> '144390000'
m                                           -> 'FM\n10000'
\set_bookmark 99                            -> 'RPRT 1'
\set_bookmark_freq 123                      -> 'RPRT 1'
\reload_bookmarks                           -> 'RPRT 0'
```

Notes:
- the `|` in the second bookmark's name comes back as a space (the documented sanitisation), so each record stays one line;
- `\set_bookmark` / `\set_bookmark_freq` actually retune — `f` and `m` reflect the bookmark's frequency and mode afterwards;
- out-of-range index, no-match frequency, and `lo > hi` all return `RPRT 1`.

dsward2:
Rebased onto current `master` — applies cleanly with no conflicts, GitHub shows the PR as MERGEABLE. Diff is unchanged; just confirming it's current and the runtime test above still holds. Ready for review whenever you have a chance.

